PR #2, opened 2016-08-22
Add CORS param to prevent browsers from logging cors errors in the co…

mikeng13:
Also bumps version and updates to newest version of the node-sdk

@haleybash-optimizely @aliabbasrizvi @delikat @vraja2


vraja2:
looks good

PR #3, opened 2016-08-24
Add environment detection to event builder so it can distinguish betw…

mikeng13:
Update to use node 0.1.3

@haleybash-optimizely @aliabbasrizvi @delikat @vignesh-optimizely


vraja2:
lgtm

PR #18, opened 2016-11-14
Add global export variable

### LINKED ISSUE #14: global export
I see that the SDK compiles a browser distributable via webpack. However, I don't see what the globally exported variable name is for the distributable. Unfortunately none of the docs show usage other than `require(...` which isn't helpful.

More specifically, if webpack's configuration option `output.library` were set explicitly, this variable name would be abundantly clear.

mikeng13:
@optimizely/fullstack-devs 

Fixes #14


package.json
```diff
@@ -1,10 +1,11 @@
 {
   "name": "optimizely-client-sdk",
-  "version": "1.0.0",
-  "description": "JavaScript SDK for client testing",
+  "version": "1.1.0",
+  "description": "JavaScript SDK for Optimizely X Full Stack",
   "main": "index.js",
   "scripts": {
     "test": "./node_modules/.bin/mocha ./tests.js",
+    "build": "webpack index.js dist/optimizely.min.js",
```
jasonkarns: I'd suggest changing this to: `"build": "webpack $npm_package_main dist/optimizely.min.js",` so as to reduce the duplicate knowledge of the entrypoint. This way `main` is the canonical location for defining the entrypoint and the build script doesn't need tweaked if the entrypoint is moved.

jasonkarns:
Seems that with the removal of the json-loader from the webpack config, it can be removed as a devDep, as well.

mikeng13:
Thanks for the suggestions @jasonkarns . Will address those in a PR into the devel branch as it would just delay this PR.

PR #29, opened 2017-02-01
Publish distributable

### LINKED ISSUE #28: Distributable is missing from npm package
The compiled distributable is missing from the npm package.

See: https://unpkg.com/optimizely-client-sdk@1.2.1/

The entire dist dir is being ignored from npm because of .npmignore: https://github.com/optimizely/javascript-sdk/blob/master/.npmignore#L9

The .npmignore file ought to be removed (so that ignores are only listed in .gitignore in a single place, not duplicated in two files).

dist/ should be git-ignored (compiled assets are not supposed to be committed to version control)

And `dist/` should be added to the package.json's `files` array so that the compiled distributable is actually included in the published npm package. (See `npm-pack` for inspecting the contents of the tarball that is published)

jasonkarns:
Fixes #28 

- Remove dist from git so that compiled assets aren't being versioned.
- Remove .npmignore so that npm will use .gitignore for the ignore rules (less overhead and maintenance; lowers risk of errors introduced because of missing ignores between the duplicated files)
- Run build step before publishing (automatically done for npm-pack as well)
- Specifically include lib and dist dirs in the published tarball on npm

Screenshot of what a built tarball will include as a result of these changes:

image


delikat:
Addressed in #30

PR #30, opened 2017-02-01
Publish compiled distributable on npm (v1.2.2)

delikat:
Addresses #28.

- Removes .npmignore and consolidates ignore rules in .gitignore
- Removes dist directory from source control
- Includes lib and dist directories in tarball published to npm
- Increments version to 1.2.2


.gitignore
```diff
@@ -1,6 +1,10 @@
 .DS_STORE
+.coveralls.yml
```
jasonkarns: Assuming this was an accident? Shouldn't coveralls be versioned?

.gitignore
```diff
@@ -1,6 +1,10 @@
...
+.travis.yml
```
jasonkarns: Same here. This file is already versioned in the repo. Shouldn't be ignored.

delikat:
@jasonkarns Thanks so much for catching those errors. I'll fix them now.

PR #35, opened 2017-03-10
Bump to version 1.3.0

onufryk:
@optimizely/fullstack-devs


mikeng13:
We should make it 1.3 because we introducing a new feature. Also we should wait until we publish node to npm. I think Delikat or Elliot still need to merge in the log level config changes to node

onufryk:
@mikeng13 right, we can't merge it yet

delikat:
@onufryk @mikeng13 The logLevel config changes are already in `devel`, so feel free to publish to npm once this lands 😸

PR #56, opened 2017-05-25
Remove redundant pathing

jasonkarns:
Manually prepending `$(npm bin)` to dependencies is redundant in npm scripts. Npm already prepends `$(npm bin)` to `$PATH` before executing scripts.

PR #58, opened 2017-06-02
npm already adds node_modules/.bin to PATH

jasonkarns:
Again, same issue as https://github.com/optimizely/javascript-sdk/pull/56, there is no need to point directly at the node_modules bin because npm purposefully prepends `$(npm bin)` to `PATH` when running scripts.

(Which is in addition to the fact that `./node_modules/karma/bin/karma` is the wrong path in the first place. It's brittle and reliant on karma always keeping that file in the same location and with the same name. Correct location for binaries is where npm symlinks them, as instructed by the `bin` property of package.json: `./node_modules/.bin/karma`. But again, all this is unnecessary because npm has the feature specifically to avoid this.)


caitlinrubin-optimizely:
Hi @jasonkarns we're reviewing that change and we're going to look into the travis test failures as well. Thanks!

caitlinrubin-optimizely:
@jasonkarns can you please pull our latest changes to master and re-push your fix?

jasonkarns:
done. still failing for unrelated reasons

mikeng13:
Closing this as it is no longer relevant.

PR #73, opened 2018-02-07
change travis file and request node module

ifidoesit:


wangjoshuah:
build

package.json
```diff
@@ -35,7 +35,7 @@
     "json-schema": "^0.2.3",
     "lodash": "^4.13.1",
     "murmurhash": "0.0.2",
-    "request": "~2.81.0",
+    "request": "^2.81.0",
```
mikeng13: The reason we pinned this in the past is because the authors broke legacy support for node.  If you make this change it means that we aren't fixed to a version anymore. The better approach would be to fix it to a version that works in Node 4+, like maybe `~2.83.0` and not let it float because the authors have a history of breaking backwards compatibility with minor version bumps.

.travis.yml
```diff
@@ -1,22 +1,23 @@
 language: node_js
 node_js:
-  - "4"
-  - "5"
-  - "6"
-  - "node"
+- '4'
+- '5'
+- '6'
+- node
```
mikeng13: Can you explicitly add 7 and 8 here? The reason is because `node` will point to latest, which I think is 9 right now. We can even get rid of 4 and 5, but maybe we can look at the node adoption rates first.

PR #90, opened 2018-04-17
Added unit tests for invalid types in User-Id validation.

mfahadahmed:


packages/optimizely-sdk/lib/utils/user_id_validator/index.tests.js
```diff
@@ -45,6 +45,24 @@ describe('lib/utils/user_id_validator', function() {
           userIdValidator.validate('');
```
mikeng13: Please update the license headers to include the year 2018 --> `2016, 2018`

packages/optimizely-sdk/lib/utils/user_id_validator/index.tests.js
```diff
@@ -45,6 +45,24 @@ describe('lib/utils/user_id_validator', function() {
...
         }, sprintf(ERROR_MESSAGES.INVALID_USER_ID, 'USER_ID_VALIDATOR'));
       });
+
+      it('should throw an error if userId is int', function() {
```
mikeng13: These can be consolidated into one test case, which is, `should throw an error if userId is not a string` and you can run the 3 assertions in that one test

PR #91, opened 2018-04-19
Input validation in Activate, Track and GetVariation methods.

mfahadahmed:


mfahadahmed:
@mikeng13, we currently have two different validators (user_id_validator and string_value_validator) which are doing the same task of validating string inputs.
Should I remove the user_id_validator and validate User ID with string_value_validator? Then we can easily change the first parameter of __validateInputs to string values map.

mikeng13:
@mfahadahmed you can repurpose the `__validateInputs` method (refactor) to take in the `values` arg and perform all the needed validation inside of that method, including null checks.

mfahadahmed:
@mikeng13, The required changes have been done. Please verify.

mikeng13:
build

PR #92, opened 2018-04-23
example using local storeage to send events

thomaszurkan-optimizely:


thomaszurkan-optimizely:
Where should this end up?  Devdocs?

mikeng13:
Yeah dev docs would be the best place to document this.

packages/optimizely-sdk-core/lib/plugins/event_dispatcher/index.browser.js
```diff
@@ -18,15 +18,104 @@ var fns = require('../../utils/fns');
...
-   * Users of the SDK can provide their own implementation
-   * @param  {Object} eventObj
-   * @param  {Function} callback
-   */
-  dispatchEvent: function(eventObj, callback) {
+
+
+  queueEvent: function(eventObj) {
+    // Check browser support
+    if (typeof(Storage) !== "undefined") {
```
tylerbrandt: nit: `Storage` is just an interface; we might as well check for the thing we really want, which is `localStorage` (or abstract that into a variable; but checking for `Storage` and then assuming `localStorage` is not ideal).

tylerbrandt: Also, it seems unnecessarily defensive to check for `Storage` on every method call, instead of just the entry point (`dispatchEvent`).

packages/optimizely-sdk-core/lib/plugins/event_dispatcher/index.browser.js
```diff
@@ -18,15 +18,104 @@ var fns = require('../../utils/fns');
...
+      var events = queue = localStorage.getItem(LOCAL_STORAGE_QUEUE_NAME);
```
tylerbrandt: Why are we defining `queue` here?

packages/optimizely-sdk-core/lib/plugins/event_dispatcher/index.browser.js
```diff
@@ -18,15 +18,104 @@ var fns = require('../../utils/fns');
...
+      if (events !== null && events !== undefined) {
```
tylerbrandt: This is not exhaustive of cases where we might want to use `[]`; for example empty string will also blow up `JSON.parse`. Maybe try something like:
```
var storedEventsStr = localStorage.getItem(...)
var events = [];
if (typeof storedEventsStr === 'string') {
  try {
    events = JSON.parse(storedEventsStr);
  } catch(e) {
    // handle/log it if you want
  }
}
```

packages/optimizely-sdk-core/lib/plugins/event_dispatcher/index.browser.js
```diff
@@ -56,6 +145,29 @@ module.exports = {
...
+  /**
+   * Sample event dispatcher implementation for tracking impression and conversions
+   * Users of the SDK can provide their own implementation
+   * @param  {Object} eventObj
+   * @param  {Function} callback
+   */
+  dispatchEvent: function(eventObj, callback) {
+    if (typeof(Storage) !== "undefined") {
+      queueEvent(eventObj);
+      var event = getQueuedEvent();
```
greeshma-optimizely: `getQueuedEvent` function definition seems to require an `index` as a parameter

packages/optimizely-sdk-core/lib/plugins/event_dispatcher/index.browser.js
```diff
@@ -18,15 +18,104 @@ var fns = require('../../utils/fns');
...
-   */
-  dispatchEvent: function(eventObj, callback) {
+
+
+  queueEvent: function(eventObj) {
+    // Check browser support
+    if (typeof(Storage) !== "undefined") {
+      var events = queue = localStorage.getItem(LOCAL_STORAGE_QUEUE_NAME);
+      if (events !== null && events !== undefined) {
+        events = JSON.parse(events);
```
tylerbrandt: This is going to have performance problems, as the number of queued events increases. In particular the need to repeatedly `parse`/`stringify` the same data will likely become a bottleneck. Since we don't seem to care about collisions with other tasks writing to "our" localStorage bucket, we can cut out a lot of that overhead by keeping the queue in-memory as a list of objects, _reading_ it only on initialization, and only _writing_ it to storage otherwise.
```
// If you want you could make the EventDispatcher a class, so that it gets initialized on construction instead of when the module is loaded.

var eventQueue = [];
try {
  eventQueue = JSON.parse(localStorage.getItem(LOCAL_STORAGE_QUEUE_NAME));
} catch(e) {
  // input data corrupted, log if you want
}

exports.dispatchEvent = function(ev, callback) {
  var evIndex = eventQueue.push(ev);
  persistQueue();
  sendEvent(ev, function() {
    removeFromQueue(evIndex);
    callback();
  }
}

function persistQueue() {
  localStorage.setItem(LOCAL_STORAGE_QUEUE_NAME, JSON.stringify(eventQueue));
}
```

packages/optimizely-sdk-core/lib/plugins/event_dispatcher/index.browser.js
```diff
@@ -18,15 +18,104 @@ var fns = require('../../utils/fns');
...
+
+      var event = null;
+
+      if (events.length > 0) {
+        // perfect, it is the first one in the queue.  just remove it.
+        if (events[0] == eventToRemove) {
+          event = events.shift();
+        }
+        else {
+          var index = events.indexOf(eventToRemove);
```
tylerbrandt: This isn't going to work, once `eventToRemove` no longer referentially equals `event` that was enqueued (e.g. after one stringify/parse cycle).
```
var evs = [{foo: 'bar'}];
evs.indexOf({foo: 'bar'})
> -1
```
It would be possible to use something like lodash to fuzzy match the event object but that will be pretty expensive (in addition to needing a dependency). If you are ok with using a plain object instead as a "sparse array", you could replace the "value" with `undefined`, leaving everything at its original index. Then you just need to keep track of the index (key) it was enqueued at and pass that in to removeEvent. You'll also need to keep track of the "next index" (i.e. length) of the queue.

thomaszurkan-optimizely: i now keep an array of strings.  so, this all actually works.  however, i do like the idea we discussed.

PR #96, opened 2018-05-14
feat(utils): Add ConfigCache and ClientCache

spencerwilson-optimizely:
## Summary

- Add public types `PollingConfigCache` and `ClientCache`, both of which are implementations of a [currently] private type `AsyncCache`

How to review this: start with `AsyncCache`, then move on to the other types. It is a bit like the popular [async-cache](https://www.npmjs.com/package/async-cache) package, but with some differences. My implementation:

- Is a class, so can be `extend`ed, having methods overridden (like `seed`, whose semantics are quite different for `ClientCache` than they are for `PollingConfigCache`).
- Lacks any notions of entry expiration or eviction, since our dataset is small (Full Stack configs).
- Subclasses may own refreshing entries by calling `__execRefresh(key)`.

This doesn't yet define an _interface_ `ConfigCache`, but that could easily be done when the time comes.

## Test plan
Currently some very basic unit tests; need to flesh those out before merging. Some manual tests have validated basic `getAsync` use of both types, at-most-one-in-flight behavior, and support for tracking headers so `PollingConfigCache` can get 304 responses from the CDN.

### DIFF
--- a/packages/optimizely-sdk/lib/utils/async_cache.js
+++ b/packages/optimizely-sdk/lib/utils/async_cache.js
@@ -13,6 +13,8 @@ const enums = {
     YES_AWAIT: 'await',
   },
 
+  // Use a plain object as a unique value, against which to compare strict equality.
+  // Symbols were invented for this purpose, but those are non-transpilable ES2015.
   UNCHANGED: {},
 };
 
@@ -20,7 +22,7 @@ exports.enums = enums;
 
 /**
  * A read-through cache for any kinds of values which require async work to refresh.
- * LiveCache...
+ * AsyncCache...
  *   - lets you synchronously seed a cache entry with a given value,
  *   - offers configurable semantics for async entry access, and
  *   - ensures at most one pending refresh per entry at a time.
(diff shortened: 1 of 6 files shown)


packages/optimizely-sdk/lib/utils/config_cache.js
```diff
@@ -0,0 +1,106 @@
+const { AsyncCache, enums } = require('./async_cache');
+
+/**
+ * A ConfigCache that syncs by polling the CDN.
+ *
+ * Parametrized by which requester to use as a default, since Node.js and browser
```
mikeng13: _Parameterized_

packages/optimizely-sdk/lib/index.browser.js
```diff
@@ -13,61 +13,34 @@
...
+  ClientCache,
+};
 
-    if (config.skipJSONValidation == null) {
-      config.skipJSONValidation = true;
-    }
+/**
+ * The function that PollingConfigCache should use by default to update a config.
+ */
+async function browserRequester(url, headers) {
```
mikeng13: Maybe move this into `utils` or even `plugins` (with the idea that we'll let you configure your own requester in a later version)? Let's try not to bloat the entry point file

packages/optimizely-sdk/lib/utils/async_cache.js
```diff
@@ -0,0 +1,226 @@
...
+
+    switch (directive) {
+      case enums.refreshDirectives.ONLY_IF_CACHE_MISS:
+        return Promise.resolve(this.get(key));
+
+      case enums.refreshDirectives.YES_DONT_AWAIT:
+        this.__execRefresh(key);
+        return this.get(key);
+
+      case enums.refreshDirectives.YES_AWAIT:
```
mikeng13: We do need to handle the default case in case of user error and probably default to one of the above and log a warning about it.

packages/optimizely-sdk/lib/utils/client_cache.js
```diff
@@ -0,0 +1,109 @@
...
+    return enums.refreshDirectives.YES_AWAIT;
+  }
+
+
+  /**
+   * Ensure the cached Client for the given key is making decisions based on the config
+   * within this.configCache. _May_ involve losing per-instance state. In particular,
+   * forcedVariations and notificationListeners are lost. TODO: Fix this.
+   */
+  async __refresh(configKey, { client: currentClient, config: currentConfig } = {}) {
```
mikeng13: please document params

mikeng13:
This is stale, gonna close for now.

PR #100, opened 2018-05-19
chore: add package-lock.json

tylerbrandt:


tylerbrandt:
Was this intentional, or an oversight? If the former, we should add it to .gitignore.

PR #103, opened 2018-05-21
chore(doc): Add API docs for index and types

tylerbrandt:
- Add autogenerated API docs in api.md (generated via jsdoc2md).
- Clean up a number of typedefs and docstrings

These API docs will be easier to keep in-sync with the implementation,
and provide a basis for consumers to understand the integration
requirements, in more detail than is currently available in devdocs.


packages/optimizely-sdk/lib/types.js
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Event Dispatcher module
+ * @typedef {Object} EventDispatcher
+ * @property {EventDispatcherFn} dispatchEvent
+ */
+
+/**
+ * @callback EventDispatcherFn
+ * @param {Event} event Event to dispatch
+ * @param {Function} callback Function to call when...
```
spencerwilson-optimizely: > Function to call when...
Is this intended?

tylerbrandt: No, I'll fix it

mikeng13:
Closing as it is stale.

PR #114, opened 2018-06-12
Zeeshan/fault injection and treatment

zashraf1985:
1) Added try/catch blocks to deal with potential exceptions.
2) Initially added supporting code to collect improvement metrics. Removed all the support code in the final commit.


kellyroach-optimizely:
REQUESTING MIKE NG TO DO A FIRST CODE REVIEW

We can have additional feedback from more people besides myself.
Previously, I've mentioned that I would like SDK and Fault Injection code
separated so we continually reuse same Fault Injection code to incrementally
evaluate hypothetical crash proofing improvements to the SDK and also be
confident that the Fault Injection code not itself add any additional weight
to the SDK that Optimizely's customers will install in their apps.  I'd like the
code review to give a verdict on this goal.

The Description should also say something about the quantitative
improvement in "crash proofing" that is being accomplished by this P.R. according
to the "success metric" defined in our Crash Proofing and Fault Tolerance
design doc.  We want to be measuring this number in this P.R. and future P.R.'s,
whether they are crash proofing related or not.

zashraf1985:
@spencerwilson-optimizely I totally agree with that. Initially, i did that just for the sake of collecting results quickly. I should have taken another look on the overall design before creating PR. I will make suggested changes and run the tests again to verify results. I should be able to make the said changes by tomorrow i.e Thursday.

zashraf1985:
As per discussion with @spencerwilson-optimizely  and @kellyroach-optimizely  on the Slack thread [crash-proof-sdks](https://optimizely.slack.com/messages/CABD0P9PE/), I am closing this pull request. We have decided that we will go with a strategy of adding try catch blocks only at the entry points as suggested by @spencerwilson-optimizely and endorsed by @mikeng13. I am not making requested changes to this branch in order to keep the history clean so that it does not include interim commits where we added try catch blocks at every level and reverted back. 

I believe a better strategy is to reset to an older commit and then do it cleanly from there in another branch. I have created a new branch [zeeshan/faultInjectionAndTreatmentAtEntryPoints](https://github.com/optimizely/javascript-sdk/tree/zeeshan/faultInjectionAndTreatmentAtEntryPoints). I will create a new pull request for this branch when i am done making the suggested changes

PR #115, opened 2018-06-13
Re-enable sourceclear

ooola:


.travis.yml
```diff
@@ -19,8 +19,7 @@ before_script:
 - npm install grunt-cli -g
 install: npm install
 addons:
-  # Disabled temporarily while we work out integration issues
-  # srcclr: true
+  srcclr: true
```
aliabbasrizvi: Should we only run it for 1 node configuration

ooola: No, we'll run it on everything to get better usage data.

PR #123, opened 2018-06-22
Fix(nodejs/event_dispatcher): update error/resp handlers

### LINKED ISSUE #122: Impression events don't respect HTTPS_PROXY
Since #98 removed the request library in favour of the stock Node HTTP implementation, the Optimizely SDK no longer supports HTTP(S) Proxies, as Node does not proxy by default (as documented here https://github.com/nodejs/node/issues/15620).

We host our feature branches on a PAAS that requires outbound HTTP connections be proxied: As a result this bug currently causes our application to crash with the error below when calling `optimizely.activate`

```Shell
   2018-06-21T16:49:56.60+0100 [APP/PROC/WEB/0] ERR events.js:183
   2018-06-21T16:49:56.60+0100 [APP/PROC/WEB/0] ERR       throw er; // Unhandled 'error' event
   2018-06-21T16:49:56.60+0100 [APP/PROC/WEB/0] ERR       ^
   2018-06-21T16:49:56.60+0100 [APP/PROC/WEB/0] ERR Error: connect ECONNREFUSED 34.228.76.60:443
   2018-06-21T16:49:56.60+0100 [APP/PROC/WEB/0] ERR     at Object._errnoException (util.js:1022:11)
   2018-06-21T16:49:56.60+0100 [APP/PROC/WEB/0] ERR     at _exceptionWithHostPort (util.js:1044:20)
   2018-06-21T16:49:56.60+0100 [APP/PROC/WEB/0] ERR     at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1198:14)

```

This error can't be caught, because it occurs asynchronously to the main control flow of the application, and thus causes the entire application server to crash.

### LINKED ISSUE #124: (nodejs) Failure to log "success" message when event dispatched
Since #98, the `requestCallback` never gets called with a second argument, because its role is a 'response' handler for `HttpClientRequest`:

> The optional callback parameter will be added as a one-time listener for the 'response' event.
src: https://nodejs.org/api/http.html#http_http_request_options_callback

As a result, the callback defined in `__dispatchEvent` currently never gets called in nodejs, and as a result we don't log any messages for `LOG_MESSAGES.ACTIVATE_USER` or `LOG_MESSAGES.TRACK_EVENT`.

tylerbrandt:
- (nodejs) Prevent crash when `http`/`https` emits an error by adding an 'error' listener
- (nodejs) Fix `requestCallback` to conform to its role as a 'response' listener; notably, ensure it is called when the first argument is interpeted as the `http.IncomingMessage`, and _not_ called in the event of an error (as expected by `Optimizely#_sendImpressionEvent`/`Optimizely#track`).

Tested that this version (as 2.1.2-0) _does_ emit the messages expected in the demo app.

Fixes #122 
Fixes #124


packages/optimizely-sdk/lib/plugins/event_dispatcher/index.node.tests.js
```diff
@@ -98,5 +98,18 @@ describe('lib/plugins/event_dispatcher/node', function() {
         eventDispatcher.dispatchEvent(eventObj, callback);
       });
     });
+
+    it('calls callback with err when one is emitted', function(done) {
```
tylerbrandt: Note: before changing the event_dispatcher code, the `.dispatchEvent` call did indeed throw an error.

PR #127, opened 2018-06-26
Fault Treatment at entry points of SDK

zashraf1985:
1. Added try/catch blocks at the entry points in the SDK.
2. Added ErrorHandler and logging to all the catch blocks.


packages/optimizely-sdk/lib/core/notification_center/index.js
```diff
@@ -30,10 +30,12 @@ var MODULE_NAME = 'NOTIFICATION_CENTER';
  * @constructor
  * @param {Object} options
  * @param {Object} options.logger An instance of a logger to log messages with
+ * @param {object} options.errorHandler An instance of errorHandler to handle any unexpected error
  * @returns {Object}
  */
 function NotificationCenter(options) {
```
spencerwilson-optimizely: `NotificationCenter` is actually not an entry point to the SDK. [This line](https://github.com/optimizely/javascript-sdk/blob/6d6bb2400fb3adad1ce0d91e0cb6980e68685768/packages/optimizely-sdk/lib/optimizely/index.js#L113-L115) within the `Optimizely` constructor function, which in turn is only called from the package's public `createInstance` function, is the sole place that calls into any code from this module. This means that all of the changes in this file can be reverted:
```sh
git checkout master -- packages/optimizely-sdk/lib/core/notification_center/index.js
```

packages/optimizely-sdk/lib/index.browser.js
```diff
@@ -39,36 +40,43 @@ module.exports = {
...
+        clientVersion: enums.CLIENT_VERSION,
+        errorHandler: defaultErrorHandler,
+        eventDispatcher: defaultEventDispatcher,
+        logger: logger.createLogger({logLevel: logLevel})
+      }, config);
 
-    return new Optimizely(config);
+      return new Optimizely(config);
+    } catch (e) {
+
```
spencerwilson-optimizely: Could you please delete empty lines like these? The codebase currently doesn't have this style, and any PR that changes that should probably do so wholesale at the same time across the whole codebase, to maintain consistency.

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -47,73 +47,84 @@ var FEATURE_VARIABLE_TYPES = enums.FEATURE_VARIABLE_TYPES;
  * @param {Object} config.userProfileService
  */
 function Optimizely(config) {
-  var clientEngine = config.clientEngine;
-  if (clientEngine !== enums.NODE_CLIENT_ENGINE && clientEngine !== enums.JAVASCRIPT_CLIENT_ENGINE) {
-    config.logger.log(LOG_LEVEL.INFO, sprintf(LOG_MESSAGES.INVALID_CLIENT_ENGINE, MODULE_NAME, clientEngine));
-    clientEngine = enums.NODE_CLIENT_ENGINE;
-  }
```
spencerwilson-optimizely: Same thing here and elsewhere: please delete new blank lines.

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -220,71 +246,87 @@ Optimizely.prototype._sendImpressionEvent = function(experimentKey, variationKey
...
+        }
+      );
+    } catch (ex) {
+      this.logger.log(LOG_LEVEL.ERROR, ex.message);
+      var failedTrackLogMessage = sprintf(LOG_MESSAGES.NOT_TRACKING_USER, MODULE_NAME, userId);
+      this.logger.log(LOG_LEVEL.INFO, failedTrackLogMessage);
+      this.errorHandler.handleError(ex);
+    }
+  } catch (e) {
+    if (e.message.startsWith("PROJECT_CONFIG")) {
```
spencerwilson-optimizely: What's this branch about? Could you add a comment above it explaining?

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -296,25 +338,38 @@ Optimizely.prototype.track = function(eventKey, userId, attributes, eventTags) {
...
+      }
+
+      return this.decisionService.getVariation(experimentKey, userId, attributes);
+    } catch (ex) {
+      this.logger.log(LOG_LEVEL.ERROR, ex.message);
+      this.errorHandler.handleError(ex);
+      return null;
+    }
+  } catch (e) {
+    if (e.message.startsWith(MODULE_NAME)) {
```
spencerwilson-optimizely: Likewise here, what is this?

packages/optimizely-sdk/package.json
```diff
@@ -19,7 +19,8 @@
     "lint": "eslint lib/**",
     "cover": "istanbul cover _mocha ./lib/*.tests.js ./lib/**/*.tests.js ./lib/**/**/*tests.js",
     "coveralls": "npm run cover -- --report lcovonly && cat ./coverage/lcov.info | coveralls",
-    "prepare": "npm test && npm run build"
+    "prepare": "npm test && npm run build",
+    "fault-injection": "node lib/fault_injection/test_processes_manager.js"
```
spencerwilson-optimizely: Changes to this file and to `package-lock.json` can be discarded, since I don't think (please correct if I'm wrong) you intend to merge lib/fault_injection_test_processes_manager.js. If you _do_ intend to commit that to the repo, it would probably go in a directory other than `lib/`, which contains the actual source code of the SDK. Maybe a new directory calls 'tools' or something.

```
git checkout master -- packages/optimizely-sdk/package.json packages/optimizely-sdk/package-lock.json
```

packages/optimizely-sdk/lib/core/notification_center/index.js
zashraf1985: I agree to the extent of `NotificationCenter` constructor and that is the reason why you wont see a try/catch block in `NotificationCenter` constructor. The reason to put try/catch blocks in other methods is because an instance of `NotificationCenter` is available to the consumer as a public member of `OptimizelyInstance`. Consumer application can call `addNotificationListener`, `removeNotificationListener`, `clearAllNotificationListeners`, `clearNotificationListeners`,  `sendNotifications` using `NotificationCenter` instance. The idea was to have try/catch blocks on all the functions which can be directly called by consumer application so that if any of ther function crashes because of an unexpected exception/error, it does not in turn crash the consumer application. If you still think we should revert it, please let me know.

packages/optimizely-sdk/lib/optimizely/index.js
zashraf1985: Some methods throw expected exceptions just like checked exceptions in java. We do not want to catch and eat those exceptions because we are expecting consumer app to expect and handle them, so we simply re-throw them if they are expected ones. This is at least my understanding. Please let me know if you see any problem with my observation. And yes, i will add a comment in the code once you approve the approach.

packages/optimizely-sdk/lib/optimizely/index.js
zashraf1985: Same reason as the last one.

packages/optimizely-sdk/lib/core/notification_center/index.js
spencerwilson-optimizely: Oh yes, you're right, excellent catch. `NotificationCenter` instance methods are entry points which I overlooked.

packages/optimizely-sdk/lib/optimizely/index.js
spencerwilson-optimizely: > we are expecting consumer app to expect and handle them

On the contrary, I don't think we want to throw _any_ exceptions from the SDK. So this `if` block can just be deleted, and L327-L330 can remain.

packages/optimizely-sdk/lib/optimizely/index.js
spencerwilson-optimizely: Likewise here, please remove this `if` branch.

packages/optimizely-sdk/package.json
spencerwilson-optimizely: Please revert the changes to both package.json and package-lock.json.

packages/optimizely-sdk/lib/index.browser.js
zashraf1985: That was a mistake on my part and i have removed them. I did not make any intentional changes to the style, it happened because i used regex based find/replace to remove fault injection code, which removed the fault injection lines but it did not remove the line breaks above and below.

packages/optimizely-sdk/package.json
zashraf1985: Done

packages/optimizely-sdk/lib/optimizely/index.js
zashraf1985: Done

packages/optimizely-sdk/lib/index.browser.js
```diff
@@ -39,36 +40,43 @@ module.exports = {
...
+        errorHandler: defaultErrorHandler,
+        eventDispatcher: defaultEventDispatcher,
+        logger: logger.createLogger({logLevel: logLevel})
+      }, config);
 
-    return new Optimizely(config);
+      return new Optimizely(config);
+    } catch (e) {
+      
+      config.logger.log(LOG_LEVEL.ERROR, e.message);
```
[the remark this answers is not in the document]
zashraf1985: Done

packages/optimizely-sdk/lib/core/notification_center/index.js
zashraf1985: I assume you agree this is correct and does not need to be reverted back. I am keeping it as is. let me know if my understanding is wrong.

packages/optimizely-sdk/lib/index.node.js
```diff
@@ -25,6 +25,8 @@ var sprintf = require('sprintf');
 
 var Optimizely = require('./optimizely');
 
+var LOG_LEVEL = enums.LOG_LEVEL;
```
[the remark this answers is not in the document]
spencerwilson-optimizely: While that is true, I think we can accept that risk for now, and make a note to ourselves to return to this problem later once we know _how_ we wish to handle this risk.

PR #128, opened 2018-06-27
Patrickshih/webpack.config-file

shihpatrick:
Summary
- 
- Rewrote build scripts as a webpack.config.js file
- Changed relevant build scripts to support new config file
- Utilized environmental variables as flags/parameters to separate and specify builds
    - `env.mode = [development | production]`
    - `env.platform = [node | browser]`
    - `env.targ = [cjs | umd]`
- upgraded webpack to 3.x, but could not do webpack 4.x because webpack 4.x does not support node4

With a webpack config file, I would be able to simplify the task to add typescript into the Javascript SDK. Also, the config file provides added modularity to the existing codebase. 

Test plan
-
- Ran `npm run build` to ensure that dist/ files were being created
- Built all dist/ files twice (one with webpack.config.js file, one with master) and `diff` each of them
  - Expected: no difference between each file
  - Actual: Some noticeable differences. Discussed with @spencerwilson-optimizely, and the differences could be because of the upgrade from webpack 2.x.x to 4.x.x. Concluded that these differences should not affect the functional outcome. 
  - Example of a diff-*.txt file (diff-cjs.txt): 
image


packages/optimizely-sdk/package.json
```diff
@@ -10,12 +10,12 @@
     "build": "npm run build-node && npm run build-browser",
     "build-node": "npm run build-node-dev",
     "build-browser": "npm run build-dev-cjs && npm run build-dev-umd && npm run build-prod-cjs && npm run build-prod-umd",
-    "build-node-dev": "webpack lib/index.node.js dist/optimizely.node.js --target=node --output-library-target=commonjs2",
-    "build-node-prod": "webpack -p lib/index.node.js dist/optimizely.node.min.js --target=node --output-library-target=commonjs2",
-    "build-dev-cjs": "webpack lib/index.browser.js dist/optimizely.browser.cjs.js --output-library-target=commonjs",
-    "build-dev-umd": "webpack lib/index.browser.js dist/optimizely.browser.umd.js --output-library=optimizelyClient --output-library-target=umd",
-    "build-prod-cjs": "webpack -p lib/index.browser.js dist/optimizely.browser.cjs.min.js --output-library-target=commonjs",
-    "build-prod-umd": "webpack -p lib/index.browser.js dist/optimizely.browser.umd.min.js --output-library=optimizelyClient --output-library-target=umd",
+    "build-node-dev": "webpack --env.mode=development --env.platform=node --env.targ=cjs",
```
spencerwilson-optimizely: nit: I'd probably like to rename the env option to `--env.target` rather than `targ`, since abbreviations can add unnecessary complexity.

spencerwilson-optimizely:
I think these changes were obviated by https://github.com/optimizely/javascript-sdk/pull/135. Going to close this.

PR #130, opened 2018-06-28
chore(ci): use verbose logs for lerna

tylerbrandt:
Attempt to understand why travis-ci builds are hanging (due to how
travis-ci terminates builds that don't log after a long time, this might
have the "side effect" of preventing them from terminating as well).


tylerbrandt:
obsolete

PR #133, opened 2018-06-28
fix(package): Give Node.js consumers the unbundled package

spencerwilson-optimizely:
I'm not sure it's conventional to distribute pre-bundled packages.


spencerwilson-optimizely:
I don't think so! See https://github.com/optimizely/javascript-sdk/pull/135

PR #134, opened 2018-06-28
chore(test): Delete unused karma.conf.js

spencerwilson-optimizely:
## Summary

- Delete unused Karma conf file.

We seem to only use karma.bs.conf.js.

## Test plan
CI.


mikeng13:
I think this is Auto generated when you run the browser tests locally. We could probably add it to .gitignore as well. Anyway, not gonna block the merge =)

PR #129, opened 2018-06-28
refactor(ci): Bypass Lerna and repurpose the 'test-travis' run-script

spencerwilson-optimizely:
## Summary

- Bypass Lerna things in CI
- Don't install `grunt`. I presume it was installed because `test-travis` refers to it, but locally-installed executable modules are on the $PATH in run-scripts so global installation is unnecessary.
- Rename `test-travis` to `test-xbrowser` and modify it such that it only runs cross-browser tests in BrowserStack (via `grunt`), rather than re-running unit tests headlessly+locally _again_ with `npm test` (which is done as part of the `prepare` lifecycle hook that's invoked during `npm install`). The rename makes it clearer that the run-script could plausibly, usefully be run locally as well (say, if one was debugging a consistent failure in some browser).

Lerna seems to be correlated with some recent sporadic timeout failures:
- https://travis-ci.org/optimizely/javascript-sdk/jobs/397898591
- https://travis-ci.org/optimizely/javascript-sdk/jobs/397898592
- https://travis-ci.org/optimizely/javascript-sdk/jobs/397568472
- https://travis-ci.org/optimizely/javascript-sdk/jobs/397568474
- https://travis-ci.org/optimizely/javascript-sdk/jobs/396745267

This PR `cd`'s the Travis CI pipeline directly into the optimizely-sdk package's dir, and does things there. Doing so seems to reduce the mean CI time (compute time on Travis machines, not wall-clock time) from about 21 mins to 15 mins.

## Test plan
This PR's CI run: [#288](https://travis-ci.org/optimizely/javascript-sdk/builds/397977186).

PR #138, opened 2018-06-30
fix(publish): Change prepare to prepublishOnly

spencerwilson-optimizely:
## Summary

- Convert `prepare` hook to a (not-equivalent) `prepublishOnly`, reducing the number of times unit tests are run in `prepare`-literate builds by 1.

https://github.com/optimizely/javascript-sdk/pull/129 introduced a regression where unit tests stopped being run for Node.js v4 and v6. Prior to that PR, unit tests were run
- v4 and v6: once
- v8, v9, v10, ... (versions whose npm understands the `prepare` hook): twice

After that PR was merged, the number of times unit tests were run in each build went down by 1, _for every build_. That means v4 and v6 stopped running unit tests altogether.

https://github.com/optimizely/javascript-sdk/commit/2faa9c433f79ab720fdd337b81f7003d999dfbdb fixed the regression by unconditionally running unit tests on all matrix-based builds (and explicitly adding a build to run xbrowser tests). With this, Node.js 8+ builds resumed running unit tests twice ([example](https://travis-ci.org/optimizely/javascript-sdk/jobs/399704185)).

Regarding the `prepare` hook: npm@4 added the [`prepare`](https://docs.npmjs.com/misc/scripts#prepublish-and-prepare) hook with that fires on prepublish and postinstall. This is strange. Why run tests when someone installs the package? They see the mocha output in their shell, which is likely to surprise.

This PR stops the unconventional test runs on postinstall.

## Test plan
CI.


tylerbrandt:
Hm I'm not sure what happened here; [at the time ](https://github.com/optimizely/javascript-sdk/pull/108)that I _added_ the "prepare" hook I didn't remove anything; I was just attempting to codify in code what we currently [require](https://docs.google.com/document/d/1ICRgozSO_S30IuJ8mVD4VwbAdmNuPV9mtu8U3knI1VI/edit#heading=h.l821qcggqhfm) from users to deploy. So the regression must have come later.

spencerwilson-optimizely:
@tylerbrandt I think this was a regression due to https://github.com/optimizely/javascript-sdk/pull/129. Before _that_ PR, we unconditionally ran unit tests once, and we ran them twice in the case of npms which understand `prepare`.

spencerwilson-optimizely:
Was in the process of doing that, you're just darn fast at reviewing, haha.

PR #139, opened 2018-06-30
feat(ci): Only run test-xbrowser once

spencerwilson-optimizely:
## Summary

- Only run cross-browser tests once, rather than _n_ times for each Node.js version in the Travis CI matrix

This should reduce contention in both Travis CI and BrowserStack.

## Test plan
CI.


spencerwilson-optimizely:
Gotta try rewriting to use [this](https://docs.travis-ci.com/user/customizing-the-build#Explicitly-Including-Jobs)

PR #144, opened 2018-07-03
chore(GitHub): Add CODEOWNERS

spencerwilson-optimizely:
Could also be `* @optimizely/fullstack-devs`, but I'm not sure we want to email everyone on the team for every SDK repo.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+*       @mikeng13 @tylerbrandt @spencerwilson-optimizely


tylerbrandt:
I would prefer to use teams for this. Could we (should we) create an `@optimizely/javascript-sdk` team?

tylerbrandt:
PS I have been using fullstack-devs in fullstack-sdk-compatibility-suite and nobody has complained...

spencerwilson-optimizely:
> Could we (should we) create an @optimizely/javascript-sdk team?

I think that's reasonable, though not sure if having the team defined within an organization precludes people from outside the org being on the team. If the team is semantically 'required reviewers' though, well, that probably just means optinauts, anyway.

> I have been using fullstack-devs in fullstack-sdk-compatibility-suite and nobody has complained...

heh, true, but all SDK folks have _some_ stake in that repo, whereas (arguably) Sally iOS Expert is probably less interested in the PR-by-PR goings-on of javascript-sdk.

mikeng13:
build-e2e

mikeng13:
Closing as obsolete

PR #149, opened 2018-08-08
docs(README): Add credits and additional code

spencerwilson-optimizely:
Marcelo's asked us to add "Credits" and "Additional Code" sections, where:

> "Credits" is for dependencies that Optimizely distributes (i.e. any code that they download from us, for example, from a repo we set up). If there is a dependency that the customer separately installs / pulls from some non-Optimizely source, then that goes into the "Additional Code" section. The Python SDK, for example, has a few runtime dependencies like `pip` that the customer needs to install when setting up their application environment, but these are not distributed by Optimizely.

I take this to mean Credits is the code checked into this repo (including vendored deps, of which javascript-sdk has none), and "Additional Code" is any other code that is required for prod usage.

### DIFF
--- a/packages/optimizely-sdk/README.md
+++ b/packages/optimizely-sdk/README.md
@@ -67,3 +67,49 @@ npm run test-xbrowser
 
 [.travis.yml](/.travis.yml) contains the definitions for `BROWSER_STACK_USERNAME` and `BROWSER_STACK_ACCESS_KEY` used in CI. These values are Optimizely's BrowserStack credentials, encrypted with our Travis CI public key. These creds can be rotated by following [these docs](https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml).
 
+## Credits
+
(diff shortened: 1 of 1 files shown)


packages/optimizely-sdk/README.md
```diff
@@ -67,3 +67,49 @@ npm run test-xbrowser
...
+    "licenses": "MIT",
+    "repository": "https://github.com/kelektiv/node-uuid"
+  }
+}
+```
+
+To regenerate this, run the following command:
+
+```sh
+npx license-checker --production --json | jq 'map_values({ licenses, publisher, repository }) | del(.[][] | nulls)'
```
tylerbrandt: Is 'license-checker' built into node or something? I don't see it in package.json.

spencerwilson-optimizely: That's the magic of `npx`. I think it fetches it if necessary, storing it in some temp location.

mikeng13: That is pretty neat

PR #152, opened 2018-08-20
fix(build): Resume UMD bundle build, and restore docs

spencerwilson-optimizely:
## Summary
- Restore the `build-browser-umd` npm script, which produces dist/optimizely.browser.umd.min.js
- Change the UMD lib name `optimizelyClient` -> `optimizelySdk`, since the `createInstance` client factory function may be one of several exports of the SDK, in the future.
- (not a visible change) Change Markdown indentation of some bull points, to please my Markdown editor [Typora](https://typora.io/)

This undoes one bit of #135. I hadn't realized that this asset _was_ documented, and does have some uses, like faster/easier prototyping in a browser.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,10 @@
   "version": "1.0.0",
   "name": "optimizely-sdk-packages",
   "scripts": {
-    "postinstall": "lerna bootstrap",
+    "postinstall": "lerna bootstrap && lerna run build",
+    "build": "lerna run build",
     "clean": "lerna run clean",
-    "publish": "lerna publish",
+    "publish": "npm run build && lerna publish",
     "test": "lerna run test --stream",
     "test-xbrowser": "lerna run test-xbrowser --stream"
   },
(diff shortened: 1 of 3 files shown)


spencerwilson-optimizely:
The [current docs](https://docs.developers.optimizely.com/full-stack/docs/install-the-sdk#section-javascript) say
> If you aren't able to install via npm or to use the SDK in a non-CommonJS fashion, follow the instructions in the SDK's README to build a minified snippet. You can then access the SDK from window.optimizelyClient. Note that this exposes the SDK as a whole, and you can instantiate a specific client using window.optimizelyClient.

...which IMO is pretty confusing, but yes they _do_ duplicate what was previously (and with this PR, again) mentioned in the package readme. I suppose we could keep it until if/when the package has exports from its entry point other than `createInstance`. I'm not a huge fan of the alias idea since while it makes backwards compat totally smooth, autocomplete in browser JS consoles will show both `optimizelyClient` and `optimizelySdk`, which may cause confusion.

@mikeng13 what do you think?

mikeng13:
Yeah we do have devs out there who are using `optimizelyClient`. I am fine with breaking support for it now in favor of `optimizelySdk`. Just make sure to update the old docs (dev docs) and the new docs at [READme](https://docs.developers.optimizely.com/full-stack/docs/instantiate-a-client)

mikeng13:
build

spencerwilson-optimizely:
Okay, that's good few users are affected. However, I think it's valuable to maintain backwards compatibility. What do yall think of

1. Aliasing, per @tylerbrandt 's suggestion
2. Logging a warning on first access to the alias

See the updated test plan in the PR description for an example of how this works.

PR #153, opened 2018-08-20
chore(deps): Migrate to sprintf-js

spencerwilson-optimizely:
## Summary

- Migrate from [`sprintf`](https://www.npmjs.com/package/sprintf) to [`sprintf-js`](https://www.npmjs.com/package/sprintf-js)

The former is deprecated, and as a result when customers install the package they see
```sh
$ npm i @optimizely/optimizely-sdk
npm WARN deprecated sprintf@0.1.5: The sprintf package is deprecated in favor of sprintf-js.
```
This migration resolves that, so now it's a silent, warning-less install.

## Test plan
Exhaustiveness:
```sh
$ ag 'require.*sprintf('\''|")\)' lib/ ; echo $?
1
```

Also, CI.


tylerbrandt:
I doubt we are stubbing sprintf so if the tests pass should be fine.

PR #157, opened 2018-08-22
chore(repo): Create issue template

spencerwilson-optimizely:
I made this bespoke, but reviewers please let me know if you're aware of any prior art that you want to just adapt.

Is the `Issue = Enhancement | Bug` taxonomy comprehensive (for this SDK, at least)?


.github/issue_template.md
```diff
@@ -0,0 +1,29 @@
+Thanks for filing in issue! Are you proposing an enhancement or reporting a bug?
```
tylerbrandt: We should separate instructions from the content we want them to provide (GitHub supports `<!-- comments -->` for the former; see e.g. https://raw.githubusercontent.com/mishoo/UglifyJS2/master/.github/ISSUE_TEMPLATE.md; that way you don't have to *delete* a bunch of stuff.).

.github/issue_template.md
```diff
@@ -0,0 +1,29 @@
...
+
+**Browser and version:**
+
+**`node` version:** 
+
+**`npm` version:** 
+
+Any other relevant tools (like module bundlers, transpilers, etc.): 
+
+## Expected behavior
```
tylerbrandt: I like the three-part guide of "What were you trying to do?" (may help resolve [X/Y problem](http://xyproblem.info) issues), "What did you expect to happen?", "What actually happened?" I encountered something like this scheme recently when reporting an issue with [npm](https://npm.community) and it was helpful.

PR #159, opened 2018-08-27
Eliminate lodash functions that have browser-supported native alternatives

quarklemotion:
- This PR is intended to slightly reduce the bundle size impact of the optimizely SDK when the client project does not already utilize lodash v4.x.
- Switch from `_.isArray` to `Array.isArray`
    - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray
- Switch from `_.filter` to `Array.prototype.filter`
    - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter
- Switch from `_.reduce` to `Array.prototype.reduce`
    - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce
- Fix existing lint warnings

### DIFF
--- a/packages/optimizely-sdk/lib/core/event_builder/index.js
+++ b/packages/optimizely-sdk/lib/core/event_builder/index.js
@@ -115,7 +115,7 @@ function getImpressionEventParams(configObj, experimentId, variationId) {
  * @param  {Object} logger                    Logger object
  * @return {Object}                           Conversion event params
  */
-function getVisitorSnapshot(configObj, eventKey, eventTags, experimentsToVariationMap, logger) {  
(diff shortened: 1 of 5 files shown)


quarklemotion:
interesting - I would not have thought that our clients would be monkey-patching the native APIs - scary!  Oh well, will close this then - thanks!

nchilada:
@quarklemotion haha yeah, _we_ generally don't monkey-patch, but it's possible that a _customer's_ first-party or third-party code would apply a breaking patch.

This concern technically applies to **all** dynamic languages for which we have SDKs, but I think it's especially relevant for the JavaScript SDK which may run in relatively-lawless Web contexts. 😅

PR #160, opened 2018-08-30
Updating lerna to latest version

aliabbasrizvi:


spencerwilson-optimizely:
```
[javascript] > lerna bootstrap && lerna run build
[javascript] 
[javascript] lerna info version 2.11.0
[javascript] lerna ERR! EMISMATCH Incompatible local version of lerna detected!
[javascript] lerna ERR! EMISMATCH The running version of lerna is 2.11.0, but the version in lerna.json is 3.2.1.
[javascript] lerna ERR! EMISMATCH You can either run 'lerna init' again or install 'lerna@^3.2.1'.
[javascript] npm info lifecycle optimizely-sdk-packages@1.0.0~postinstall: Failed to exec postinstall script
```

Looks like in [the `postinstall` hook](https://github.com/optimizely/javascript-sdk/blob/8f9bcb3a7e826b4830195cf088b2d6f0a9c91452/package.json#L6) triggered during the testapp's [installation step](https://github.com/optimizely/javascript-testapp/blob/a61f6d6a5bde11ed303a7d33730e44082a4a8146/Dockerfile#L22), global lerna detects a local `lerna@2` was just installed, and blows up.

I think the testapp being aware of lerna is lame, and I'd like to move in a direction where that's not the case. As a first step, I tried moving this repo's top-level package.json's `lerna` from devDependencies to dependencies, and that installed okay. This is because the fundamental problem is: `postinstall` requires that things in `devDependencies` are installed, but **devDependencies aren't always installed.** (e.g., if you do `npm install foo`, foo's devDependencies are not installed).

I tried moving it to a regular dep, and it started working. I'll commit to this PR and see if it works against quay's javascript-testapp:latest.

spencerwilson-optimizely:
Oh, and for the record, this PR is to satisfy SourceClear reporting vulns in this repo.

aliabbasrizvi:
@mikeng13 good to go?

mikeng13:
Yep! This is exactly what I was talking to you about yesterday. Thanks @spencerwilson-optimizely

PR #161, opened 2018-09-03
fix(initialize): Prevent SDK from initializing if the datafile version in invalid.

mfahadahmed:


mfahadahmed:
Currently most of the datafile parsing and validations are performed inside Optimizely constructor. The config_validator module actually validates logger and error handler instances.
Should we move all those datafile validations from Optimizely constructor to config_validator?

mikeng13:
Yep, we should move the version validation out of the project_config into the config validator, as well as lines 63 through 77 from the Optimizely constructor into the config validator as well. Moving code out of that constructor makes it more maintainable and easier to follow

msohailhussain:
Please don't review this PR. Some of the changes needed to make in this PR, will let you know when ready.

packages/optimizely-sdk/lib/utils/config_validator/index.js
```diff
@@ -39,10 +46,38 @@ module.exports = {
...
     }
 
+    return true;
+  },
+
+  /**
+   * Validates the datafile
+   * @param {string}  datafile
+   * @return {Boolean} True if the datafile is valid
+   * @throws If any of the datafile validations get failed
```
mikeng13:
```
@throws If the datafile is not valid for any of the following reasons:
                - The datafile string is undefined
                - The datafile string cannot be parsed as a JSON object
                - The datafile version is not supported
```

packages/optimizely-sdk/lib/utils/config_validator/index.tests.js
```diff
@@ -46,6 +47,30 @@ describe('lib/utils/config_validator', function() {
...
+        }, sprintf(ERROR_MESSAGES.NO_DATAFILE_SPECIFIED, 'CONFIG_VALIDATOR'));
+      });
+
+      it('should complain if datafile is malformed', function() {
+        assert.throws(function() {
+          configValidator.validateDatafile('abc');
+        }, sprintf(ERROR_MESSAGES.INVALID_DATAFILE_MALFORMED, 'CONFIG_VALIDATOR'));
+      });
+
+      it('should complain if datafile version is not supported', function() {
```
mikeng13: Can you add some more test cases for things like `boolean`, `number`, `array`

packages/optimizely-sdk/lib/utils/enums/index.js
```diff
@@ -59,6 +59,7 @@ exports.ERROR_MESSAGES = {
   VARIATION_ID_NOT_IN_DATAFILE: '%s: No variation ID %s defined in datafile for experiment %s.',
   VARIATION_ID_NOT_IN_DATAFILE_NO_EXPERIMENT: '%s: Variation ID %s is not in the datafile.',
   INVALID_INPUT_FORMAT: '%s: Provided %s is in an invalid format.',
+  INVALID_DATAFILE_VERSION: '%s: This version of the Javascript SDK does not support the given datafile version: %s',
```
mikeng13: `JavaScript`

packages/optimizely-sdk/lib/index.node.tests.js
```diff
@@ -72,5 +76,60 @@ describe('optimizelyFactory', function() {
...
+
+      beforeEach(function() {
+        sinon.stub(createdLogger, 'log');
+      });
+
+      afterEach(function() {
+        createdLogger.log.restore();
+      });
+
+      it('should log an error when datafile is not provided', function() {
```
mikeng13: These are duplicate test cases from the config validator. We just need two test cases:
- datafile is valid
- datafile validation throws error (any) error

packages/optimizely-sdk/lib/utils/config_validator/index.tests.js
mikeng13: nvm this, it is not really needed.

packages/optimizely-sdk/lib/index.node.tests.js
oakbani: I think that we need to add only 1 test case here. That config validator does get called from createInstance and if it throws an error, assert that the isValidInstance is set to False. 
The datafile validations are already taken care of in config validator tests.

packages/optimizely-sdk/lib/index.node.tests.js
```diff
@@ -72,5 +76,60 @@ describe('optimizelyFactory', function() {
...
+        });
+
+        sinon.assert.called(createdLogger.log);
+        var logMessage = createdLogger.log.args[0][1];
+        assert.strictEqual(logMessage, sprintf(ERROR_MESSAGES.INVALID_DATAFILE_VERSION, 'CONFIG_VALIDATOR', '5'));
+      });
+
+      it('should create optimizely instance with valid datafile', function() {
+        var optlyInstance = optimizelyFactory.createInstance({
+          datafile: JSON.stringify(testData.getUnsupportedVersionConfig()),
```
oakbani: In case, you keep this test case, assert that validator doesn't thrown an error and isValidInstance is true. Instance will always be created like it does here with UnsupportedVersionConfig

mfahadahmed:
Oh I think none of the datafile validation tests are needed in index.node.tests.js as all of these test cases are already covered in optimizely.tests. Sorry, I forgot to remove these test cases from index.node.tests.js after changing validation logic.

packages/optimizely-sdk/lib/utils/config_validator/index.js
```diff
@@ -17,6 +17,13 @@ var sprintf = require('sprintf-js').sprintf;
 
 var ERROR_MESSAGES = require('../enums').ERROR_MESSAGES;
 var MODULE_NAME = 'CONFIG_VALIDATOR';
+var DATAFILE_VERSIONS = require('../enums').DATAFILE_VERSIONS;
```
oakbani: @mfahadahmed Please update header for this file and it's test file.

PR #162, opened 2018-09-07
fix(audiences): return null variation when audience is specified and user attributes are empty.

mikeng13:
## Summary

This fixes a bug where passing in empty attribute maps are causing the `not` audience match to evaluate to true.

```
// Given audience condition is 
// [\"not\", [\"or\", {\"type\": \"custom_attribute\", \"name\": \"customattr\", \"value\": \"exact_match\"}]]

let attributes = {}
let variation = optimizelyClient.activate('experiment_with_audience', 'user_1', attributes)

// variation should be `null`. But this bug causes audience evaluation to return true, and thus the client returns a valid variation.
```

### DIFF
--- a/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
+++ b/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
@@ -31,7 +31,7 @@ module.exports = {
     }
 
     // if no user attributes specified, return false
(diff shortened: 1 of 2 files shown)


mikeng13:
After some discussion, going to close as this is not the behavior we want to introduce.

PR #164, opened 2018-09-10
chore(ci): Ensure Travis unit tests also run on release branches.

mikeng13:
## Summary

Release branches need to have unit tests run on them given that we will soon be in a state (3.x release) where we can't always fast-forward these branches to the latest master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,7 @@ node_js:
 branches:
   only:
   - master
-  - 2.0.x
-  - 2.1.x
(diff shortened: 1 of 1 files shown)


tylerbrandt:
Can we just run on all branches?

mikeng13:
Were you thinking about running these when we open up PRs against feature branches? My only concern is that could exhaust our resource alloc on Travis. With this we will at least run Travis unit tests whenever there are PRs open against those branches.

PR #166, opened 2018-09-20
feat(eventbuilder): Track attributes with valid attribute types

msohailhussain:


packages/optimizely-sdk/lib/utils/attributes_validator/index.js
```diff
@@ -43,4 +45,8 @@ module.exports = {
       throw new Error(sprintf(ERROR_MESSAGES.INVALID_ATTRIBUTES, MODULE_NAME));
     }
   },
+
+  isAttributeValid: function(attributeKey, attributeValue) {
```
msohailhussain: need to update attribute value 'null' shouldn't be sent to logtier.

mfahadahmed: Done.

packages/optimizely-sdk/lib/core/event_builder/index.js
```diff
@@ -58,15 +59,18 @@ function getCommonEventParams(options) {
...
+    if (attributeValidator.isAttributeValid(attributeKey, attributeValue)) {
+      var attributeId = projectConfig.getAttributeId(options.configObj, attributeKey, options.logger);
+      if (attributeId) {
+        commonParams.visitors[0].attributes.push({
+          entity_id: attributeId,
+          key: attributeKey,
+          type: CUSTOM_ATTRIBUTE_FEATURE_TYPE,
+          value: attributes[attributeKey],
+        });
+      }
```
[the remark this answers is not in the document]
msohailhussain: I think not, because attributes are omitted for log tier, nothing to deal at user side. User will get all attributes. it will be confusing if logtier message is logged.

PR #167, opened 2018-09-21
fix (isFeatureEnabled): filter empty attribute values from impression event

mjc1283:
## Summary

With this change, attribute properties with null or undefined values are filtered out of impression events that are sent as a result of a feature being enabled via a feature test. This makes isFeatureEnabled behave the same way as activate with respect to null/undefined values in attributes in impression events.

## Test plan

Added new unit test


msohailhussain:
@mikeng13 @aliabbasrizvi 
I assume the right place to filter attributes is in eventbuilder which is done in this PR. 
https://github.com/optimizely/javascript-sdk/pull/166

mjc1283:
I think @msohailhussain is right - that's the best place to fix this.

PR #173, opened 2018-10-05
feat (audience match types): Update audience evaluator and project config for new audience match types

mjc1283:
### Summary

This updates project config and the audience evaluator to finish the implementation of typed audience evaluation:

- In audience_evaluator, allow evaluation to continue when no user attributes are passed (previously we were immediately returning `false` when no attributes were passed)
- Update project_config: in `createProjectConfig`, set up a by-id object for accessing audiences from both `audiences` and `typedAudiences`. Also, update `getAudiencesForExperiment` to first look in `projectConfig.typedAudiences`, and only look in `projectConfig.audiences` as a fallback.
- Added unit tests checking that all top-level methods that accept user attributes can bucket users into experiments or rollouts that use typed audiences.

### Test Plan

- New & existing unit tests
- Ran compatibility suite tests

### DIFF
--- a/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
+++ b/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016, Optimizely
+ * Copyright 2016, 2018 Optimizely
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
+++ b/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
@@ -20,8 +20,9 @@ module.exports = {
    * Determine if the given user attributes satisfy the given audience conditions
    * @param  {Object[]} audiences            Audiences to match the user attributes against
    * @param  {Object[]} audiences.conditions Audience conditions to match the user attributes against
-   * @param  {Object}   userAttributes       Hash representing user attributes which will be used in determining if
-   *                                         the audience conditions are met
+   * @param  {Object}   [userAttributes]     Hash representing user attributes which will be used in
+   *                                         determining if the audience conditions are met. If not
+   *                                         provided, defaults to an empty object.
    * @return {Boolean}  True if the user attributes match the given audience conditions
    */
   evaluate: function(audiences, userAttributes) {
@@ -30,9 +31,8 @@ module.exports = {
       return true;
     }
 
-    // if no user attributes specified, return false
     if (!userAttributes) {
-      return false;
+      userAttributes = {};
     }
 
     for (var i = 0; i < audiences.length; i++) {
--- a/packages/optimizely-sdk/lib/core/project_config/index.js
+++ b/packages/optimizely-sdk/lib/core/project_config/index.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016-2017, Optimizely
+ * Copyright 2016-2018, Optimizely
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -30,9 +31,8 @@ module.exports = {
       return true;
     }
 
-    // if no user attributes specified, return false
     if (!userAttributes) {
```
mikeng13: I think this should be fine because we do validate that the `userAttributes` is either an object or `null` and thus falsy values don't make it here

packages/optimizely-sdk/lib/tests/test_data.js
```diff
@@ -546,6 +598,73 @@ var configWithFeatures = {
       'id': '594017',
       'name': 'test_audience',
       'conditions': '["and", ["or", ["or", {"type": "custom_attribute", "name": "test_attribute", "value": "test_value"}]]]'
+    },
```
mikeng13: Just a thought... but at some point maybe we can consolidate the test datafiles that we use across unit and compat tests

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -20,8 +20,9 @@ module.exports = {
    * Determine if the given user attributes satisfy the given audience conditions
    * @param  {Object[]} audiences            Audiences to match the user attributes against
    * @param  {Object[]} audiences.conditions Audience conditions to match the user attributes against
-   * @param  {Object}   userAttributes       Hash representing user attributes which will be used in determining if
-   *                                         the audience conditions are met
+   * @param  {Object}   [userAttributes]     Hash representing user attributes which will be used in
```
mikeng13: Please also update the license header in this file and others you have touched that haven't had their license headers updated to this year.

packages/optimizely-sdk/lib/core/project_config/index.js
```diff
@@ -40,6 +40,11 @@ module.exports = {
     fns.forEach(projectConfig.audiences, function(audience) {
       audience.conditions = JSON.parse(audience.conditions);
     });
+    projectConfig.audienceIdMap = fns.keyBy(projectConfig.audiences, 'id');
+    fns.forEach(projectConfig.typedAudiences, function(audience) {
+      audience.conditions = JSON.parse(audience.conditions);
+    });
+    projectConfig.typedAudienceIdMap = fns.keyBy(projectConfig.typedAudiences, 'id');
```
mikeng13: Is there any reason to maintain a separate map? Any conflicts in ID's would get overwritten with the typed audience, which is what we want.

packages/optimizely-sdk/lib/core/project_config/index.js
```diff
@@ -215,8 +220,16 @@ module.exports = {
 
     var audienceIds = experiment.audienceIds;
     var audiencesInExperiment = [];
-    var audiencesInExperiment = fns.filter(projectConfig.audiences, function(audience) {
-      return audienceIds.indexOf(audience.id) !== -1;
+    fns.forEach(audienceIds, function(audienceId) {
+      var typedAudience = projectConfig.typedAudienceIdMap[audienceId];
```
mikeng13: Haha I dunno why we did the lookup the other way around in the first place. This is so much better :)

packages/optimizely-sdk/lib/core/project_config/index.js
```diff
@@ -215,8 +220,16 @@ module.exports = {
...
+      if (typedAudience) {
+        audiencesInExperiment.push(typedAudience);
+        return;
+      }
+      var audience = projectConfig.audienceIdMap[audienceId];
```
mikeng13: If we have just one consolidated map, we'd just have to look in one place.

packages/optimizely-sdk/lib/core/project_config/index.js
nchilada: +1, a single map would be nice.

And maybe call it `audiencesById` since it's storing audiences and not IDs? But maybe I'm just parsing it weirdly

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
nchilada: Yeah, I don't think it needs to be the job of this module to handle other types of falsy values. Gracefully overwriting all falsy values seems reasonable.

mjc1283: @mikeng13 @nchilada I'm not sure if you're agreeing with my change, or requesting that this be left the way it was.

packages/optimizely-sdk/lib/tests/test_data.js
mjc1283: Agree. The test datafiles have become large and unwieldy, and this PR is making it worse. Given @nchilada's other test-related comment also, I'm rethinking some of these tests.

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
nchilada: I _think_ we're both agreeing! At least I am. Sorry for the confusion.

mikeng13:
build

PR #175, opened 2018-10-16
feat (audiences): Audience combinations

mjc1283:
This adds support for audience combinations on experiments. If `experiment.audienceConditions` is present, it will be used as a condition tree where the leaf conditions are audience Ids.

- Condition tree evaluation and custom attribute condition evaluation are split into separate modules. Condition tree evaluation accepts a leaf evaluator function argument.
- The audience evaluator uses the condition tree evaluator on both the experiment audience conditions and the individual audience conditions.
- Project config methods are updated to support audience combinations. `getExperimentAudienceConditions` returns `audienceConditions` if it exists, otherwise it falls back to `audienceIds`.
- The audience evaluator treats conditions trees and flat lists of audience ids (the old format) the same way, because a flat list of audience ids is a valid condition tree with an implicit OR condition.


packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -13,34 +13,47 @@
...
 module.exports = {
   /**
    * Determine if the given user attributes satisfy the given audience conditions
-   * @param  {Object[]} audiences            Audiences to match the user attributes against
-   * @param  {Object[]} audiences.conditions Audience conditions to match the user attributes against
-   * @param  {Object}   [userAttributes]     Hash representing user attributes which will be used in
-   *                                         determining if the audience conditions are met. If not
-   *                                         provided, defaults to an empty object.
-   * @return {Boolean}  True if the user attributes match the given audience conditions
+   * @param  {Array}    audienceConditions    Audience conditions to match the user attributes against - can be an array
```
nchilada: Can we declare something like `{Array|Object|null|undefined}`? Since we support a single leaf condition without any operators and we accept empty conditions too.

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -13,34 +13,47 @@
...
+   * @return {Boolean}                        true if the user attributes match the given audience conditions, false
+   *                                          otherwise
    */
-  evaluate: function(audiences, userAttributes) {
+  evaluate: function(audienceConditions, audiencesById, userAttributes) {
     // if there are no audiences, return true because that means ALL users are included in the experiment
-    if (!audiences || audiences.length === 0) {
+    if (!audienceConditions || audienceConditions.length === 0) {
       return true;
     }
```
nchilada: Nit: while we're at it, maybe throw if `audiencesById` isn't an object? I think this would happen automatically if/when we get to `audiencesById[audienceId]` several lines below, so it doesn't really matter. 🤷

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -13,34 +13,47 @@
...
 
-    for (var i = 0; i < audiences.length; i++) {
-      var audience = audiences[i];
-      var conditions = audience.conditions;
-      if (conditionEvaluator.evaluate(conditions, userAttributes)) {
-        return true;
+    var evaluateConditionWithUserAttributes = function(condition) {
+      return customAttributeConditionEvaluator.evaluate(condition, userAttributes);
+    };
```
nchilada: Nit: could also use
```
var evaluateConditionWithUserAttributes = _.partialRight(customAttributeConditionEvaluator.evaluate, userAttributes)
```
, or (my preference) change the order of arguments in `customAttributeConditionEvaluator.evaluate`'s signature so that we can use the more conventional
```
var evaluateConditionWithUserAttributes = _.partial(customAttributeConditionEvaluator.evaluate, userAttributes)
```
😁

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -13,34 +13,47 @@
...
+        return conditionTreeEvaluator.evaluate(audience.conditions, evaluateConditionWithUserAttributes);
       }
+      return null;
+    };
+
+    if (conditionTreeEvaluator.evaluate(audienceConditions, evaluateAudience)) {
+      return true;
     }
 
     return false;
```
nchilada: Maybe
```
return conditionTreeEvaluator.evaluate(audienceConditions, evaluateAudience) || false;
```
? It's a little more concise and highlights the fact that we're converting falsy values (`null`) to `false`.

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -13,34 +13,47 @@
...
+      return customAttributeConditionEvaluator.evaluate(condition, userAttributes);
+    };
+
+    var evaluateAudience = function(audienceId) {
+      var audience = audiencesById[audienceId];
+      if (audience) {
+        return conditionTreeEvaluator.evaluate(audience.conditions, evaluateConditionWithUserAttributes);
       }
+      return null;
+    };
```
nchilada: Nit: for symmetry with `customAttributeConditionEvaluator.evaluate`, I wonder if we should export **this** functionality as `audienceEvaluator.evaluate`, and hoist all the other code up to `DecisionService.prototype.__checkIfUserIsInAudience`?

🔭:architecture_👨‍🚀:🤔

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -13,34 +13,47 @@
...
+    var evaluateConditionWithUserAttributes = function(condition) {
+      return customAttributeConditionEvaluator.evaluate(condition, userAttributes);
+    };
+
+    var evaluateAudience = function(audienceId) {
+      var audience = audiencesById[audienceId];
+      if (audience) {
+        return conditionTreeEvaluator.evaluate(audience.conditions, evaluateConditionWithUserAttributes);
       }
+      return null;
```
mjc1283: @nchilada @mikeng13 Any thoughts on returning `null` here? Maybe this should throw an error, and project config should reject a datafile as invalid if an experiment has invalid audience ids.

nchilada: I believe returning `null` is correct here - this function is being called at activate time so it's probably too late to decide that the datafile is invalid and to stop handling API calls.

That said, we could add an additional check for invalid audience references during initialization. If we do so, we reduce this `return null` to a mere sanity check.

[In the design doc](https://docs.google.com/document/d/12aoOd1g5naAynHGdMW1vpaGbBaaUCQHGtRhUDfFRGOg/edit#heading=h.by7wosg2x56t) I'd written:

> If an audience can't be found in [`audiences` or `typedAudiences`], the SDK may fail gracefully or spectacularly, during initialization or during experiment activation—whatever it was doing before.

but I don't mind if someone want to propose that we consistently check for invalid audience references during initialization. @mikeng13 @thomaszurkan-optimizely any thoughts on this?

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
mjc1283: What's the benefit that would justify introduction of `_.partial`?

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
mjc1283: Interesting. This does feel cleaner. I could see the other code fitting into `__checkIfUserIsInAudience`. @mikeng13 thoughts?

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
mjc1283: :+1:

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
mjc1283: With this approach, `audience_evaluator` ends up being pretty uninteresting, to the extent that it maybe shouldn't be a module.

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
nchilada: No worries. Agreed!

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
nchilada: Not much, it'd just be a little more concise, and might be more conventional in certain contexts (not sure about javascript-sdk). Don't worry about it 😄

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
nchilada: Up to you. I was just thinking out loud, especially if there was interest in using the `_.partial` pattern as we do throughout client-js. Doesn't really make a difference

packages/optimizely-sdk/lib/core/project_config/index.tests.js
```diff
@@ -532,18 +521,45 @@ describe('lib/core/project_config', function() {
...
+
+      it('should return experiment audienceIds if experiment has no audienceConditions', function() {
+        configObj = projectConfig.createProjectConfig(testDatafile.getTypedAudiencesConfig());
+        var result = projectConfig.getExperimentAudienceConditions(configObj, 'feat_with_var_test');
+        assert.deepEqual(result, ['3468206642', '3988293898', '3988293899', '3468206646', '3468206647', '3468206644', '3468206643']);
+      });
+
+      it('should return experiment audienceConditions if experiment has audienceConditions', function() {
+        configObj = projectConfig.createProjectConfig(testDatafile.getTypedAudiencesConfig());
+        var result = projectConfig.getExperimentAudienceConditions(configObj, 'audience_combinations_experiment');
```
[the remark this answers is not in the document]
mjc1283: Will add a comment - this experiment does have both `audienceIds` and `audienceConditions`

packages/optimizely-sdk/lib/core/project_config/index.tests.js
```diff
@@ -532,18 +521,45 @@ describe('lib/core/project_config', function() {
...
+        assert.deepEqual(result, ['and', ['or', '3468206642', '3988293898'], ['or', '3988293899', '3468206646', '3468206647', '3468206644', '3468206643']]);
+      });
+    });
```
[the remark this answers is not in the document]
mjc1283: I'm not sure this is worth testing.

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -3418,4 +3418,122 @@ describe('lib/optimizely', function() {
...
+      assert.strictEqual(variableValue, 150);
+    });
+
+    it('can return the default value for a feature variable from getFeatureVariableString, via excluding a user from a feature test with complex audience conditions', function() {
+      var variableValue = optlyInstance.getFeatureVariableInteger('feat2_with_var', 'z', 'user1', {
+        // Should be excluded - no audiences match with no attributes
+      });
+      assert.strictEqual(variableValue, 10);
+    });
+  });
```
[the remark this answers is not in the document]
mjc1283: Do you think it'd satisfy the same need if we assert that `audience_evaluator`'s `evaluate` was called? `audience_evaluator` has its own tests verifying the integration between `evaluate` and `condition_tree_evaluator` / `custom_attribute_condition_evaluator`.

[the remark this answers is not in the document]
nchilada: Oh, that's right. Sure, that'd probably be even better (given that it's so hard to differentiate between the two places that make direct calls to `conditionTreeEvaluator`)

packages/optimizely-sdk/lib/core/project_config/index.tests.js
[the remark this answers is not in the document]
nchilada: No worries at all!

PR #178, opened 2018-10-22
fix (typed audience evaluation): When evaluating exact match condition, properly check type of condition value

mjc1283:
## Summary

Fixes a bug in evaluation of exact match conditions. Before this change, evaluation of a condition with a wrongly-typed `value` property could return a boolean, when it should abort and return `null`. With this change, when `condition.value` of an exact match condition is not a string, boolean, or finite number, the condition will evaluate to `null`.

## Test Plan

- Added a test for an exact match condition with a wrongly-typed `value`
- Also added tests for wrongly-typed condition `value`s in 'greater than' and 'less than' conditions


mjc1283:
build

PR #182, opened 2018-11-01
fix (audiences): Update project config - no longer expect conditions of audiences in typedAudiences to be strings

mjc1283:
## Summary

Update project config to handle the latest datafile format for 3.0 with typed audience evaluation. `conditions` properties of objects in `typedAudiences` will be `Array`s, and no longer JSON-`Strings`.

## Test Plan

Updated typed audiences datafile in test data to the latest format. Existing tests should pass - no change in behavior.

### DIFF
--- a/packages/optimizely-sdk/lib/core/project_config/index.js
+++ b/packages/optimizely-sdk/lib/core/project_config/index.js
@@ -40,9 +40,10 @@ module.exports = {
     fns.forEach(projectConfig.audiences, function(audience) {
       audience.conditions = JSON.parse(audience.conditions);
     });
-    fns.forEach(projectConfig.typedAudiences, function(audience) {
-      audience.conditions = JSON.parse(audience.conditions);
-    });
+    /*
+     * Note: conditions of audiences in projectConfig.typedAudiences are not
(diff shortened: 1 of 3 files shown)


packages/optimizely-sdk/lib/core/project_config/index.js
```diff
@@ -40,9 +40,10 @@ module.exports = {
     fns.forEach(projectConfig.audiences, function(audience) {
       audience.conditions = JSON.parse(audience.conditions);
     });
-    fns.forEach(projectConfig.typedAudiences, function(audience) {
-      audience.conditions = JSON.parse(audience.conditions);
-    });
+    /*
+     * Note: conditions of audiences in projectConfig.typedAudiences are not
+     * expected to be string-encoded as they are in projectConfig.audiences.
+     */
```
nchilada: Can you move this comment three lines up and maybe adapt it as necessary? I'm kinda trained to assume each code comments applies to the statement that follows it, which is not the case here.

I guess something like 
```js
fns.forEach(projectConfig.typedAudiences, function(audience) {
  // Comment here
});
```
would also work but that's weird and might be inefficient. 😕

mjc1283:
build

PR #187, opened 2018-11-14
fix(bundling): Publish the unminified UMD bundle along with the minified one

mikeng13:
## Summary
- We need to build and publish the unminified bundle to NPM as we publicly refer to it in our docs.

This is the output of cleaning the local `/dist` directory and building manually with `npm run build-browser-umd`:
```
Hash: 1e3850c525529035af49
Version: webpack 2.7.0
Time: 1222ms
                        Asset     Size  Chunks             Chunk Names
optimizely.browser.umd.min.js  82.8 kB       0  [emitted]  main
   [1] ./lib/utils/enums/index.js 11.8 kB {0} [built]
   [3] ./~/sprintf-js/src/sprintf.js 8.67 kB {0} [built]
   [4] ./lib/utils/fns/index.js 1.21 kB {0} [built]
  [44] ./lib/utils/config_validator/index.js 2.95 kB {0} [built]
  [74] ./lib/optimizely/index.js 28.7 kB {0} [built]
  [75] ./lib/plugins/error_handler/index.js 809 bytes {0} [built]
  [76] ./lib/plugins/event_dispatcher/index.browser.js 2.42 kB {0} [built]
  [77] ./lib/plugins/logger/index.js 4.09 kB {0} [built]
  [84] ./lib/index.browser.js 2.78 kB {0} [built]
  [85] ./lib/optimizely/project_config_schema.js 7.02 kB {0} [built]
  [87] ./lib/utils/event_tags_validator/index.js 1.29 kB {0} [built]
  [88] ./lib/utils/string_value_validator/index.js 864 bytes {0} [built]
  [89] ./lib/utils/user_profile_service_validator/index.js 2.14 kB {0} [built]
 [190] ./~/lodash/forEach.js 1.35 kB {0} [built]
 [193] ./~/lodash/isEmpty.js 2 kB {0} [built]
    + 194 hidden modules
Hash: 41332d833233093fb3ac
Version: webpack 2.7.0
Time: 458ms
                    Asset    Size  Chunks                    Chunk Names
optimizely.browser.umd.js  319 kB       0  [emitted]  [big]  main
   [1] ./lib/utils/enums/index.js 11.8 kB {0} [built]
   [3] ./~/sprintf-js/src/sprintf.js 8.67 kB {0} [built]
   [4] ./lib/utils/fns/index.js 1.21 kB {0} [built]
  [44] ./lib/utils/config_validator/index.js 2.95 kB {0} [built]
[...]

PR #189, opened 2018-11-14
refactor(dist): update bundled files

tylerbrandt:
## Summary

**BREAKING**: Drop support for `window.optimizelyClient`, as presaged in the [CHANGELOG](https://github.com/optimizely/javascript-sdk/blob/master/packages/optimizely-sdk/CHANGELOG.MD#213---august-21-2018).

- Pull in changes from 2.3.x, to ensure that both optimizely.browser.umd.js and optimizely.browser.umd.min.js are defined.
- Refactor those changes to use webpack config/index.browser code instead of appending to the bundle after-the-fact.
- Actually install `webpack` as a devDependency rather than always using latest

In addition to fixing the issue from #187 in this branch, it simplifies the build script.

## Test plan

Existing unit tests pass since the "node" API for index.browser.js wasn't changed.

Created a test file called "dist/test.html" and loaded it in a browser (as well as an equivalent one that loads the min.js variant):
```html
<!DOCTYPE html>
<html>

<head>
	<script src="/optimizely.browser.umd.js"></script>
</head>

<body>
	<h1>Hello</h1>
	<script>
		console.log('optimizelySdk', typeof optimizelySdk);
		console.log('optimizelyClient', typeof optimizelyClient);
	</script>
</body>

</html>
```
Ensured that it logged "object" for the first, and "undefined" for the second:
screen shot 2018-11-14 at 3 50 30 pm

## Issues
- OASIS-3790


mikeng13:
This needs to be accompanied by a change to our public docs when we release. Can you include the an entry for a breaking change under unreleased for the deprecated `optimizelyClient` reference?

tylerbrandt:
> This needs to be accompanied by a change to our public docs when we release. Can you include the an entry for a breaking change under unreleased for the deprecated optimizelyClient reference?

Why? The docs already only refer to `window.optimizelySdk` and we published a changelog entry in August that said that `window.optimizelyClient` is going away.

tylerbrandt:
Any comment on this?

> perhaps they should be modified to expect window.optimizelySdk; comment if you agree/disagree

mikeng13:
> Any comment on this?
> 
> > perhaps they should be modified to expect window.optimizelySdk; comment if you agree/disagree

I agree we should be doing this check in our tests.

mikeng13:
> Hmm I can't figure out how to set it up with webpack/karma. Might have to be an integration-type test. Added #190

Alright we'll take care of it in another PR then.

PR #193, opened 2018-11-20
[OASIS-3799] Add testing for window.optimizelySdk

caitlinrubin-optimizely:
## Summary
We currently don't have any tests for the UMD bundle of the OptimizelySDK.
Changes made:
- Added a new karma config file which requires the local umd bundle and the umd tests
- Added a umd tests file which has all (minus one, see below) the tests from the xbrowser tests but using `window.optimizelySdk`
- Added a new script for running the new tests

Note: There is one test from xbrowser testing I wasn't able to port, `should create an instance of optimizely`. This test basically just checked that the `optlyInstance` was an instance of `Optimizely`  (defined `require('./optimizely')`),which won't work on the instance created through the umd bundle.

## Test plan
- Ran these tests locally, still need to figure out what work is needed to automate this

## Issues
- Addresses [OASIS-3799](https://optimizely.atlassian.net/browse/OASIS-3799)


packages/optimizely-sdk/karma.umd.conf.js
```diff
@@ -0,0 +1,140 @@
+// Karma configuration
+// Generated on Wed May 24 2017 14:10:20 GMT-0700 (PDT)
```
mikeng13: Was this really generated back then? I think we copied this over. We can just note that here instead.

packages/optimizely-sdk/karma.umd.conf.js
```diff
@@ -0,0 +1,140 @@
+// Karma configuration
```
mikeng13: Instead of repeating the same config across two different karma files, we can instead have a shared config that just exports a base configuration with all the stuff that is the same between the karma files and then extend that object in here.

Example:

```
const baseConfig = require('karma.base.conf.js')

module.exports = function(config) {
  config.set({
    // this spread simply assigns all properties from baseConfig to this object (it's ES6 syntax)
    ...baseConfig,   
    
    // list of files / patterns to load in the browser
    files: [
      './dist/optimizely.browser.umd.min.js',
      './lib/index.browser.umdtests.js'
    ]
  })
}
```

packages/optimizely-sdk/lib/index.browser.umdtests.js
```diff
@@ -0,0 +1,244 @@
+/**
+ * Copyright 2016-2017, Optimizely
```
mikeng13: 2018 :)

mikeng13:
Also, merge master into your branch and push.

packages/optimizely-sdk/lib/index.browser.umdtests.js
```diff
@@ -0,0 +1,244 @@
...
+          datafile: {},
+          errorHandler: fakeErrorHandler,
+          eventDispatcher: fakeEventDispatcher,
+          logger: silentLogger,
+        });
+        assert.equal('javascript-sdk', optlyInstance.clientEngine);
+        assert.equal(packageJSON.version, optlyInstance.clientVersion);
+      });
+
+      it('should activate with provided event dispatcher', function(done) {
```
mjc1283: Let's remove `done`. The `done` argument is for asynchronous tests, but this test isn't asynchronous.

PR #198, opened 2018-12-04
[OASIS-3799]  Adding test-ci to prepublish hoook

caitlinrubin-optimizely:
## Summary
- @mjc1283 suggested adding test-ci into the prepublish hook, so I've swapped it out for test-xbrowser there

## Issues
- OASIS-3799

### DIFF
--- a/packages/optimizely-sdk/package.json
+++ b/packages/optimizely-sdk/package.json
@@ -13,7 +13,7 @@
     "lint": "eslint lib/**",
     "cover": "istanbul cover _mocha ./lib/*.tests.js ./lib/**/*.tests.js ./lib/**/**/*tests.js",
     "coveralls": "npm run cover -- --report lcovonly && cat ./coverage/lcov.info | coveralls",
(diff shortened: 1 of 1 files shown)


packages/optimizely-sdk/package.json
```diff
@@ -13,7 +13,7 @@
     "lint": "eslint lib/**",
     "cover": "istanbul cover _mocha ./lib/*.tests.js ./lib/**/*.tests.js ./lib/**/**/*tests.js",
     "coveralls": "npm run cover -- --report lcovonly && cat ./coverage/lcov.info | coveralls",
-    "prepublishOnly": "npm run build-browser-umd && npm test && npm run test-xbrowser"
+    "prepublishOnly": "npm run build-browser-umd && npm test && npm run test-ci"
```
mikeng13: Instead of `test-ci` we should run both `test-xbrowser` and `test-umdbrowser` since prepublish happens on your local dev environment and not in CI.

caitlinrubin-optimizely: 👍 done

PR #200, opened 2018-12-14
fix(decision-service): Adds bucketing id in getVariationForRollout method.

mfahadahmed:
## Summary
- Adds getBucketingId call in getVariationForRollout method to bucket user with provided bucketing id.


mfahadahmed:
Closing this PR as Bucketing ID is not supported in the rollout yet.

aliabbasrizvi:
@mfahadahmed @msohailhussain I see that this was closed with a comment and then re-opened. I do not recall if bucketing ID is supported for rollout, so may be this change is not needed and PR can be closed?

PR #203, opened 2018-12-21
refac(ci): switch stage order

juancarlostong:
## Summary
- switch the order of stages so that integration tests occur ahead of unit tests. this allows resetting the build status to pending without having to wait for unit tests to finish first.


.travis.yml
```diff
@@ -27,6 +27,10 @@ srcclr: true
 script: npm test
 after_success: npm run coveralls
 
+stages:
```
mikeng13: Can you leave a doc comment here explaining its placement?

PR #204, opened 2018-12-26
feat(attribute_value): Don't target NAN, INF, -INF and > 2^53.

mfahadahmed:
Summary
-------
Don't target NAN, INF or -INF doubles
Don't target Values > 2^53. 
Modified fns.isFinite to check if the number is greater than 2^53. 

Test plan
---------

Issues
------

### DIFF
--- a/packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.tests.js
+++ b/packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.tests.js
@@ -200,6 +200,10 @@ describe('lib/core/custom_attribute_condition_evaluator', function() {
         };
         var result = customAttributeEvaluator.evaluate(invalidValueCondition, { lasers_count: 9000 });
         assert.isNull(result);
+
+        invalidValueCondition.value = Math.pow(2, 53) + 2;
+        result = customAttributeEvaluator.evaluate(invalidValueCondition, { lasers_count: 9000 });
+        assert.isNull(result);
       });
     });
 
@@ -305,6 +309,11 @@ describe('lib/core/custom_attribute_condition_evaluator', function() {
         meters_travelled: Infinity,
       });
       assert.isNull(result);
+
+      result = customAttributeEvaluator.evaluate(gtCondition, {
+        meters_travelled: Math.pow(2, 53) + 2,
+      });
+      assert.isNull(result);
     });
 
     it('should return null if there is no user-provided value', function() {
@@ -326,6 +335,11 @@ describe('lib/core/custom_attribute_condition_evaluator', function() {
       invalidValueCondition.value = null;
       result = customAttributeEvaluator.evaluate(invalidValueCondition, userAttributes);
       assert.isNull(result);
+
+      invalidValueCondition.value = Math.pow(2, 53) + 2;
+      result = customAttributeEvaluator.evaluate(invalidValueCondition, userAttributes);
+      assert.isNull(result);
+
     });
   });
 
@@ -366,6 +380,11 @@ describe('lib/core/custom_attribute_condition_evaluator', function() {
         meters_travelled: Infinity,
       });
       assert.isNull(result);
+
+      result = customAttributeEvaluator.evaluate(ltCondition, {
+        meters_travelled: Math.pow(2, 53) + 2,
+      });
+      assert.isNull(result);
     });
 
     it('should return null if there is no user-provided value', function() {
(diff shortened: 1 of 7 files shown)


packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.tests.js
```diff
@@ -200,6 +200,10 @@ describe('lib/core/custom_attribute_condition_evaluator', function() {
         };
         var result = customAttributeEvaluator.evaluate(invalidValueCondition, { lasers_count: 9000 });
         assert.isNull(result);
+
+        invalidValueCondition.value = Math.pow(2, 53) + 2;
```
mikeng13: Seems wasteful that we compute `Math.pow(2, 53)` repeatedly. Let's throw it into our `enums` or `constants` file and reference that instead.

packages/optimizely-sdk/lib/utils/fns/index.js
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 var uuid = require('uuid');
+var finite = require('lodash/isFinite');
```
mikeng13: nit: call this `_isFinite`

packages/optimizely-sdk/lib/utils/attributes_validator/index.js
```diff
@@ -47,6 +48,7 @@ module.exports = {
   },
 
   isAttributeValid: function(attributeKey, attributeValue) {
-    return typeof attributeKey === 'string' && VALID_ATTRIBUTE_TYPES.indexOf(typeof attributeValue) !== -1;
```
nchilada: Looks like we don't need to define `VALID_ATTRIBUTE_TYPES` any more! (At the top of this module)

packages/optimizely-sdk/lib/utils/fns/index.js
```diff
@@ -24,7 +25,9 @@ module.exports = {
   },
   isArray: require('lodash/isArray'),
   isEmpty: require('lodash/isEmpty'),
-  isFinite: require('lodash/isFinite'),
+  isFinite: function(number) {
+    return finite(number) && Math.abs(number) <= Math.pow(2, 53);
```
nchilada:
```suggestion
    return finite(number) && Math.abs(number) <= Number.MAX_SAFE_INTEGER + 1;
```

packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.tests.js
nchilada: `Number.MAX_SAFE_INTEGER + 1` might be the most efficient option here... but maybe we should do that in the code under test while continuing to be explicit in the test code.

packages/optimizely-sdk/lib/core/event_builder/index.tests.js
```diff
@@ -525,9 +510,9 @@ describe('lib/core/event_builder', function() {
 
         var eventOptions = {
           attributes: {
-            'browser_type': 'Chrome',
-            'integer_key': 10,
-            'boolean_key': false,
+            'browser_type': Infinity,
+            'integer_key': NaN,
+            'boolean_key': (Math.pow(2, 53) + 2) * -1,
             'double_key': [1, 2, 3],
```
nchilada: Nit: strange attribute keys 🤔

nchilada: I think we should continue including some valid attributes so that we can make sure they're kept even when other attributes are filtered out? For example
```javascript
'browser_type': 'Chrome',
'valid_positive_number': Math.pow(2, 53),
'valid_negative_number': -Math.pow(2, 53),
'invalid_number': Math.pow(2, 53) + 2,
'array': [1, 2, 3]
```
and then assert that only the first three are accepted

packages/optimizely-sdk/lib/core/event_builder/index.tests.js
```diff
@@ -972,6 +957,65 @@ describe('lib/core/event_builder', function() {
...
+          userId: 'testUser',
+          experimentsToVariationMap: {
+            '111127': '111128',
+            '122227': '122228'
+          },
+          attributes: {
+            'browser_type': -Infinity,
+            'integer_key': NaN,
+            'boolean_key': Math.pow(2, 53) + 2,
+            'double_key': [1, 2, 3],
```
nchilada: Same suggestion - we should include some valid attribute values too

packages/optimizely-sdk/lib/utils/fns/index.js
```diff
@@ -24,7 +25,9 @@ module.exports = {
   },
   isArray: require('lodash/isArray'),
   isEmpty: require('lodash/isEmpty'),
-  isFinite: require('lodash/isFinite'),
+  isFinite: function(number) {
```
nchilada: Consider adding unit tests so that we can try all the special values (NaN, +Infinity, -Infinity), sharpness of bounds (i.e. confirm that 2^53 works and 2+53 + 1 doesn't), and negative values as well.

PR #206, opened 2019-01-09
fix(decision_service): Return null variation if experiment key does not exist in feature's experimentIds array.

mfahadahmed:
## Summary
- Return null variation if experiment key does not exist in feature's experimentIds array.

## Issues
- "OASIS-3950"

### DIFF
--- a/packages/optimizely-sdk/lib/core/decision_service/index.tests.js
+++ b/packages/optimizely-sdk/lib/core/decision_service/index.tests.js
@@ -1,5 +1,5 @@
 /****************************************************************************
- * Copyright 2017-2018, Optimizely, Inc. and contributors                        *
+ * Copyright 2017-2019, Optimizely, Inc. and contributors                        *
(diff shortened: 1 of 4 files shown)


packages/optimizely-sdk/lib/core/decision_service/index.tests.js
```diff
@@ -889,6 +889,18 @@ describe('lib/core/decision_service', function() {
             assert.deepEqual(decision, expectedDecision);
             sinon.assert.calledWithExactly(mockLogger.log, LOG_LEVEL.DEBUG, 'DECISION_SERVICE: User user1 is not in any experiment on the feature feature_with_group.');
           });
+
+          it('bucketed group experiment does not exist in the feature', function() {
```
mikeng13: wording does not conform the BDD pattern. Should be `it(returns null decision for group experiment not referenced by the feature)`

PR #209, opened 2019-01-15
fix(default logger): log each message at the correct level

nchilada:
## Summary

* For messages at INFO level, call `console.info` instead of `console.log`.
* Prepend each log message with the message's own log level instead of prepending with the _threshold_ log level. This is particularly important [for contexts like Node.js](https://nodejs.org/api/console.html) where there is (currently) no differentiation between `console.log` and `console.info` or between `console.warn` and `console.error`.
* Fix the code for the "Setting log level" message, which wasn't showing up at all before. Note that this message will only appear when the threshold level is DEBUG.

## Test plan

* Unit tests.
* Manual testing in my local test app.

## Issues

* OASIS-4021

### DIFF
--- a/packages/optimizely-sdk/lib/plugins/logger/index.js
+++ b/packages/optimizely-sdk/lib/plugins/logger/index.js
@@ -15,6 +15,9 @@
  */
 var fns = require('../../utils/fns');
 var enums = require('../../utils/enums');
+var sprintf = require('sprintf-js').sprintf;
+
+var MODULE_NAME = 'LOGGER';
 
 /**
  * Default logger implementation
@@ -37,9 +40,9 @@ function Logger(config) {
     prefix: '[OPTIMIZELY]',
   }, config);
 
-  this.setLogLevel(config.logLevel);
   this.logToConsole = config.logToConsole;
   this.prefix = config.prefix;
+  this.setLogLevel(config.logLevel);
 }
 
 /**
@@ -50,7 +53,8 @@ function Logger(config) {
 Logger.prototype.log = function(logLevel, logMessage) {
   if (this.__shouldLog(logLevel)) {
     if (this.prefix) {
-      logMessage = this.prefix + ' - ' + this.logLevelName + ' ' + getTime() + ' ' + logMessage;
+      var logLevelName = getLogLevelName(logLevel);
+      logMessage = this.prefix + ' - ' + logLevelName + ' ' + getTime() + ' ' + logMessage;
     }
 
     if (this.logToConsole) {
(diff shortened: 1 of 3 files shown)


packages/optimizely-sdk/lib/plugins/logger/index.js
```diff
@@ -37,9 +40,9 @@ function Logger(config) {
     prefix: '[OPTIMIZELY]',
   }, config);
 
-  this.setLogLevel(config.logLevel);
   this.logToConsole = config.logToConsole;
   this.prefix = config.prefix;
```
nchilada: These properties must be set in order for `this.setLogLevel` to log a message of its own.

packages/optimizely-sdk/lib/plugins/logger/index.js
```diff
@@ -50,7 +53,8 @@ function Logger(config) {
 Logger.prototype.log = function(logLevel, logMessage) {
   if (this.__shouldLog(logLevel)) {
     if (this.prefix) {
-      logMessage = this.prefix + ' - ' + this.logLevelName + ' ' + getTime() + ' ' + logMessage;
+      var logLevelName = getLogLevelName(logLevel);
+      logMessage = this.prefix + ' - ' + logLevelName + ' ' + getTime() + ' ' + logMessage;
```
nchilada: This is the main bugfix!

packages/optimizely-sdk/lib/plugins/logger/index.tests.js
```diff
@@ -38,15 +38,27 @@ describe('lib/plugins/logger', function() {
         sinon.stub(defaultLogger, '__consoleLog');
       });
 
-      it('should log the given message', function() {
+      it('should log a message at the threshold log level', function() {
         defaultLogger.log(LOG_LEVEL.INFO, 'message');
-        assert.isTrue(defaultLogger.__consoleLog.calledOnce);
-        assert.notStrictEqual(defaultLogger.__consoleLog.firstCall.args, [LOG_LEVEL.INFO, ['message']]);
```
nchilada: `notStrictEqual` === `!==` 😨

packages/optimizely-sdk/lib/plugins/logger/index.js
mikeng13: Can you use `sprintf` here instead of manually doing string concatenation

packages/optimizely-sdk/lib/plugins/logger/index.tests.js
mikeng13: 😱

packages/optimizely-sdk/lib/plugins/logger/index.js
```diff
@@ -15,6 +15,9 @@
  */
 var fns = require('../../utils/fns');
 var enums = require('../../utils/enums');
+var sprintf = require('sprintf-js').sprintf;
+
+var MODULE_NAME = 'LOGGER';
```
mikeng13: nit: `DEFAULT_LOGGER`?

nchilada: Ooh, good catch!

nchilada:
Hmm, I guess we were calling the correct `console` method, for what it's worth. That makes a difference in browser dev tools though maybe not in Node.

I'll update the PR description to clarify that (1) the bug is only in our default logger and (2) it relates to the _prefix_ that we use for each message

mikeng13:
@nchilada looks like some unit tests break on Node6

nchilada:
I've just confirmed that `log`/`info`/`warn`/`error` are all [supported](https://nodejs.org/dist/v4.0.0/docs/api/console.html) in Node 4.0, which is the [oldest](https://github.com/optimizely/javascript-sdk/blob/fe6916fc9252c2f8d3e71a1e24595d209fe859f1/packages/optimizely-sdk/README.md#prerequisites) that our SDK currently supports.

nchilada:
@mikeng13 @mjc1283 PR is updated. Can you take another look and maybe merge?

nchilada:
@mikeng13 @mjc1283 bump

PR #213, opened 2019-01-25
fix(event-tags): do not exclude falsy revenue and event values in the event payload

mikeng13:
## Summary
- The revenue and event values were excluded from the event object if they evaluated as falsy.
- The fix is to check against `null` instead of performing a truthy check because the utility method will return `null` if the revenue or event values cannot be parsed from the tags.

## Test plan
- Unit tests and integration tests

### DIFF
--- a/packages/optimizely-sdk/lib/core/event_builder/index.js
+++ b/packages/optimizely-sdk/lib/core/event_builder/index.js
@@ -133,12 +133,12 @@ function getVisitorSnapshot(configObj, eventKey, eventTags, logger) {
 
   if (eventTags) {
     var revenue = eventTagUtils.getRevenueValue(eventTags, logger);
-    if (revenue) {
+    if (revenue !== null) {
       eventDict[enums.RESERVED_EVENT_KEYWORDS.REVENUE] = revenue;
     }
 
     var eventValue = eventTagUtils.getEventValue(eventTags, logger);
-    if (eventValue) {
+    if (eventValue !== null) {
       eventDict[enums.RESERVED_EVENT_KEYWORDS.VALUE] = eventValue;
     }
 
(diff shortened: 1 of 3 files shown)


mikeng13:
It passes @nchilada's tests here: https://travis-ci.com/optimizely/fullstack-sdk-compatibility-suite/builds/98673737

packages/optimizely-sdk/lib/core/event_builder/index.js
```diff
@@ -133,12 +133,12 @@ function getVisitorSnapshot(configObj, eventKey, eventTags, logger) {
 
   if (eventTags) {
     var revenue = eventTagUtils.getRevenueValue(eventTags, logger);
-    if (revenue) {
+    if (revenue !== null) {
```
aliabbasrizvi: Should we check for undefined?

mikeng13: Nah because the `getRevenueValue` function will always return either `null` or the parsed value

nchilada: The _most_ robust (redundant) solution would be to check whether the returned value is a number, but I'm not sure it matters

packages/optimizely-sdk/lib/core/event_builder/index.tests.js
```diff
@@ -1097,6 +1097,54 @@ describe('lib/core/event_builder', function() {
             assert.deepEqual(actualParams, expectedParams);
           });
 
+          it('should include falsy revenue values in the event object', function() {
```
nchilada: Maybe just mention `0` directly? The current test name incorrectly implies that `NaN` and non-numeric falsy values like `''` would _also_ be sent.

PR #215, opened 2019-02-02
fix(srcclr): Update package-lock.json to satisfy SourceClear

nchilada:
## Summary
SourceClear seems to install `lodash@4.17.10` per our package-lock.json, rather than installing `lodash@4.17.11` (the latest) as they might if installing directly via npm. Update package-lock.json to mitigate this.

Preserve `"lodash": "^4.0.0"` in package.json so that our customers remain free to control their lodash subdependencies.

## Issues
- OASIS-4112

### DIFF
--- a/packages/optimizely-sdk/package-lock.json
+++ b/packages/optimizely-sdk/package-lock.json
@@ -78,7 +78,7 @@
       "integrity": "sha512-Mmqx/cS68K1tSrvRLtaV/Lp3NZWzXtOHUW2IvDvl2sihAwJh4ACE0eL6A8FvMyDG9abes3saB6dMimLOs+HMoQ==",
       "dev": true,
       "requires": {
-        "@xtuc/ieee754": "1.2.0"
+        "@xtuc/ieee754": "^1.2.0"
       }
     },
     "@webassemblyjs/leb128": {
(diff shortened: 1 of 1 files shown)


nchilada:
@mikeng13 @spencerwilson-optimizely @mjc1283 does this seem reasonable?

Or should we specify a more recent version of `lodash` in our package.json? That would make it easiest for all customers to install the latest (and presumably safer) subdependency, with the caveats that this would marginally bloat their node_modules/ and would not really make sense unless we were to continually re-release our package with latest dependencies.

nchilada:
I've [confirmed](https://optimizely.sourceclear.io/teams/3OOu2k/projects/65534/vulnerabilities?branch=nikhil/lodash) that this works

nchilada:
Thanks @spencerwilson-optimizely! Yeah, the remaining question is whether we should be doing _more_ and ensuring that we always use a secure version of `lodash` when installed as a dependency. Presumably we haven't cared in the past, given that we only specify `^4.0.0`.

The challenge, of course, is that we'd have to keep re-releasing if our philosophy is to pin to secure packages.

That's a gnarly problem so I'm just going to merge this change for now!

PR #216, opened 2019-02-12
DO NOT REVIEW - SEE https://github.com/optimizely/javascript-sdk/pull/288

jamesopti:
## Summary

This PR adds the ability for an SDK consumer to pass their own set of condition evaluators when creating an optimizely instance.

The core motivator for this is to be able to evaluate WEB audience conditions using the Javascript SDK as part of the EDGE project (go/edge).

Future fullstack use cases for this functionality might include audience condition plugins, where the customer would define their own match types and provide their own matchers.

```js
/**
 * Condition evaluators for conditions that can be evaluated in a Cloudflare worker using
 * information like cookies, query parameters, userAgent, etc...
 */
function getConditionEvaluators(inputs) {
  return {
    cookies: {
      evaluate: matchCookies.match.bind(this, { cookies: inputs.cookies }),
    },
    device: {
      evaluate: matchDevice.match.bind(this, { device: Detect.parseUA(inputs.ua).device }),
    },
    ...
    /* Other custom condition evaluators */
  }
}


var optlyInstance = optimizely.createInstance({
  datafile: datafile,
  conditionEvaluators: getConditionEvaluators(inputs),
  skipJSONValidation: true
});
```

## Test plan

Unit tests would be pretty straightforward to add for this


packages/optimizely-sdk/lib/utils/enums/index.js
```diff
@@ -26,6 +26,7 @@ exports.LOG_LEVEL = {
 };
 
 exports.ERROR_MESSAGES = {
+  AUDIENCE_EVALUATOR_EXPECTED: '%s: No audience evaluator present for condition type: %s.',
```
nchilada: Can you use `exports.LOG_MESSAGES.UNKNOWN_CONDITION_TYPE` instead? We can consider updating it if we want to acknowledge the possibility of custom condition evaluators

nchilada: And I think you can update custom_attribute_condition_evaluator/index.js to stop producing that log message, since we'll be producing it from the general audience evaluator. It has a unit test that we'll want to move as well.

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -15,55 +15,68 @@
...
+ *                                                              are met. If not provided, defaults to an empty object
+ * @param  {Object}                       logger                Logger instance.
+ * @return {Boolean}                                            true if the user attributes match the given audience conditions, false
+ *                                                              otherwise
+ */
+AudienceEvaluator.prototype.evaluate = function(audienceConditions, audiencesById, userAttributes, logger) {
+  // if there are no audiences, return true because that means ALL users are included in the experiment
+  if (!audienceConditions || audienceConditions.length === 0) {
+    return true;
+  }
```
nchilada: Oh dear, you must have been on an _ancient_ version of master if this block was still around 😞 Would you mind re-implementing your changes to this module to make sure we're not reverting any other patterns?

jamesopti: Hmmm, I just rebased my PR off master that you updated on friday.

This block is in master though: https://github.com/optimizely/javascript-sdk/blob/master/packages/optimizely-sdk/lib/core/audience_evaluator/index.js#L42-L44

nchilada: Ah, my bad! I thought this was the original `!userAttributes` check, and failed to double-check amidst the ensuing heart attack 😬

Sorry! I'll review the rest

jamesopti: Haha - Well that check is still there too: https://github.com/optimizely/javascript-sdk/blob/master/packages/optimizely-sdk/lib/core/audience_evaluator/index.js#L46-L48

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -15,55 +15,68 @@
...
-      return true;
-    }
 
-    if (!userAttributes) {
-      userAttributes = {};
-    }
+function AudienceEvaluator(conditionEvaluators) {
+  this.typeToEvaluatorMap = Object.assign({
+    'custom_attribute': customAttributeConditionEvaluator
+  }, conditionEvaluators);
```
nchilada: Do we just want allow for additional condition types beyond those supported by the javascript-sdk, or do we also want to allow customers to modify how the built-in condition types are evaluated?

If the former, I wonder if we can add some kind of namespacing (as in Optimizely Web audience integrations) to protect custom conditions types from conflicting with (existing _and_ future) built-in condition types.

If the latter, I have serious reservations, since that could make it [difficult](https://www.joelonsoftware.com/2008/03/17/martian-headsets/) for us to modify our own condition types without affecting existing customers.

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -15,55 +15,68 @@
...
+  if (!userAttributes) {
+    userAttributes = {};
+  }
 
-      return null;
-    };
+  var evaluateConditionWithUserAttributes = function(condition) {
+    var evaluator = this.typeToEvaluatorMap[condition.type];
+    if (!evaluator) {
+      logger.log(LOG_LEVEL.WARNING, sprintf(ERROR_MESSAGES.AUDIENCE_EVALUATOR_EXPECTED, MODULE_NAME, condition.type));
```
nchilada: We want to move the code that's currently in custom_attribute_condition_evaluator/index.js.
```suggestion
      logger.log(LOG_LEVEL.WARNING, sprintf(LOG_MESSAGES.UNKNOWN_CONDITION_TYPE, MODULE_NAME, JSON.stringify(condition)));
```

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -15,55 +15,68 @@
...
+      return false;
```
nchilada:
```suggestion
      return null;
```

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
nchilada: TODO: additional unit tests, eventually

nchilada: Yikes, we don't seem to catch errors [when constructing the decision service](https://github.com/optimizely/javascript-sdk/blob/de83b9b51d040f8d1f135c22d7a2c1fe3ca97c94/packages/optimizely-sdk/lib/optimizely/index.js#L97-L101). So if we apply any validation at this stage, or do _anything_ that could throw an error, that error will bubble up into the customer's application.

Maybe we should do something about that! Though for the moment, I'm just pointing this out so that we remain careful not to throw.

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -15,55 +15,68 @@
...
-        logger.log(LOG_LEVEL.INFO, sprintf(LOG_MESSAGES.AUDIENCE_EVALUATION_RESULT, MODULE_NAME, audienceId, resultText));
-        return result;
-      }
+  if (!userAttributes) {
+    userAttributes = {};
+  }
 
-      return null;
-    };
+  var evaluateConditionWithUserAttributes = function(condition) {
```
nchilada: Might be a good time to define this as a separate method that makes use of `this.typeToEvaluatorMap`

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
nchilada: True, there is still a `!userAttributes` check, but the original one was much different! We used to `return false;`

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
jamesopti: I like the approach of not allowing the built in evaluators to be overridden. If you need to do that, then you'll be able to write your own.

Good call on the error catching. Added an error message for that which is used in a try/catch now.

mikeng13:
Unit tests aren't passing in Travis, might wanna fix that and also merge `master` in. I will be reviewing this PR in the meantime!

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -16,54 +16,88 @@
...
+ * @return {Boolean|null}              true if the condition is satisfied, null if a matcher is not found.
+ */
+AudienceEvaluator.prototype.evaluateConditionWithUserAttributes = function(userAttributes, condition) {
+  var evaluator = this.typeToEvaluatorMap[condition.type];
+  if (!evaluator) {
+    this.logger.log(LOG_LEVEL.WARNING, sprintf(LOG_MESSAGES.UNKNOWN_CONDITION_TYPE, MODULE_NAME, JSON.stringify(condition)));
+    return null;
+  }
+  try {
+    return evaluator.evaluate(condition, userAttributes, this.logger);
```
mikeng13: if the evaluator we get is not the `custom_attribute`, we should log a debug message to let the user know that. Might be helpful when trying to debug. Maybe we can even just log which evaluator we ended using (by key)

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -16,54 +16,88 @@
...
-      return true;
-    }
 
-    if (!userAttributes) {
-      userAttributes = {};
-    }
+/**
+ * Construct an instance of AudienceEvaluator with a given logger and options
+ * @param {Logger}  logger                           The Logger instance
+ * @param {Object=} __exploratoryConditionEvaluators A map of condition evaluators provided by the consumer. This enables matching
```
mikeng13: What's with this name? Can we call em `customConditionEvaluators`? Also, if we expect it to be overrideable, we make it part of the official public APIs, and thus shouldn't need to prefix with `__`

packages/optimizely-sdk/lib/core/decision_service/index.js
```diff
@@ -52,6 +52,7 @@ function DecisionService(options) {
   this.configObj = options.configObj;
   this.userProfileService = options.userProfileService || null;
   this.logger = options.logger;
+  this.audienceEvaluator = new AudienceEvaluator(options.logger, options.__exploratoryConditionEvaluators);
```
mikeng13: This feels like we are polluting our factory method haha but I see no other easy way to integrate into the current setup.

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
jamesopti: The idea here is to indicate to any prying eyes to "use at your own risk" before this is made publicly available/official.

Until we officially support a way to use different audience condition types in FS audiences, there would be no other reason for a customer to actually use this.

Thoughts?

packages/optimizely-sdk/lib/core/decision_service/index.js
jamesopti: You mean the constructor?

Yea in the current setup, all this state must be threaded through (config, logger instance, etc...)

Would be nice to have a global registration system by instance ID so that this kinda global state can be easily accessed.

packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.js
```diff
@@ -56,7 +56,6 @@ EVALUATORS_BY_MATCH_TYPE[SUBSTRING_MATCH_TYPE] = substringEvaluator;
  */
 function evaluate(condition, userAttributes, logger) {
   if (condition.type !== CUSTOM_ATTRIBUTE_CONDITION_TYPE) {
-    logger.log(LOG_LEVEL.WARNING, sprintf(LOG_MESSAGES.UNKNOWN_CONDITION_TYPE, MODULE_NAME, JSON.stringify(condition)));
```
lpappone: @jamesopti I see this log message now in `AudienceEvaluator.prototype.evaluateConditionWithUserAttributes` - what is the purpose of this evaluate function, now?

lpappone:
DO NOT REVIEW - SEE https://github.com/optimizely/javascript-sdk/pull/288

mikeng13:
Obsolete

PR #220, opened 2019-02-14
Jordan/implement typescript

jordangarcia:
## Summary
- Introduce typescript compiler and use the `allowJs` option
- Move all files from `lib` -> `src`

Immediate benefits are we ensure that any code we're writing is being compiled to `es5` through typescript.  This approach will allow us to gradually introduce typescript.

## Test plan
- All tests pass including doing an `npm publish --dry-run`


jordangarcia:
Going to hold off on this until we have a better reason to bring typescript into the `optimizely-sdk` folder.

PR #225, opened 2019-02-16
Add core logger package

jordangarcia:
## Summary
- Add `logger` `errorHandler` and `utils` as a core package via lerna

This allows us to have other packages, such as the DatafileManager that leverage the LogFactory singleton.

## Test plan
- Unit tests in the core package


jordangarcia:
Opening new PR

PR #226, opened 2019-02-18
Add core logger package

jordangarcia:
## Summary
- Add `js-sdk-logging` package that supplies a consistent logging interface with error handling
- Add `js-sdk-utils` which are only utils, this package is meant to eventually replace the `utils/fns.js` in the `optimizely-sdk` package

This allows us to have other packages, such as the DatafileManager that leverage the LogFactory singleton.

[Design doc for Core Logger Improvements](https://docs.google.com/document/d/1rQ13WS0quy9NreX1xLQVpSHYFiY_oyqcNDISgALVyY8/edit#heading=h.6jynaot9cbnq)

## Test plan
- Unit tests


packages/utils/src/utils.ts
```diff
@@ -0,0 +1,70 @@
+import * as uuid from 'uuid'
```
mjc1283: Several exports of the `utils` module, such as `groupBy` and `objectValues`, are unused. Why are we adding them?

packages/utils/jest.config.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+  // "roots": [
+  //   "./src"
+  // ],
+  "transform": {
+    "^.+\\.tsx?$": "ts-jest"
```
mjc1283: Why `tsx`?

packages/utils/src/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+export function getTimestamp(): number {
+  return new Date().getTime()
+}
+
+export function generateUUID(): string {
+  return uuid.v4()
+}
+
+/**
+ * Valids a valid is a valid TypeScript enum
```
mjc1283: Nit: grammar

packages/utils/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+  })
+
+  describe('objectValues', () => {
+    it('should return object values', () => {
+      expect(objectValues({ foo: 'bar', bar: 123 })).toEqual(['bar', 123])
+    })
+    // TODO test for enumerable properties only
+  })
+
+  describe('objectValues', () => {
```
mjc1283: Two `describe` blocks here for `'objectValues'` - mistake?

packages/logging/test.js
```diff
@@ -0,0 +1,76 @@
+
```
mjc1283: These seem more like code samples or documentation than a test - should these live elsewhere?

packages/logging/src/logger.ts
```diff
@@ -0,0 +1,309 @@
...
+  ERROR: 4,
+}
+
+function coerceLogLevel(level: any): LogLevel {
+  if (typeof level !== 'string') {
+    return level
+  }
+
+  level = level.toUpperCase()
+  if (level === 'WARN') {
```
mjc1283: I see how this coercion is nice for usability, but it seems weird that we are treating `'WARN'` as a special case. Why `'WARN`' and not also `'ERR'`?

packages/logging/src/models.ts
```diff
@@ -0,0 +1,27 @@
...
+  info(message: string | Error, ...splat: any[]): void
+
+  debug(message: string | Error, ...splat: any[]): void
+
+  warn(message: string | Error, ...splat: any[]): void
+
+  error(message: string | Error, ...splat: any[]): void
+}
+
+export interface LogManager {
```
mjc1283: If there is only one class implementing `LogManager`, and there only will ever be one, what do we gain by declaring this interface?

packages/logging/src/logger.ts
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  if (!stringToLogLevel[level]) {
+    return level
+  }
+
+  return stringToLogLevel[level]
+}
+
+type LogData = {
```
mjc1283: Why is `LogData` declared as a type alias, as opposed to an interface? Same question also for other uses of `type` in this module.

packages/logging/src/logger.ts
```diff
@@ -0,0 +1,309 @@
...
+  private internalLog(level: LogLevel, data: LogData): void {
+    if (!globalLogHandler) {
+      return
+    }
+
+    if (level < globalLogLevel) {
+      return
+    }
+
+    // TODO turn this into a format log function
```
mjc1283: What does it mean to turn this into a format log function?

packages/logging/__tests__/logger.spec.ts
```diff
@@ -0,0 +1,386 @@
...
+      logger = getLogger()
+    })
+
+    describe('setLogLevel', () => {
+      it('should coerce "debug"', () => {
+        setLogLevel('debug')
+        expect(getLogLevel()).toBe(LogLevel.DEBUG)
+      })
+
+      it('should coerce "deBug"', () => {
```
mjc1283: Nit: test doesn't match description

packages/logging/src/logger.ts
jordangarcia: `warn` is the standard.  for instance `console.warn`.

packages/logging/src/logger.ts
jordangarcia: It's not going to be extended or reused outside of this file.  My general rule of thumb is to use interfaces if they are going to be extended.

packages/logging/src/logger.ts
jordangarcia: it's already done, will remove the TODO

packages/logging/src/models.ts
jordangarcia: I think the idea of making  `LogManager` something replaceable is something we might want to do in the future.

Also the three main components are `LogManager` `LogHandler` and `LoggerFacade` so it seemed like good documentation to have them here.

packages/logging/test.js
jordangarcia: illremove

packages/utils/__tests__/utils.spec.ts
jordangarcia: will remove

packages/utils/jest.config.js
jordangarcia: this is jsut copy/pasted

Doesn't hurt to have thsi in here though

packages/utils/src/utils.ts
jordangarcia: They are used in other packages that we will use in the near future.

packages/utils/src/utils.ts
jordangarcia: hahahahaha

packages/logging/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+ "extends": "../../tsconfig.json",
+ "compilerOptions": {
+   "outDir": "./lib",
+   "noImplicitAny": true,
```
mjc1283: Don't need `noImplicitAny` and `strictNullChecks` here - same as the parent config.

tsconfig.json
```diff
@@ -0,0 +1,64 @@
...
+    // "removeComments": true,                /* Do not emit comments to output. */
+    // "noEmit": true,                        /* Do not emit outputs. */
+    "suppressImplicitAnyIndexErrors": true,
+    "importHelpers": true /* Import emit helpers from 'tslib'. */,
+    // "downlevelIteration": true,            /* Provide full support for iterables in 'for-of', spread, and destructuring when targeting 'ES5' or 'ES3'. */
+    // "isolatedModules": true,               /* Transpile each file as a separate module (similar to 'ts.transpileModule'). */
+
+    /* Strict Type-Checking Options */
+    "strict": true /* Enable all strict type-checking options. */,
+    "noImplicitAny": true /* Raise error on expressions and declarations with an implied 'any' type. */,
```
mjc1283: I think if `"strict"` is `true` it has the same effect of setting all of `noImplicityAny`, `strictNullChecks`, etc. (everything listed here under Strict Type-Checking Options) to true.

PR #227, opened 2019-02-20
(WIP) Datafile management

mjc1283:
## Summary
Prototype of datafile management. For review only.


packages/datafile-management/src/browser_datafile_manager.ts
```diff
@@ -0,0 +1,87 @@
...
+  return (Date.now() - cacheEntry.timestamp) < MAX_CACHE_AGE_MS
+}
+
+class BrowserDatafileManager extends DefaultDatafileManager {
+  // TODO: Better error handling, reject reasons/messages
+  protected fetchDatafile(datafileUrl: string): Promise<string> {
+    return new Promise((resolve, reject) => {
+      let resolvedFromCache = false
+      const cacheEntryStr = window.localStorage.getItem(STORAGE_KEY)
+      if (cacheEntryStr !== null) {
```
jordangarcia: this seems too strict,

if `cacheEntryStr` is `""` or `undefined` what should happen?

packages/datafile-management/src/browser_datafile_manager.ts
```diff
@@ -0,0 +1,87 @@
...
+        let cacheEntry: CacheEntry | undefined
+        try {
+          // TODO: JSON.parse bull shit (fix this)
+          cacheEntry = JSON.parse(cacheEntryStr)
+        } catch(e) {
+        }
+        if (typeof cacheEntry !== 'undefined' && isFreshAndValid(cacheEntry)) {
```
jordangarcia: [nit] i generally prefer `cacheEntry !== undefined` if we know the variable in question is initialized

packages/datafile-management/src/node_datafile_manager.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      req.on('error', () => {
+        // TODO: Handle error
+      })
+
+      req.end()
+    })
+  }
+
+  protected setInterval(listener: Interval.IntervalListener, intervalMs: number) {
```
jordangarcia: seems like this isnt' used.

packages/datafile-management/src/node_datafile_manager.ts
```diff
@@ -0,0 +1,85 @@
...
+      clearTimeout(timeout)
+    }
+  }
+}
+
+// TODO: argument options type should be Partial (or something's not quite right about the interfaces here)
+export default function create(options: DatafileManagerConfig): DefaultDatafileManager {
+  return new NodeDatafileManager({
+    ...options,
+    liveUpdates: true,
```
jordangarcia: shouldn't we allow this to be configurable.

Seems weird that we are putting this config after the options passed in

packages/datafile-management/src/optimizely_with_managed_datafile.ts
```diff
@@ -0,0 +1,248 @@
...
+    this.client = createDefaultClient()
+
+    if (sdkKey) {
+      // TODO: Provide ability to pass through datafile manager options
+      this.datafileManager = createDatafileManager({ sdkKey, datafile })
+    } else if (datafile) {
+      this.datafileManager = createStaticDatafileManager(datafile)
+    } else {
+      // TODO: Log? Should throw earlier?
+      // As it stands, we will have a default client, isValidInstance will be true, but onReady will be rejected
```
jordangarcia: We should have logging / errorHandler.

But this is basically what happens today in `createInstance` if an invalid datafile is supplied.

jordangarcia:
A few nits, but this looks really good!

PR #228, opened 2019-02-20
Prepare for 3.0.1 release

jordangarcia:
## Summary
- Expose all of the `enums` `plugins` via the top level object

```
const optimizelySdk = require('@optimizely/optimizely-sdk')
const logger = optimizelySdk.logging.createLogger({
  logLevel: optimizelySdk.enums.LOG_LEVEL.DEBUG,
});

const errorHandler = optimizelySdk.errorHandler
const eventDispatcher = optimizelySdk.eventDispatcher
```

We plan to deprecate direct requires of files (which is a really bad practice) in 4.0.  This gives customers time to change how they are requiring plugins / enums in the mean time.

## Test plan
- Unit tests

## Issues
- OASIS-4189

### DIFF
--- a/packages/optimizely-sdk/CHANGELOG.MD
+++ b/packages/optimizely-sdk/CHANGELOG.MD
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 Changes that have landed but are not yet released.
 
+## [3.0.1] - February 15, 2019
+
+### Changed
+- Expose default `loggers`, `errorHandlers`, `eventDispatcher` and `enums` on top level require.
+- `createLogger` and `createNoOpLogger` are available as methods on `optimizelySdk.logging`
+- Added `optimizelySdk.errorHandler`
+- Added `optimizelySdk.eventDispatcher`
+- Added `optimizelySdk.enums`
+
 ## [3.0.0] - February 13, 2019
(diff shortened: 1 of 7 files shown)


packages/optimizely-sdk/CHANGELOG.MD
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 Changes that have landed but are not yet released.
 
+## [3.0.1] - February 15, 2019
```
mikeng13: I wish I could go back in time too!

packages/optimizely-sdk/lib/index.browser.js
```diff
@@ -29,7 +29,7 @@ var MODULE_NAME = 'INDEX';
  * Entry point into the Optimizely Browser SDK
  */
 module.exports = {
-  logger: logger,
+  logging: logger,
```
mikeng13: I know this is tedious, but these changes should be made in a separate commit from the commit that prepares for the release. It is bad practice to bundle them together as it makes for discoverability and (self) documentation hard.

packages/optimizely-sdk/CHANGELOG.MD
jordangarcia: lol dates are hard will fix

PR #231, opened 2019-02-22
Jordan/optimize utils

jordangarcia:
## Summary
- Implement a homegrown `sprintf` function in`utils`.  This will save us over 1kb gzipped in the final build.  It's also twice as fast as the `sprintf-js` package
- Add license headers to all source files in `js-sdk-logging` and `js-sdk-utils`


packages/utils/src/index.ts
```diff
@@ -1 +1,100 @@
...
+export function keyBy<K>(arr: K[], keyByFn: (item: K) => string): { [key: string]: K } {
+  let map: { [key: string]: K } = {}
+  arr.forEach(item => {
+    const key = keyByFn(item)
+    map[key] = item
+  })
+  return map
+}
+
+export function sprintf(format: string, ...args: any[]): string {
```
mjc1283: Should we add docstring comments for all the functions in this module?

jordangarcia: honeslty i dont see a ton of value in adding doc strings to simple typescript functions.

PR #232, opened 2019-02-22
Implement js-sdk-logging in optimizely-sdk

jordangarcia:
## Summary
- Inject the new LoggerFacade and use as the logger
- Change createLogger() and createNoopLogger()
- Update tests to work with new logger

## Test plan
- Existing unit tests
- I've compared log outputs from a stock `3.0.1` `optimizely-sdk` and this one to verify they work the same.


packages/optimizely-sdk/lib/index.browser.js
```diff
@@ -47,43 +51,52 @@ module.exports = {
...
-          if (config.logger) {
-            config.logger.log(enums.LOG_LEVEL.ERROR, errorMessage);
-          } else {
-            defaultLogger.log(enums.LOG_LEVEL.ERROR, errorMessage);
-          }
-          config.isValidInstance = false;
-        }
+      config = config || {};
+
+      // TODO warn about setting per instance errorHandler / logger / logLevel
```
mjc1283: Should we just do this now?

packages/optimizely-sdk/lib/plugins/logger/index.js
```diff
@@ -13,137 +13,17 @@
...
-var enums = require('../../utils/enums');
-var sprintf = require('sprintf-js').sprintf;
+var logging = require('@optimizely/js-sdk-logging');
 
-var MODULE_NAME = 'DEFAULT_LOGGER';
-
-/**
- * Default logger implementation
- */
 function NoOpLogger() {}
```
mjc1283: Is `NoOpLogger` needed anymore?

packages/optimizely-sdk/lib/index.browser.js
```diff
@@ -47,43 +51,52 @@ module.exports = {
...
-      }, config);
+      config = fns.assignIn(
+        {
+          eventDispatcher: defaultEventDispatcher,
+        },
+        config,
+        {
+          clientEngine: enums.JAVASCRIPT_CLIENT_ENGINE,
+          // always get the OptimizelyLogger facade from logging
+          logger: logger,
```
mjc1283: We are still passing a `logger` into `Optimizely`. I thought the plan was to stop passing `logger` around and instead import the singleton wherever necessary. Is that still true/would we implement this in follow-up PRs?

jordangarcia: I left the internal logger passing in place since it reduces the risk of this breaking anything.

As we rewrite things we should be using the singleton.  I didn't see any benefit in doing that now, best case scenario things work exactly as before, worse case scenario the SDK breaks.

packages/optimizely-sdk/lib/index.browser.js
jordangarcia: Yeah I can do that.

packages/optimizely-sdk/lib/plugins/logger/index.js
jordangarcia: I kept it around.  The preferred way of setting no-op logging is to just call `optimizelySDK.setLogger(null)`.

But people are using `NoOpLogger` today, so we need to keep it in.

packages/optimizely-sdk/lib/index.node.js
```diff
@@ -48,31 +51,56 @@ module.exports = {
...
         config.isValidInstance = true;
       } catch (ex) {
-        if (config.logger) {
-          config.logger.log(enums.LOG_LEVEL.ERROR, sprintf('%s: %s', MODULE_NAME, ex.message));
+        if (hasLogger) {
+          logger.error(ex);
         } else {
-          var simpleLogger = logger.createLogger({logLevel: 4});
-          simpleLogger.log(enums.LOG_LEVEL.ERROR, sprintf('%s: %s', MODULE_NAME, ex.message));
+          console.error(ex.message);
```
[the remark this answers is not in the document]
jordangarcia: I didn't quite see the point in creating a `ConsoleLogHandler` just to call it once.  I feel like we shouldn't be using the `LogHandlers` now and only interacting with the Facade.  

The result of this whole thing was just to `console.error` something.  So why not just `console.error` something directly.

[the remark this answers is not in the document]
mjc1283: Sounds good, let's merge it.

PR #233, opened 2019-02-23
Implement event processor

jordangarcia:


jordangarcia:
Going to close and re-open with rebased diff.

PR #235, opened 2019-02-26
feat(isFeatureEnabled): Adds notification listener for isFeatureEnabled API

mfahadahmed:
## Summary
- Added notification listener for is_feature_enabled API

## Test plan
- Added unit tests for is_feature_enabled listener


packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -517,6 +526,50 @@ Optimizely.prototype.getEnabledFeatures = function (userId, attributes) {
...
+ * Check if the feature is enabled for the given user.
+ * @param {string} featureKey   Key of feature which will be checked
+ * @param {string} userId       ID of user which will be checked
+ * @param {Object} attributes   Optional user attributes
+ * @return {Object}             Object containing information about feature enabled,
+ * decision source and impression event.
+ */
+Optimizely.prototype._isFeatureEnabledForUser = function (featureKey, userId, attributes) {
+  var feature = projectConfig.getFeatureFromKey(this.configObj, featureKey, this.logger);
+    if (!feature) {
```
jordangarcia: fix this spacing / formatting

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -517,6 +526,50 @@ Optimizely.prototype.getEnabledFeatures = function (userId, attributes) {
...
+    var impressionEvent = null;
+    var decision = this.decisionService.getVariationForFeature(feature, userId, attributes);
+    var variation = decision.variation;
+    if (!!variation) {
+      if (decision.decisionSource === DECISION_SOURCES.EXPERIMENT) {
+        var experimentKey = decision.experiment.key;
+        var variationKey = variation.key;
+        
+        // got a variation from the exp, so we track the impression
+        impressionEvent = this._createImpressionEvent(experimentKey, variationKey, userId, attributes);
```
jordangarcia: This function shouldn't be sending the impression event.  Why does this function sendImpressionEvent but `isFeatureEnabled` calls the notification listener.  They should happen in the same place.

Remove the functionality for sending the impression event to `isFeatureEnabled`

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -517,6 +526,50 @@ Optimizely.prototype.getEnabledFeatures = function (userId, attributes) {
...
+/**
+ * Check if the feature is enabled for the given user.
+ * @param {string} featureKey   Key of feature which will be checked
+ * @param {string} userId       ID of user which will be checked
+ * @param {Object} attributes   Optional user attributes
+ * @return {Object}             Object containing information about feature enabled,
+ * decision source and impression event.
+ */
+Optimizely.prototype._isFeatureEnabledForUser = function (featureKey, userId, attributes) {
+  var feature = projectConfig.getFeatureFromKey(this.configObj, featureKey, this.logger);
```
jordangarcia: Why is the calling function `isFeatureEnabled` responsible for validating inputs but not for resolving the feature by key.

I think this function should take in `feature, userId, attributes` and the `getFeatureFromKey` should happen in the calling function

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -137,7 +137,8 @@ Optimizely.prototype.activate = function (experimentKey, userId, attributes) {
         return variationKey;
       }
 
-      this._sendImpressionEvent(experimentKey, variationKey, userId, attributes);
+      var impressionEvent = this._createImpressionEvent(experimentKey, variationKey, userId, attributes);
+      this._sendImpressionEvent(experimentKey, variationKey, userId, attributes, impressionEvent);
```
msohailhussain: Why other parameters are need to send to _sendImpressionEvent?

packages/optimizely-sdk/lib/optimizely/index.js
msohailhussain: Yes, when we will be implementing getEnabledFeatures, this will cause problem. 
Notifications & Sending Events must be on the root function.

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -517,6 +526,50 @@ Optimizely.prototype.getEnabledFeatures = function (userId, attributes) {
...
+        this.logger.log(LOG_LEVEL.INFO, sprintf(LOG_MESSAGES.FEATURE_ENABLED_FOR_USER, MODULE_NAME, featureKey, userId));
+        result = true;
+      }
+    }
+
+    if (!result) {
+      this.logger.log(LOG_LEVEL.INFO, sprintf(LOG_MESSAGES.FEATURE_NOT_ENABLED_FOR_USER, MODULE_NAME, featureKey, userId));
+    }
+
+    return {
```
msohailhussain: It would be better if you just send decision. Not prepare payload to send for notification here.
This method only return `decision`

packages/optimizely-sdk/lib/optimizely/index.js
mfahadahmed: The existing implementation of getEnabledFeatures sends impression event. The suggested change will refrain getEnabledFeatures from sending that event.

mfahadahmed:
Closing this PR due to the change in implementation.

PR #238, opened 2019-03-04
Prepare 3.1.0-beta1 release

jordangarcia:
## Summary
- New APIs for setting `logger` and `logLevel` on the optimizelySDK singleton
- `logger` and `logLevel` are now set globally for all instances of Optimizely.  If you were passing
different loggers to individual instances of Optimizely, logging behavior may now be different.

## Test plan
- Existing unit + compat tests

## Issues
- OASIS-4297

### DIFF
--- a/packages/optimizely-sdk/CHANGELOG.MD
+++ b/packages/optimizely-sdk/CHANGELOG.MD
@@ -7,6 +7,36 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 Changes that have landed but are not yet released.
 
+## [3.1.0-beta1] - March 5th, 2019
+
+### Changed
+
+- New APIs for setting `logger` and `logLevel` on the optimizelySDK singleton
+- `logger` and `logLevel` are now set globally for all instances of Optimizely.  If you were passing
+different loggers to individual instances of Optimizely, logging behavior may now be different.
+
+#### Setting a ConsoleLogger
+
+```js
+var optimizelySDK = require('@optimizely/optimizely-sdk')
+
+// logger and logLevel are now set on the optimizelySDK singleton
+optimizelySDK.setLogger(optimizelySDK.logging.createLogger())
+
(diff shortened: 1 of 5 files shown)


packages/optimizely-sdk/CHANGELOG.MD
```diff
@@ -7,6 +7,36 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 Changes that have landed but are not yet released.
 
+## [3.1.0-beta1] - March 5th, 2019
+
+### Changed
+
+- New APIs for setting `logger` and `logLevel` on the optimizelySDK singleton
```
mjc1283: Can we link to PRs for each change log item as we've done previously?

packages/optimizely-sdk/CHANGELOG.MD
```diff
@@ -7,6 +7,36 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+- `logger` and `logLevel` are now set globally for all instances of Optimizely.  If you were passing
+different loggers to individual instances of Optimizely, logging behavior may now be different.
+
+#### Setting a ConsoleLogger
```
mjc1283: Do you think we should stick with the standard sections of Changed, New Features, Breaking Changes, or is it OK to introduce these examples here? If we had a link to the logging package README, it might solve the same need while keeping the changelog format the same.

jordangarcia: the logging package isn't great for external documentation because it's the internal interfaces, not what we expose through the SDK.

packages/optimizely-sdk/CHANGELOG.MD
jordangarcia: Sure i think it's just one PR though

packages/optimizely-sdk/CHANGELOG.MD
jordangarcia: Also i personally like the examples.  I think it's the best way to convey the idea to a lot of people.

PR #239, opened 2019-03-04
Add event-processor package

jordangarcia:
## Summary
- This adds the `js-sdk-event-processor` packages to the main repo.

Integration into the `optimizely-sdk` package will occur in a later PR.

## Test plan
- Unit tests


packages/event-processor/src/events.ts
```diff
@@ -0,0 +1,206 @@
...
+  type: 'impression'
+
+  user: {
+    id: string
+    attributes: VisitorAttribute[]
+  }
+
+  layer: {
+    id: string
+  } | null
```
jordangarcia: This shouldn't be null ever really.

Do we want to use campaign here?

packages/event-processor/src/events.ts
```diff
@@ -0,0 +1,206 @@
...
+      anonymizeIP: projectConfig.getAnonymizeIP(),
+      botFiltering: projectConfig.getBotFiltering(),
+    },
+
+    event: {
+      id: event.id,
+      key: event.key,
+    },
+
+    revenue: getRevenueValue(eventTags),
```
jordangarcia: Should we just let strings pass through here?

jordangarcia: use regex to verify that it's a number only.

always send as a string

packages/event-processor/src/events.ts
```diff
@@ -0,0 +1,206 @@
...
+  revenue: number | null
+  value: number | null
+  tags: EventTags
+}
+
+export type EventTags = {
+  [key: string]: string | number | null
+}
+
+export function buildConversionEvent(opts: {
```
jordangarcia: this appears to be unused in the integration of event-processor

packages/event-processor/src/events.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  const parsed = parseInt(eventTags[EVENT_REVENUE_KEY] as string, 10)
+  if (isNaN(parsed)) {
+    logger.warn(
+      'Failed to parse revenue value "%s" from event tags.',
+      eventTags[EVENT_REVENUE_KEY],
+    )
+    return null
+  }
+  return parsed
```
jordangarcia: Dont return the parsed value here, return the unparsed (could be string)

packages/event-processor/src/events.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  const parsed = parseFloat(eventTags[EVENT_VALUE_KEY] as string)
+  if (isNaN(parsed)) {
+    logger.warn(
+      'Failed to parse event value "%s" from event tags.',
+      eventTags[EVENT_VALUE_KEY],
+    )
+    return null
+  }
+  return parsed
```
jordangarcia: same as above, return unparsed.

packages/event-processor/src/v1/buildEventV1.ts
```diff
@@ -0,0 +1,231 @@
...
+      let visitor = makeVisitor(event)
+
+      if (event.type === 'impression') {
+        visitor.snapshots.push(makeDecisionSnapshot(event))
+      } else if (event.type === 'conversion') {
+        visitor.snapshots.push(makeConversionSnapshot(event))
+      }
+
+      visitors.push(visitor)
+    } else if (event.type === 'summary') {
```
jordangarcia: remove summary stuff for now.

packages/event-processor/src/v1/buildEventV1.ts
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/**
+ * Event for usage with v1 logtier
+ *
+ * @export
+ * @interface EventBuilderV1
+ */
+
+export function buildImpressionEventV1(data: ImpressionEvent): EventV1 {
```
jordangarcia: check if this is unused and remove.

packages/event-processor/src/eventProcessor.ts
```diff
@@ -0,0 +1,145 @@
...
+    eventQueueFactory?: EventQueueFactory<ProcessableEvents>
+    transformers?: EventTransformer[]
+    interceptors?: EventInterceptor[]
+    callbacks?: EventCallback[]
+    flushInterval?: number
+    maxQueueSize?: number
+  }) {
+    this.dispatcher = dispatcher
+
+    const factory = eventQueueFactory || new DefaultEventQueueFactory()
```
jordangarcia: do we need the queue to be replaceable

packages/event-processor/src/eventQueue.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  stop(): Promise<any> {
+    const result = this.sink(this.buffer)
+    this.timer.stop()
+    return result
+  }
+
+  enqueue(event: K): void {
+    this.buffer.push(event)
```
jordangarcia: start the timer only after the first event is put in

packages/event-processor/src/eventQueue.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    if (this.buffer.length >= this.maxQueueSize) {
+      this.flush()
+      this.timer.refresh()
```
jordangarcia: This shouldn't refresh if the timer if `maxQueueSize` is 1

packages/event-processor/src/eventProcessor.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    this.transformers = transformers
+    this.interceptors = interceptors
+    this.callbacks = callbacks
+
+    this.queue = factory.createEventQueue({
```
jordangarcia: Swap out the queue if `maxQueueSize = 1`

packages/event-processor/src/eventProcessor.ts
```diff
@@ -0,0 +1,145 @@
...
+
+          resolve()
+        })
+      })
+    })
+
+    return Promise.all(promises)
+  }
+
+  // TODO pass 2nd argument EventMeta /w ProjectConfig
```
jordangarcia: remove this comment

packages/event-processor/src/eventProcessor.ts
```diff
@@ -0,0 +1,145 @@
...
+  async process(event: ProcessableEvents, projectConfig: ProjectConfig): Promise<void> {
+    // loop and apply all transformers
+    for (let transformer of this.transformers) {
+      try {
+        await transformer(event, projectConfig)
+      } catch (ex) {
+        logger.error('eventTransformer threw error', ex.message, ex)
```
jordangarcia: should we actually stop the event here.  

Could the event be malformed, is it riskier than not sending?

packages/event-processor/src/eventProcessor.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    return Promise.all(promises)
+  }
+
+  // TODO pass 2nd argument EventMeta /w ProjectConfig
+  async process(event: ProcessableEvents, projectConfig: ProjectConfig): Promise<void> {
+    // loop and apply all transformers
+    for (let transformer of this.transformers) {
+      try {
+        await transformer(event, projectConfig)
```
jordangarcia: Should we give them a wrapper / interface to do mutation on here

What are the things they'd want to change:
1. set user attribute
2. set event

packages/event-processor/src/eventProcessor.ts
```diff
@@ -0,0 +1,145 @@
...
+    // loop and apply all interceptors
+    for (let interceptor of this.interceptors) {
+      let result
+      try {
+        result = await interceptor(event, projectConfig)
+      } catch (ex) {
+        logger.error('eventInterceptor threw error', ex.message, ex)
+        return
+      }
+      if (result === null) {
```
jordangarcia: maybe this should be a boolean?

packages/event-processor/src/eventProcessor.ts
```diff
@@ -0,0 +1,145 @@
...
+        return
+      }
+    }
+    Object.freeze(event)
+
+    // loop and apply all interceptors
+    for (let interceptor of this.interceptors) {
+      let result
+      try {
+        result = await interceptor(event, projectConfig)
```
jordangarcia: turn this into a boolean.

packages/event-processor/src/eventProcessor.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    this.queue.enqueue(event)
+  }
+
+  stop(): Promise<any> {
+    try {
+      // swallow, an error stopping this queue should prevent this from stopping
+      return this.queue.stop()
+    } catch (e) {
+      // TODO error
```
jordangarcia: `logging.error()`

packages/event-processor/src/eventProcessor.ts
```diff
@@ -0,0 +1,145 @@
...
+              })
+            })
+          })
+
+          resolve()
+        })
+      })
+    })
+
+    return Promise.all(promises)
```
jordangarcia: add test case for this being called when there is nothing in the buffer.

packages/event-processor/src/events.ts
jordangarcia: move all of these functions out to `decision-service` for now

packages/event-processor/__tests__/eventQueue.spec.ts
```diff
@@ -0,0 +1,217 @@
+/// <reference types="jest" />
+
+import { DefaultEventQueue, SingleEventQueue } from '../src/eventQueue'
+
+describe('eventQueue', () => {
+  beforeEach(() => {
+    jest.useFakeTimers()
```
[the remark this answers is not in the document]
jordangarcia: Ah good catch.  It's not actually using the real timers ever here.

For the test it doesn't really need to use real timers.

I'm not sure how jest works between running different files (if they are run in the same global context).  But i'll add a `useRealTimers` in `afterEach` here.

packages/models/src/models.ts
```diff
@@ -0,0 +1,459 @@
...
+  experimentIds: string[]
+  // return the experimentId or null
+  trafficAllocation: TrafficAllocation<string>
+}
+
+type UserVariationMap = {
+  [userId: string]: string // variationId
+}
+
+type ExperimentStatus = 'Running' | 'Launched' | 'Paused' | 'Not started'
```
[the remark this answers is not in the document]
jordangarcia: I dont have a great answer to this, they both have drawbacks.  

An enum is probably better here, as it's easier to not make a typing mistake when comparing it

packages/event-processor/src/eventQueue.ts
```diff
@@ -0,0 +1,115 @@
...
+    this.buffer = []
+    this.maxQueueSize = Math.max(maxQueueSize, 1)
+    this.sink = sink
+    this.timer = new Timer({
+      callback: this.flush.bind(this),
+      timeout: flushInterval,
+    })
+  }
+
+  start(): void {
```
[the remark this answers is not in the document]
jordangarcia: It implements `Managed`.  Yes in this instance having a `start` method is pointless.  But for us to use the type `Managed` and have other parts of the system only know about `Managed` it needs to implement a no-op

packages/event-processor/src/eventProcessor.ts
```diff
@@ -0,0 +1,146 @@
...
+export type ProcessableEvents = ConversionEvent | ImpressionEvent
+
+export type EventDispatchResult = { result: boolean; event: ProcessableEvents }
+
+export type EventCallback = (result: EventDispatchResult) => void
+
+export type EventTransformer = (
+  event: ProcessableEvents,
+  projectConfig: ProjectConfig,
+) => Promise<any>
```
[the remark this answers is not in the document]
jordangarcia: Probably should. 

Once it's compiled this wont be enforced at all, but in TypeScript land you probably shouldn't be returning anything here.

packages/event-processor/__tests__/v1EventProcessor.spec.ts
```diff
@@ -0,0 +1,585 @@
...
+      processor.process(impressionEvent, testProjectConfig)
+
+      processor.stop().then(() => {
+        done()
+      })
+
+      localCallback(true)
+    })
+
+    it('should return a promise that is resolved when the dispatcher callback fires false', done => {
```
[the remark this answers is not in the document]
jordangarcia: this is testing the `processor.stop()` functionality.

This test is saying that even if the request fails to send but the `dispatcher` yielded control back, then the `.stop()` promise should be resolved

packages/event-processor/src/eventDispatcher.ts
```diff
@@ -0,0 +1,18 @@
...
+export interface EventV1Request {
+  url: string
+  method: 'POST' | 'PUT' | 'GET' | 'PATCH'
+  headers: {
+    [key: string]: string[]
+  }
+  event: EventV1,
+}
+
+export interface HttpEventDispatcher extends EventDispatcher {
```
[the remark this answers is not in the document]
jordangarcia: It's ensuring this interface defined in `HttpEventDispatcher` is still a valid `EventDispatcher`

PR #244, opened 2019-03-08
feat(api): Decision Notification Listener for activate and getVariation APIs.

mfahadahmed:
## Summary
- Implemented onDecisionListener in activate and getVariation APIs.

## Test plan
- Added unit tests.


packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -392,14 +395,30 @@ describe('lib/optimizely', function() {
     });
 
     describe('#activate', function() {
-      it('should call bucketer and dispatchEvent with proper args and return variation key', function() {
+      it('should call bucketer, send notifications and dispatchEvent with proper args and return variation key', function() {
```
jordangarcia: lets split this test up, please add a new `#activate` for these notification center tests.

 it's important to keep the base case of these tests without notificationCenter.

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -2519,6 +2520,80 @@ describe('lib/optimizely', function() {
...
+
+        describe('getVariation', function() {
+          it('should send notification with actual variation key when getVariation returns variation', function() {
+            bucketStub.returns('111129');
+            var variation = optlyInstance.getVariation('testExperiment', 'testUser');
+            assert.strictEqual(variation, 'variation');
+            sinon.assert.calledWith(onDecisionListener, {
+              type: DECISION_INFO_TYPES.EXPERIMENT,
+              userId: 'testUser',
+              attributes: undefined,
```
jordangarcia: this listener should be called with `attributes: {}` even if no attributes are provided.

Please change the logic where you send the notification

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -304,7 +306,21 @@ Optimizely.prototype.getVariation = function(experimentKey, userId, attributes)
...
       }
 
-      return this.decisionService.getVariation(experimentKey, userId, attributes);
+      var variationKey = this.decisionService.getVariation(experimentKey, userId, attributes);
+      this.notificationCenter.sendNotifications(
+        NOTIFICATION_TYPES.ON_DECISION,
+        {
+          type: DECISION_INFO_TYPES.EXPERIMENT,
+          userId: userId,
+          attributes: attributes,
```
jordangarcia: linking to comment below, this should default to empty object if `attribute` isn't passed through

```
attributes: attributes || {},
```

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -2287,7 +2288,7 @@ describe('lib/optimizely', function() {
           browser_type: 'firefox',
         };
         optlyInstance.notificationCenter.addNotificationListener(
-          enums.NOTIFICATION_TYPES.ACTIVATE,
```
jordangarcia: Also, why make this change.  It seems we are pretty consistent about using `enums.XXX`

Please change it back since this seems unnecessary and not the conventional way we access enums.

packages/optimizely-sdk/lib/optimizely/index.js
mfahadahmed: Done for decision listener. I just want to confirm that should we change it for activate and track listeners also?

msohailhussain: I confirmed with Ali, for Decision Listener it's fine.

aliabbasrizvi:
LGTM. @jordangarcia may want to take another look.

packages/optimizely-sdk/lib/utils/enums/index.js
```diff
@@ -173,10 +173,23 @@ exports.NODE_CLIENT_VERSION = '3.1.0-beta1';
...
+ *  Callbacks will receive an object argument with the following properties:
+ *    - type {string}
+ *    - userId {string}
+ *    - attributes {Object|undefined}
+ *    - decisionInfo {Object|undefined}
  */
 exports.NOTIFICATION_TYPES = {
   ACTIVATE: 'ACTIVATE:experiment, user_id,attributes, variation, event',
   TRACK: 'TRACK:event_key, user_id, attributes, event_tags, event',
+  ON_DECISION: 'ON_DECISION:type, userId, attributes, decisionInfo',
```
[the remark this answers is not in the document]
mfahadahmed: Yes, it has already been renamed. Please see the latest changes.

PR #245, opened 2019-03-08
feat(api): Decision Notification Listener for isFeatureEnabled and getEnabledFeatures APIs.

mfahadahmed:
## Summary
- Implemented isFeatureEnabled in activate and getEnabledFeatures APIs.

## Test plan
- Added unit tests.

PR #246, opened 2019-03-08
feat(api): Decision Notification Listener for feature variable APIs.

mfahadahmed:
## Summary
- Implemented onDecisionListener in activate and getVariation APIs.

## Test plan
- Added unit tests.


packages/optimizely-sdk/lib/core/decision_service/index.tests.js
```diff
@@ -729,22 +729,22 @@ describe('lib/core/decision_service', function() {
                     'variables': [
                       {
                         'id': '4792309476491264',
-                        'value': '10'
+                        'value': '40'
```
aliabbasrizvi: Why the change in these values?

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -2519,6 +2521,504 @@ describe('lib/optimizely', function() {
...
+                });
+              });
+            });
+
+            describe('not bucketed into an experiment or a rollout', function() {
+              beforeEach(function() {
+                sandbox.stub(optlyInstance.decisionService, 'getVariationForFeature').returns({
+                  experiment: null,
+                  variation: null,
+                  decisionSource: null,
```
aliabbasrizvi: I feel the actual `decisionService` needs to be updated to set decisionSource as rollout.

Thoughts @jordangarcia @mjc1283 ?

packages/optimizely-sdk/lib/core/decision_service/index.tests.js
mfahadahmed: I used this variation and its related variables to verify default values when feature toggle is off. But now I realized that this can easily be done without changing values in the datafile. 

Reverted decision_service and testdata files. Thanks for pointing this out.

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -563,17 +564,51 @@ Optimizely.prototype._getFeatureVariableForType = function(featureKey, variableK
...
+  var featureEnabled = false;
+  var variableValue = variable.defaultValue;
   var decision = this.decisionService.getVariationForFeature(featureFlag, userId, attributes);
-  var variableValue;
+  
   if (decision.variation !== null) {
-    variableValue = projectConfig.getVariableValueForVariation(this.configObj, variable, decision.variation, this.logger);
-    this.logger.log(LOG_LEVEL.INFO, sprintf(LOG_MESSAGES.USER_RECEIVED_VARIABLE_VALUE, MODULE_NAME, variableKey, featureFlag.key, variableValue, userId));
+    featureEnabled = decision.variation.featureEnabled;
+    if (featureEnabled === true) {
```
jordangarcia: This logic is incorrect.  

Feature variables should always be the variation variable value, even if the feature is disabled.

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -563,17 +564,51 @@ Optimizely.prototype._getFeatureVariableForType = function(featureKey, variableK
...
+      variableValue = projectConfig.getVariableValueForVariation(this.configObj, variable, decision.variation, this.logger);
+      this.logger.log(LOG_LEVEL.INFO, sprintf(LOG_MESSAGES.USER_RECEIVED_VARIABLE_VALUE, MODULE_NAME, variableKey, featureFlag.key, variableValue, userId));
+    } else {
+      this.logger.log(LOG_LEVEL.INFO, sprintf(LOG_MESSAGES.FEATURE_NOT_ENABLED_RETURN_DEFAULT_VARIABLE_VALUE, MODULE_NAME, featureFlag.key, userId, variableKey));
+      featureEnabled = false;
```
jordangarcia: There is no reason to re-card `featureEnabled` here.

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -2519,6 +2521,511 @@ describe('lib/optimizely', function() {
...
+        });
+
+        describe('feature management', function() {
+          var sandbox = sinon.sandbox.create();
+
+          beforeEach(function() {
+            optlyInstance = new Optimizely({
+              clientEngine: 'node-sdk',
+              datafile: testData.getTestProjectConfigWithFeatures(),
+              eventBuilder: eventBuilder,
```
jordangarcia: `eventBuilder` isn't a valid option to pass to `new Optimizely`

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -2519,6 +2521,511 @@ describe('lib/optimizely', function() {
...
+                      sourceVariationKey: 'variation'
+                    }
+                  });
+                });
+              });
+      
+              describe('when the variation is toggled OFF', function() {
+                beforeEach(function() {
+                  // Setting featureEnabled to false for verifying feature toggle off.
+                  optlyInstance.configObj.experimentKeyMap.testing_my_feature.variations[0].featureEnabled = false;
```
jordangarcia: We shouldn't be reaching into the `configObj` like this.

We should change the test datafile that were using to include a feature where a variation has `featureEnabled = false`

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -2519,6 +2521,511 @@ describe('lib/optimizely', function() {
...
+                  var experiment = optlyInstance.configObj.experimentKeyMap.testing_my_feature;
```
jordangarcia: use `projectConfig.getExperimentFromKey(optlyInstance.configObj, 'testing_my_feature')` instead of directly inspecting the `configObj`

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -2519,6 +2521,511 @@ describe('lib/optimizely', function() {
...
+                    }
+                  });
+                });
+              });
+            });
+
+            describe('bucketed into variation of a rollout with variable values', function() {
+              describe('when the variation is toggled ON', function() {
+                beforeEach(function() {
+                  var experiment = optlyInstance.configObj.experimentKeyMap['594031'];
```
jordangarcia: use `projectConfig` function here instead.

packages/optimizely-sdk/lib/optimizely/index.tests.js
jordangarcia: Yeah I think `decisionSource` should be rollout even if the user is not evaluated into the experiment, as it's jkind of the bottom of evaluation chain.

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -2608,6 +3117,7 @@ describe('lib/optimizely', function() {
       sandbox.stub(uuid, 'v4').returns('a68cf1ad-0393-4e18-af87-efe8f01a7c9c');
       sandbox.stub(fns, 'currentTimestamp').returns(1509489766569);
       clock = sinon.useFakeTimers(new Date().getTime());
+      onDecisionListener = sinon.spy();
```
jordangarcia: Is this still needed?

packages/optimizely-sdk/lib/optimizely/index.js
aliabbasrizvi: @jordangarcia it is actually supposed to be the default feature value

packages/optimizely-sdk/lib/optimizely/index.js
mfahadahmed: I think this value can be undefined when featureEnabled property is missing from the variation as specified in this test https://github.com/optimizely/javascript-sdk/blob/553e6ffa485a519ed4de399114768ff8bdcc6c5b/packages/optimizely-sdk/lib/optimizely/index.tests.js#L2896

packages/optimizely-sdk/lib/optimizely/index.js
jordangarcia: @aliabbasrizvi, agree that this is the default behavior.  And the datafile constraint of a feature experiment variation (where `featureEnabled == false`) always has variables that mirror the variable defaults makes it so this works today.

Given the scope of this PR (just adding notificationCenter listener), i'd prefer not to change this core logic. 

Java also mirrors the existing JS implementation https://github.com/optimizely/java-sdk/blob/master/core-api/src/main/java/com/optimizely/ab/Optimizely.java#L615-L668

mfahadahmed: Modified the logic to return variation's variable value even when the featureEnabled property is set to false.

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -614,35 +614,46 @@ Optimizely.prototype._getFeatureVariableForType = function(featureKey, variableK
...
+    featureEnabled = decision.variation.featureEnabled;
+    variableValue = projectConfig.getVariableValueForVariation(this.configObj, variable, decision.variation, this.logger);
+    this.logger.log(LOG_LEVEL.INFO, sprintf(LOG_MESSAGES.USER_RECEIVED_VARIABLE_VALUE, MODULE_NAME, variableKey, featureFlag.key, variableValue, userId));
   } else {
-    variableValue = variable.defaultValue;
-    this.logger.log(
-      LOG_LEVEL.INFO,
-      sprintf(LOG_MESSAGES.USER_RECEIVED_DEFAULT_VARIABLE_VALUE, MODULE_NAME, userId, variableKey, featureFlag.key)
-    );
+  	variableValue = variable.defaultValue;
```
jordangarcia: fix this indentation.

packages/optimizely-sdk/lib/utils/enums/index.js
```diff
@@ -199,8 +199,8 @@ exports.DECISION_INFO_TYPES = {
  * Optimizely.
  */
 exports.DECISION_SOURCES = {
-  EXPERIMENT: 'experiment',
-  ROLLOUT: 'rollout',
+  EXPERIMENT: 'EXPERIMENT',
```
jordangarcia: This should be lowercase, right?

@aliabbasrizvi

mfahadahmed: I think it will be upper case as @aliabbasrizvi suggested here
https://github.com/optimizely/javascript-sdk/pull/245#discussion_r268360395

aliabbasrizvi:
@jordangarcia can you take a look again and approve/decline?

msohailhussain:
@jordangarcia can you please rereview and let me know if any change is needed?

jordangarcia:
LGTM

PR #250, opened 2019-03-26
Update existing eventDispatchers to properly respect the callback suc…

jordangarcia:
## Summary
- Change existing eventDispatcher implementations to use `callback(success)` instead of (inconsistently) passing back the response object. 

This is needed to be able to properly track if a request has completed or errored for the retry mechanism in an upcoming PR

## Test plan
- Unit tests


jordangarcia:
Hold off on review,  after speaking with Logan i'm going to change this interface a bit.

PR #251, opened 2019-03-27
Add PendingEventsDispatcher in event-processor package

jordangarcia:
## Summary
- Add `PendingEventsDispatcher` to wrap another EventDispatcher with retry support for
events that did not send successfully due to page navigation

```
var wrappedEventDispatcher = new eventProcessor.LocalStoragePendingEventsDispatcher({
  eventDispatcher: config.eventDispatcher,
});
wrappedEventDispatcher.sendPendingEvents()
```

## Test plan
- Unit tests

## Issues
- OASIS-4409


packages/event-processor/__tests__/pendingEventsDispatcher.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+      event: ({ id: 'event' } as unknown) as EventV1,
+    }
+
+    pendingEventsDispatcher.dispatch(eventV1Request, callback)
+
+    expect(callback).not.toHaveBeenCalled()
+    // manually invoke original eventDispatcher callback
+    const internalDispatchCall = ((originalEventDispatcher.dispatch as unknown) as jest.Mock)
+      .mock.calls[0]
+    internalDispatchCall[1](true)
```
mjc1283: Hmm, so the event dispatcher calls callbacks with booleans? I'm confused. Here it looks like it calls the callback with the response:

https://github.com/optimizely/javascript-sdk/blob/793b57866b81eebaf3c934d82904aa4aa7d9acea/packages/optimizely-sdk/lib/plugins/event_dispatcher/index.node.js#L56

And, it doesn't even call it at all when the response has an error status.

packages/event-processor/jest.config.js
```diff
@@ -11,4 +11,5 @@ module.exports = {
     "json",
     "node"
   ],
+  "setupFiles": ["jest-localstorage-mock"]
```
mjc1283: Cool 👍

packages/event-processor/src/pendingEventsStore.ts
```diff
@@ -0,0 +1,115 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { objectValues } from '@optimizely/js-sdk-utils'
+import { getLogger } from '@optimizely/js-sdk-logging';
+
+const logger = getLogger('EventProcessor')
+
+export interface PendingEventsStore<K> {
```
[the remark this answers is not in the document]
jordangarcia: Its only used for this purpose, but I was thinking this is akin to `Map` which is usually typed as a generic.

Making this generic made it a lot easier to test.

packages/event-processor/src/eventDispatcher.ts
```diff
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 import { EventV1 } from "./v1/buildEventV1";
+import { Managed } from "./managed";
 
-export interface EventDispatcher {
-  dispatch(event: object, callback: (success: boolean) => void): void
+export interface EventDispatcher extends Managed {
+  dispatch(event: EventV1Request, callback: (success: boolean) => void): void
```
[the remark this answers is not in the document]
jordangarcia: In practice they were all using EventV1Request, i'd rather be explicit and than if we need to have some sort of fork then making the type more generic at that point.

packages/event-processor/__tests__/pendingEventsDispatcher.spec.ts
jordangarcia: The https://github.com/optimizely/javascript-sdk/blob/master/packages/optimizely-sdk/lib/optimizely/event_dispatcher_bridge.js makes it such that its always a boolean.

But we need to fix the callbacks in the existing eventDispatchers to consistent give the response.

PR #256, opened 2019-04-08
fix (unit tests): Fix event dispatcher missing in unit test

mjc1283:
## Summary

`Optimizely` was constructed without an event dispatcher, causing an error to be thrown asynchronously. This fixes the test, and updates the `npm test` command to make it easier to catch such mistakes - adding a listener for `'unhandledRejection'` on `process` that calls `process.exit`.

## Test plan

Run unit tests, verify log output.


packages/optimizely-sdk/exit_on_unhandled_rejection.js
```diff
@@ -0,0 +1,7 @@
+process.on('unhandledRejection', function(err) {
```
jordangarcia: Mind adding a comment in this explaining why this is here.

packages/optimizely-sdk/exit_on_unhandled_rejection.js
```diff
@@ -0,0 +1,11 @@
+/*
```
mikeng13: Maybe we should move this into the `lib` dir and further, into the `tests` dir since this is only used for tests.

Also, please include the license header for this code file :)

PR #258, opened 2019-04-16
feat(decision-listener): Incorporated new decision notification listener changes.

mfahadahmed:
## Summary
- Incorporated new decision notification listener changes.

## Test plan
- Added unit tests.

### DIFF
--- a/packages/optimizely-sdk/lib/core/project_config/index.js
+++ b/packages/optimizely-sdk/lib/core/project_config/index.js
@@ -89,10 +89,16 @@ module.exports = {
 
     projectConfig.forcedVariationMap = {};
 
+    // Object containing experiment Ids that exist in any feature
+    // for checking that experiment is a feature experiment or not.
+    projectConfig.featureExperimentsMap = {};
+
     projectConfig.featureKeyMap = fns.keyBy(projectConfig.featureFlags || [], 'key');
     fns.forOwn(projectConfig.featureKeyMap, function(feature) {
       feature.variableKeyMap = fns.keyBy(feature.variables, 'key');
       fns.forEach(feature.experimentIds || [], function(experimentId) {
+        // Add this experiment in feature experiment map.
+        projectConfig.featureExperimentsMap[experimentId] = true;
         var experimentInFeature = projectConfig.experimentIdMap[experimentId];
         if (experimentInFeature.groupId) {
           feature.groupId = experimentInFeature.groupId;
@@ -598,17 +604,11 @@ module.exports = {
   /**
    * 
    * @param {Object} projectConfig
-   * @param {string} experimentKey
+   * @param {string} experimentId
    * @returns {boolean} Returns true if experiment belongs to
    * any feature, false otherwise.
    */
-  isFeatureExperiment: function(projectConfig, experimentKey) {
-    fns.forOwn(projectConfig.featureKeyMap, function(feature) {
-      if (feature.experimentIds && feature.experimentIds.indexOf(experimentKey)) {
-        return true;
-      }
-    });
-
-    return false;
+  isFeatureExperiment: function(projectConfig, experimentId) {
+    return projectConfig.featureExperimentsMap[experimentId];
   }
 };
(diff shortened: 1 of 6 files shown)


packages/optimizely-sdk/lib/utils/enums/index.js
```diff
@@ -190,8 +190,9 @@ exports.NOTIFICATION_TYPES = {
 };
 
 exports.DECISION_INFO_TYPES = {
-  EXPERIMENT: 'experiment',
+  AB_TEST: 'ab-test',
   FEATURE: 'feature',
+  FEATURE_TEST: 'feature-test',
   FEATURE_VARIABLE: 'feature_variable',
```
aliabbasrizvi: Missed it earlier. This needs to be `feature-variable`

packages/optimizely-sdk/lib/core/project_config/index.js
```diff
@@ -89,10 +89,16 @@ module.exports = {
 
     projectConfig.forcedVariationMap = {};
 
+    // Object containing experiment Ids that exist in any feature
+    // for checking that experiment is a feature experiment or not.
+    projectConfig.featureExperimentsMap = {};
```
msohailhussain: More meaningful is experimentFeatureMap.

packages/optimizely-sdk/lib/core/project_config/index.js
```diff
@@ -89,10 +89,16 @@ module.exports = {
...
+
     projectConfig.featureKeyMap = fns.keyBy(projectConfig.featureFlags || [], 'key');
     fns.forOwn(projectConfig.featureKeyMap, function(feature) {
       feature.variableKeyMap = fns.keyBy(feature.variables, 'key');
       fns.forEach(feature.experimentIds || [], function(experimentId) {
+        // Add this experiment in feature experiment map.
+        projectConfig.featureExperimentsMap[experimentId] = true;
```
msohailhussain: Instead of `true`, if we can assign featureId would be more meaning. @aliabbasrizvi What do you suggest?

packages/optimizely-sdk/lib/core/project_config/index.js
```diff
@@ -594,4 +600,15 @@ module.exports = {
...
+
+  /**
+   * 
+   * @param {Object} projectConfig
+   * @param {string} experimentId
+   * @returns {boolean} Returns true if experiment belongs to
+   * any feature, false otherwise.
+   */
+  isFeatureExperiment: function(projectConfig, experimentId) {
+    return projectConfig.featureExperimentsMap[experimentId];
```
msohailhussain: This condition after making suggested changes, should be like 
`projectConfig.featureExperimentsMap[experimentId] != undefined`

packages/optimizely-sdk/lib/core/project_config/index.js
aliabbasrizvi: Agreed. It would be much better if we add `featureId` on `experiment` and use that.

packages/optimizely-sdk/lib/utils/enums/index.js
```diff
@@ -190,9 +190,10 @@ exports.NOTIFICATION_TYPES = {
 };
 
 exports.DECISION_INFO_TYPES = {
```
aliabbasrizvi: nit. We should call this `DECISION_NOTIFICATION_TYPES`

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -343,10 +343,13 @@ Optimizely.prototype.getVariation = function(experimentKey, userId, attributes)
       }
 
       var variationKey = this.decisionService.getVariation(this.configObj, experimentKey, userId, attributes);
+      var decisionInfoType = projectConfig.isFeatureExperiment(this.configObj, experiment.id) ? DECISION_INFO_TYPES.FEATURE_TEST :
```
aliabbasrizvi: I'd like to call this `decisionNotificationType`.

PR #261, opened 2019-04-24
feat: Datafile management

mjc1283:
## Summary
Add support for datafile management.

## Test plan
Unit tests, compatibility suite, manual testing

## Issues
https://optimizely.atlassian.net/browse/OASIS-4122


packages/datafile-manager/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@optimizely/js-sdk-datafile-manager",
+  "version": "0.2.0",
+  "description": "Optimizely Full Stack Datafile Manager",
+  "license": "Apache-2.0",
+  "engines": {
+    "node": ">=4.0.0"
```
mikeng13: nit: We technically only support Node 6+ so I don't wanna give the false impression that we officially support 4 (even though it would work)

packages/optimizely-sdk/CHANGELOG.MD
```diff
@@ -7,6 +7,57 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+    - Enable automatic updates by passing `autoUpdate: true`. Periodically (on the provided update interval), the SDK instance will download the datafile and update itself. Use this to ensure that the SDK instance is using a fresh datafile reflecting changes recently made to your experiment or feature configuration.
+  - Add a notification listener for the `OPTIMIZELY_CONFIG_UPDATE` notification type to be notified when an instance updates its Optimizely config after obtaining a new datafile.
+  - Stop active downloads and cancel pending downloads by calling the `close` method
+
+
+  #### Create an instance with datafile management enabled
+  ```js
+  const optimizely = require('@optimizely/optimizely-sdk');
+  const optimizelyClientInstance = optimizely.createInstance({
+    sdkKey: ‘12345’, // Provide the sdkKey of your desired environment here
```
mikeng13: nit: these ticks/apostrophes seem off, is that intentional?

packages/optimizely-sdk/lib/utils/json_schema_validator/index.tests.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016-2017, Optimizely
+ * Copyright 2016-2019, Optimizely
```
mikeng13: If this file wasn't touched in 2018 it should be `2016-2017, 2019`

packages/optimizely-sdk/lib/core/project_config/project_config_schema.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016-2017, Optimizely
+ * Copyright 2016-2019, Optimizely
```
mikeng13: If this file wasn't touched in 2018 it should be `2016-2017, 2019`

packages/optimizely-sdk/CHANGELOG.MD
mjc1283: Whoa, good eye. I think it's because I copied from a google doc.

packages/optimizely-sdk/lib/core/project_config/project_config_schema.js
mjc1283: This was only moved, but not changed. Should it be left as 2016-2017?

packages/optimizely-sdk/CHANGELOG.MD
```diff
@@ -7,6 +7,57 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 Changes that have landed but are not yet released.
 
+### New Features
+
+- Added support for automatic datafile management
```
mikeng13: I think we should also mention in the unreleased notes that the forced variation logic has been moved to the decision service module because that affects where those messages are logged via the `MODULE_NAME`

packages/optimizely-sdk/lib/core/project_config/index.js
```diff
@@ -607,13 +464,35 @@ module.exports = {
...
+   * configuration properties.
+   * If successful, return the project config object, otherwise throws an error
+   * @param  {Object} config
+   * @param  {Object} config.datafile
+   * @param  {Object} config.jsonSchemaValidator
+   * @param  {Object} config.logger
+   * @param  {Object} config.skipJSONValidation
+   * @return {Object} Project config object
+   */
+  tryCreatingProjectConfig: function(config) {
```
mikeng13: We can punt this to a later PR, but I think this factory method should live in a factory class or something instead of being in the project config itself

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -67,29 +67,23 @@ function Optimizely(config) {
   this.clientVersion = config.clientVersion || enums.NODE_CLIENT_VERSION;
   this.errorHandler = config.errorHandler;
   this.eventDispatcher = config.eventDispatcher;
-  this.isValidInstance = config.isValidInstance;
+  this.__isOptimizelyConfigValid = config.isValidInstance;
```
mikeng13: nit: I think this property might be useful to developers. Did we consider exposing `__isValidInstance`? I know that both mobile SDKs make this public because it is used in one of the initialization scenarios

mjc1283: I hadn't considered this. Sounds good to me if there's a use case for it.

PR #264, opened 2019-05-02
feat(utils): Add EventEmitter to be shared between datafile-manager and optimizely-sdk

mjc1283:
## Summary
Currently, ProjectConfigManager and DatafileManager are both implementing their own simple event emitters. This adds an EventEmitter to the utils package that we can share.

I'll follow-up by replacing the versions in ProjectConfigManager and DatafileManager with this version

## Test plan

Unit tests


packages/utils/src/eventEmitter.ts
```diff
@@ -0,0 +1,59 @@
...
+      if (listenersForEvent) {
+        const index = listenersForEvent.indexOf(listener);
+        if (index > -1) {
+          listenersForEvent.splice(index, 1);
+        }
+      }
+    }
+  }
+
+  emit(eventName: string, arg?: any) {
```
mikeng13: Should this be `...args` so we can pass any number of arbitrary args to the listener? I think it makes it a bit more flexible

mjc1283: My thought is, in the current version, you could pass an object or array if you wanted to emit multiple values. I don't have a strong preference one way or the other, but at least for the current consumers of this, they don't need to use multiple argument event listeners.

mjc1283:
@mikeng13 The typedef will get updated automatically because we have `"declaration": true` in our tsconfig.json.

packages/utils/src/eventEmitter.ts
mikeng13: Ok, fine by me

PR #268, opened 2019-05-10
feat(datafile manager): Add request timeouts

mjc1283:
## Summary
This adds a 60 second timeout to all requests made by the datafile manager. In the browser, we assign to the `timeout` property of the XHR. In node, we set our own timeout, and abort the request when the timeout fires.

The response promise is rejected when the request times out.

## Test plan

Manual & unit tests

## Issues
https://optimizely.atlassian.net/browse/OASIS-4685


packages/datafile-manager/package.json
```diff
@@ -25,12 +25,12 @@
   },
   "devDependencies": {
     "@types/jest": "^24.0.9",
+    "@types/nise": "^1.4.0",
```
mikeng13: Nise!

packages/datafile-manager/__test__/nodeRequest.spec.ts
```diff
@@ -156,5 +157,41 @@ describe('nodeEnvironment', () => {
...
+
+      afterEach(() => {
+        jest.clearAllTimers()
+      })
+
+      it('rejects the response promise and aborts the request when the response is not received before the timeout', async () => {
+        const scope = nock(host)
+          .get(path)
+          .delay(61000)
+          .reply(200, '{"foo":"bar"}')
```
mikeng13: nit: do you actually need to reply if it's gunna just time out

packages/datafile-manager/src/config.ts
```diff
@@ -21,3 +21,5 @@ export const MIN_UPDATE_INTERVAL = 1000
 export const DEFAULT_URL_TEMPLATE = `https://cdn.optimizely.com/datafiles/%s.json`
 
 export const BACKOFF_BASE_WAIT_SECONDS_BY_ERROR_COUNT = [0, 8, 16, 32, 64, 128, 256, 512]
+
+export const REQUEST_TIMEOUT_MS = 60 * 1000 // 1 minute
```
mikeng13: nit: DEFAULT_REQUEST_TIME_MS?

packages/datafile-manager/src/nodeRequest.ts
```diff
@@ -64,6 +65,11 @@ function getResponseFromRequest(request: http.ClientRequest): Promise<Response>
   // TODO: When we drop support for Node 6, consider using util.promisify instead of
   // constructing own Promise
   return new Promise((resolve, reject) => {
+    const timeout = setTimeout(() => {
+      request.abort()
+      reject(new Error('Request timed out'))
+    }, REQUEST_TIMEOUT_MS)
```
mikeng13: I think we should make this configurable and have it injected in by the datafile manager. That we this doesn't have to rely directly on the config file

mikeng13: Chatted offline about doing this as a later refactor to keep the scope of this small.

PR #278, opened 2019-05-23
Update `index.d.ts` file

jordangarcia:
## Summary
- Remove declaring multiple files, this appeared to not be working in the
first place.
- Add logging / errorHandler / eventDispatcher and other top level types
- Fix the `onReady()` signature to include the proper promise resolved
value
 
## Test plan
- Manually tested with a typescript project importing these types.

## Issues
- OASIS-4742


mjc1283:
Still LGTM too

PR #288, opened 2019-06-07
feat(AudienceEvaluator): Add the ability to provide custom condition evaluators

jamesopti:
## Summary
This PR (replacing https://github.com/optimizely/javascript-sdk/pull/216) adds the ability for an SDK consumer to pass their own set of condition evaluators when creating an optimizely instance.

The core motivator for this is to be able to evaluate WEB audience conditions using the Javascript SDK as part of the EDGE project (go/edge).

Future fullstack use cases for this functionality might include audience condition plugins, where the customer would define their own match types and provide their own matchers.

```javascript
/**
 * Condition evaluators for conditions that can be evaluated in a Cloudflare worker using
 * information like cookies, query parameters, userAgent, etc...
 */
function getConditionEvaluators(inputs) {
  return {
    cookies: {
      evaluate: matchCookies.match.bind(this, { cookies: inputs.cookies }),
    },
    device: {
      evaluate: matchDevice.match.bind(this, { device: Detect.parseUA(inputs.ua).device }),
    },
    ...
    /* Other custom condition evaluators */
  }
}

var optlyInstance = optimizely.createInstance({
  datafile: datafile,
  conditionEvaluators: getConditionEvaluators(inputs),
  skipJSONValidation: true
});
```

## Test plan

Fixed existing tests. May need to add more unit tests. 

## Issues
https://optimizely.atlassian.net/browse/CJS-3034

### DIFF
--- a/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
+++ b/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
@@ -93,7 +93,7 @@ AudienceEvaluator.prototype.evaluateConditionWithUserAttributes = function(userA
     return null;
   }
   try {
-    return evaluator.evaluate(condition, userAttributes, logger);
+    return evaluator.evaluate(condition, userAttributes);
   } catch (err) {
     logger.log(LOG_LEVEL.ERROR, sprintf(ERROR_MESSAGES.CONDITION_EVALUATOR_ERROR, MODULE_NAME, condition.type, err.message));
   }
--- a/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
+++ b/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
@@ -28,14 +28,14 @@ var MODULE_NAME = 'AUDIENCE_EVALUATOR';
 /**
  * Construct an instance of AudienceEvaluator with a given logger and options
  * @param {Logger}  logger                           The Logger instance
- * @param {Object=} __exploratoryConditionEvaluators A map of condition evaluators provided by the consumer. This enables matching
+ * @param {Object=} UNSTABLE_conditionEvaluators A map of condition evaluators provided by the consumer. This enables matching
  *                                                   condition types which are not supported natively by the SDK. Note that built in
  *                                                   Optimizely evaluators cannot be overridden.
  * @constructor
  */
-function AudienceEvaluator(logger, __exploratoryConditionEvaluators) {
+function AudienceEvaluator(logger, UNSTABLE_conditionEvaluators) {
   this.logger = logger;
-  this.typeToEvaluatorMap = fns.assignIn({}, __exploratoryConditionEvaluators, {
+  this.typeToEvaluatorMap = fns.assignIn({}, UNSTABLE_conditionEvaluators, {
     'custom_attribute': customAttributeConditionEvaluator
   });
 }
--- a/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
+++ b/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
@@ -18,6 +18,8 @@ var customAttributeConditionEvaluator = require('../custom_attribute_condition_e
 var enums = require('../../utils/enums');
 var fns = require('../../utils/fns');
 var sprintf = require('@optimizely/js-sdk-utils').sprintf;
+var logging = require('@optimizely/js-sdk-logging');
+var logger = logging.getLogger();
 
 var ERROR_MESSAGES = enums.ERROR_MESSAGES;
 var LOG_LEVEL = enums.LOG_LEVEL;
@@ -26,15 +28,13 @@ var MODULE_NAME = 'AUDIENCE_EVALUATOR';
 
 
 /**
- * Construct an instance of AudienceEvaluator with a given logger and options
- * @param {Logger}  logger                           The Logger instance
+ * Construct an instance of AudienceEvaluator with given options
  * @param {Object=} UNSTABLE_conditionEvaluators A map of condition evaluators provided by the consumer. This enables matching
  *                                                   condition types which are not supported natively by the SDK. Note that built in
  *                                                   Optimizely evaluators cannot be overridden.
  * @constructor
  */
-function AudienceEvaluator(logger, UNSTABLE_conditionEvaluators) {
-  this.logger = logger;
+function AudienceEvaluator(UNSTABLE_conditionEvaluators) {
   this.typeToEvaluatorMap = fns.assignIn({}, UNSTABLE_conditionEvaluators, {
     'custom_attribute': customAttributeConditionEvaluator
   });
@@ -66,10 +66,10 @@ AudienceEvaluator.prototype.evaluate = function(audienceConditions, audiencesByI
   var evaluateAudience = function(audienceId) {
     var audience = audiencesById[audienceId];
     if (audience) {
-      this.logger.log(LOG_LEVEL.DEBUG, sprintf(LOG_MESSAGES.EVALUATING_AUDIENCE, MODULE_NAME, audienceId, JSON.stringify(audience.conditions)));
+      logger.log(LOG_LEVEL.DEBUG, sprintf(LOG_MESSAGES.EVALUATING_AUDIENCE, MODULE_NAME, audienceId, JSON.stringify(audience.conditions)));
       var result = conditionTreeEvaluator.evaluate(audience.conditions, this.evaluateConditionWithUserAttributes.bind(this, userAttributes));
       var resultText = result === null ? 'UNKNOWN' : result.toString().toUpperCase();
-      this.logger.log(LOG_LEVEL.INFO, sprintf(LOG_MESSAGES.AUDIENCE_EVALUATION_RESULT, MODULE_NAME, audienceId, resultText));
+      logger.log(LOG_LEVEL.INFO, sprintf(LOG_MESSAGES.AUDIENCE_EVALUATION_RESULT, MODULE_NAME, audienceId, resultText));
       return result;
     }
 
@@ -89,13 +89,13 @@ AudienceEvaluator.prototype.evaluate = function(audienceConditions, audiencesByI
 AudienceEvaluator.prototype.evaluateConditionWithUserAttributes = function(userAttributes, condition) {
   var evaluator = this.typeToEvaluatorMap[condition.type];
   if (!evaluator) {
-    this.logger.log(LOG_LEVEL.WARNING, sprintf(LOG_MESSAGES.UNKNOWN_CONDITION_TYPE, MODULE_NAME, JSON.stringify(condition)));
+    logger.log(LOG_LEVEL.WARNING, sprintf(LOG_MESSAGES.UNKNOWN_CONDITION_TYPE, MODULE_NAME, JSON.stringify(condition)));
     return null;
   }
   try {
-    return evaluator.evaluate(condition, userAttributes, this.logger);
+    return evaluator.evaluate(condition, userAttributes, logger);
   } catch (err) {
-    this.logger.log(LOG_LEVEL.ERROR, sprintf(ERROR_MESSAGES.CONDITION_EVALUATOR_ERROR, MODULE_NAME, condition.type, err.message));
+    logger.log(LOG_LEVEL.ERROR, sprintf(ERROR_MESSAGES.CONDITION_EVALUATOR_ERROR, MODULE_NAME, condition.type, err.message));
   }
   return null;
 };
--- a/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
+++ b/packages/optimizely-sdk/lib/core/audience_evaluator/index.js
@@ -16,54 +16,88 @@
 var conditionTreeEvaluator = require('../condition_tree_evaluator');
 var customAttributeConditionEvaluator = require('../custom_attribute_condition_evaluator');
 var enums = require('../../utils/enums');
+var fns = require('../../utils/fns');
 var sprintf = require('@optimizely/js-sdk-utils').sprintf;
 
+var ERROR_MESSAGES = enums.ERROR_MESSAGES;
 var LOG_LEVEL = enums.LOG_LEVEL;
 var LOG_MESSAGES = enums.LOG_MESSAGES;
 var MODULE_NAME = 'AUDIENCE_EVALUATOR';
 
-module.exports = {
-  /**
-   * Determine if the given user attributes satisfy the given audience conditions
-   * @param  {Array|String|null|undefined}  audienceConditions    Audience conditions to match the user attributes against - can be an array
-   *                                                              of audience IDs, a nested array of conditions, or a single leaf condition.
-   *                                                              Examples: ["5", "6"], ["and", ["or", "1", "2"], "3"], "1"
-   * @param  {Object}                       audiencesById         Object providing access to full audience objects for audience IDs
-   *                                                              contained in audienceConditions. Keys should be audience IDs, values
-   *                                                              should be full audience objects with conditions properties
-   * @param  {Object}                       [userAttributes]      User attributes which will be used in determining if audience conditions
-   *                                                              are met. If not provided, defaults to an empty object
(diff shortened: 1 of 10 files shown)


packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.js
```diff
@@ -56,7 +56,6 @@ EVALUATORS_BY_MATCH_TYPE[SUBSTRING_MATCH_TYPE] = substringEvaluator;
  */
 function evaluate(condition, userAttributes, logger) {
   if (condition.type !== CUSTOM_ATTRIBUTE_CONDITION_TYPE) {
-    logger.log(LOG_LEVEL.WARNING, sprintf(LOG_MESSAGES.UNKNOWN_CONDITION_TYPE, MODULE_NAME, JSON.stringify(condition)));
     return null;
   }
```
lpappone: delete? because we're handling this in typeToEvaluate - we only call this if there's a type.

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -16,54 +16,88 @@
...
-      return true;
-    }
 
-    if (!userAttributes) {
-      userAttributes = {};
-    }
+/**
+ * Construct an instance of AudienceEvaluator with a given logger and options
+ * @param {Logger}  logger                           The Logger instance
+ * @param {Object=} UNSTABLE_conditionEvaluators A map of condition evaluators provided by the consumer. This enables matching
```
mikeng13: Do we really need to name this `UNSTABLE?` We can make this `customConditionEvaluators` instead

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -16,54 +16,88 @@
...
+ *                                                   condition types which are not supported natively by the SDK. Note that built in
+ *                                                   Optimizely evaluators cannot be overridden.
+ * @constructor
+ */
+function AudienceEvaluator(logger, UNSTABLE_conditionEvaluators) {
+  this.logger = logger;
```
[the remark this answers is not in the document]
mikeng13: +1

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -16,54 +16,88 @@
...
 
-    return conditionTreeEvaluator.evaluate(audienceConditions, evaluateAudience) || false;
-  },
+    return null;
+  }.bind(this);
+
+  return conditionTreeEvaluator.evaluate(audienceConditions, evaluateAudience) || false;
+};
+
+/**
```
mikeng13: Thanks so much for adding docs =)

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -16,54 +16,88 @@
...
+ * @param {Logger}  logger                           The Logger instance
+ * @param {Object=} UNSTABLE_conditionEvaluators A map of condition evaluators provided by the consumer. This enables matching
+ *                                                   condition types which are not supported natively by the SDK. Note that built in
+ *                                                   Optimizely evaluators cannot be overridden.
+ * @constructor
+ */
+function AudienceEvaluator(logger, UNSTABLE_conditionEvaluators) {
+  this.logger = logger;
+  this.typeToEvaluatorMap = fns.assignIn({}, UNSTABLE_conditionEvaluators, {
+    'custom_attribute': customAttributeConditionEvaluator
```
mikeng13: This means that we can only have custom evaluators for non `custom_attribute` condition types correct? Would there be a situation where you'd want the custom evaluators to also work on `custom_attribute`?

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
lpappone: James named it this way (he had `exploratory`, Jordan suggested `UNSTABLE`) because, in his words from the original PR: 
"The idea here is to indicate to any prying eyes to "use at your own risk" before this is made publicly available/official.

Until we officially support a way to use different audience condition types in FS audiences, there would be no other reason for a customer to actually use this."

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
lpappone: No, I think we just need them for specific audiences that we want to support for Edge (that the web client already supports).

nchilada:
> ```js
> /**
>  * Condition evaluators for conditions that can be evaluated in a Cloudflare worker using
>  * information like cookies, query parameters, userAgent, etc...
>  */
> function getConditionEvaluators(inputs) {
>   return {
>     cookies: {
>       evaluate: matchCookies.match.bind(this, { cookies: inputs.cookies }),
>     },
>     device: {
>       evaluate: matchDevice.match.bind(this, { device: Detect.parseUA(inputs.ua).device }),
>     },
>     ...
>     /* Other custom condition evaluators */
>   }
> }
> 
> var optlyInstance = optimizely.createInstance({
>   datafile: datafile,
>   conditionEvaluators: getConditionEvaluators(inputs),
>   skipJSONValidation: true
> });
> ```

Thanks for the code sample in the PR description! Very useful. But I do think it's a little confusing for it to capture `inputs` at the moment the SDK is initialized. Instead, can we declare a module-level variable named `userState` or something? Might be worthwhile since the PR description / commit message is de-facto documentation for the new feature.

packages/optimizely-sdk/lib/utils/enums/index.js
```diff
@@ -26,6 +26,7 @@ exports.LOG_LEVEL = {
 };
 
 exports.ERROR_MESSAGES = {
+  CONDITION_EVALUATOR_ERROR: '%s: Error evaluating condition type %s: %s',
```
nchilada:
```suggestion
  CONDITION_EVALUATOR_ERROR: '%s: Error evaluating audience condition of type %s: %s',
```

packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.tests.js
```diff
@@ -85,42 +85,6 @@ describe('lib/core/custom_attribute_condition_evaluator', function() {
...
-      { weird_condition: 'bye' },
-      mockLogger
-    );
-    assert.isNull(result);
-    sinon.assert.calledOnce(mockLogger.log);
-    sinon.assert.calledWithExactly(mockLogger.log, LOG_LEVEL.WARNING,
-      'CUSTOM_ATTRIBUTE_CONDITION_EVALUATOR: Audience condition {"match":"exact","name":"weird_condition","value":"hi"} has an unknown condition type. You may need to upgrade to a newer release of the Optimizely SDK.');
-  });
-
-  it('should log and return null when condition has an invalid match property', function() {
```
nchilada: Why are we removing this test case? (I agree with relocating the prior two test cases 👍)

packages/optimizely-sdk/lib/core/decision_service/index.js
```diff
@@ -45,13 +45,13 @@ var DECISION_SOURCES = enums.DECISION_SOURCES;
  * @constructor
  * @param   {Object} options
  * @param   {Object} options.userProfileService An instance of the user profile service for sticky bucketing.
- * @param   {Object} options.logger             An instance of a logger to log messages with.
```
nchilada: As long as the decision service still uses `options.logger`, rather than importing `@optimizely/js-sdk-logging`, we should probably continue to document `options.logger`

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -16,54 +16,88 @@
 var conditionTreeEvaluator = require('../condition_tree_evaluator');
 var customAttributeConditionEvaluator = require('../custom_attribute_condition_evaluator');
 var enums = require('../../utils/enums');
+var fns = require('../../utils/fns');
 var sprintf = require('@optimizely/js-sdk-utils').sprintf;
+var logging = require('@optimizely/js-sdk-logging');
+var logger = logging.getLogger();
```
nchilada: Is this a new pattern for accessing the configured logger?

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -167,6 +167,7 @@ describe('lib/optimizely', function() {
           sinon.assert.calledWith(decisionService.createDecisionService, {
             userProfileService: userProfileServiceInstance,
             logger: createdLogger,
+            UNSTABLE_conditionEvaluators: undefined
```
nchilada: Nit: trailing comma. Same thing below

lpappone:
@mikeng13 @mjc1283 Could you guys weigh in on whether this is something you plan to consider exposing to customers? A lot of Nikhil's concerns seem to center on that, and it's not something that I feel I have the context to decide.

In the meantime to try to address a couple questions:
> Why require customers to pass an object with an evaluate function rather than passing a naked function? Just because that's the shape of the custom_attribute_condition_evaluator module?

Yes, I think that was the case. Maybe @jamesopti has more to say.

> I wonder if we'll eventually need to document what to do with the logger, and I think we ought to pass the user ID in addition to the user attributes.

I don't think we're passing a logger anymore - can you comment in the code if you see it? What do you think the user ID is needed for?

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
mikeng13: We envision customers being able to overwrite the evaluator for built-in types as well as well as their own condition types. However, the current architecture of the SDK does not lend itself well to cleanly expose this so I think this internal usage should suffice for now (as UNSTABLE).

We do plan on a more involved overhaul that would allow us to introduce more extensibility points throughout the SDK, possible in the next major version release.

nchilada: Thanks @mikeng13. If we do envision customers being able to overwrite builtin evaluators, I might prefer to see us switch the order from the outset:
```javascript
  this.typeToEvaluatorMap = fns.assignIn({}, {
    'custom_attribute': customAttributeConditionEvaluator,
  }, UNSTABLE_conditionEvaluators);
```
but I agree it's not a huge deal.

nchilada:
> > > Why require customers to pass an object with an evaluate function rather than passing a naked function? Just because that's the shape of the custom_attribute_condition_evaluator module?
> > 
> > 
> > Yes, I think that was the case. Maybe @jamesopti has more to say.
> 
> 🙏

@lpappone ohh, my bad, I didn't realize you were authoring this! I can try to close out some of my comment threads if you're ready to move forward

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
lpappone: I don't know if it's new; this is what @mjc1283 said was the way.

packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.tests.js
lpappone: Idk, this was probably a mistake. I put it back.

packages/optimizely-sdk/lib/core/decision_service/index.js
lpappone: Yep, put back in.

lpappone:
> @lpappone ohh, my bad, I didn't realize you were authoring this! I can try to close out some of my comment threads if you're ready to move forward

I resolved things that seemed resolvable.

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
nchilada: After offline discussion with @lpappone @mikeng13 @jamesopti: `"custom attribute"` conditions aren't supported in Edge experiments right now, so it's okay that the SDK's builtin evaluator would execute and return `null` for any such condition

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
```diff
@@ -16,54 +16,88 @@
...
+ * @return {Boolean|null}              true if the condition is satisfied, null if a matcher is not found.
+ */
+AudienceEvaluator.prototype.evaluateConditionWithUserAttributes = function(userAttributes, condition) {
+  var evaluator = this.typeToEvaluatorMap[condition.type];
+  if (!evaluator) {
+    logger.log(LOG_LEVEL.WARNING, sprintf(LOG_MESSAGES.UNKNOWN_CONDITION_TYPE, MODULE_NAME, JSON.stringify(condition)));
+    return null;
+  }
+  try {
+    return evaluator.evaluate(condition, userAttributes);
```
nchilada: I still think we need to think about the arguments we're passing here.

* Primarily we need to pass a logger or update the `custom_attribute_condition_evaluator` to import the logger directly
* Can we pass an object with named properties?

jamesopti: This is a good point.

We **_DO_** need to pass a logger - I mistakenly omitted this. We can't have `custom_attribute_condition_evaluator` import it as its stateful to this instance of `AudienceEvaluator`

> Can we pass an object with named properties?
WDYM exactly? The `custom_attribute_condition_evaluator` needs the `userAttributes` map to do it's matching, but other matchers won't necessarily need them.

lpappone:
Ok, added the logger. @nchilada I am also not sure what you intended with "Can we pass an object with named properties?"

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
nchilada: Oh. I'm proposing named parameters like this: 
```suggestion
    return evaluator.evaluate({
      condition: condition,
      userAttributes: userAttributes,
    });
```
This is how we invoke customer's notification listener callbacks ([example](https://docs.developers.optimizely.com/full-stack/docs/register-notification-listeners#section-track-notification-listener)) and, arguably, is what we ought to support in all of our public API methods.

nchilada: Oh. I'm proposing named parameters like this: 
```suggestion
    return evaluator.evaluate({
      condition: condition,
      userAttributes: userAttributes,
    });
```
This is how we invoke customer's notification listener callbacks ([example](https://docs.developers.optimizely.com/full-stack/docs/register-notification-listeners#section-track-notification-listener)) and, arguably, is how we ought to accept inputs to all of our public API methods.

lpappone: Ok, doing that here would involve changing a lot of established code (like [this](https://github.com/optimizely/javascript-sdk/blob/master/packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.js)). I don't think it makes a ton of sense to make that change in this PR.

lpappone:
@nchilada your "request changes" is blocking this from being merged, despite @mikeng13 approving it. Can you re-review please?

packages/optimizely-sdk/lib/core/audience_evaluator/index.js
nchilada: Hmm, the changes ought to be limited to [`custom_attribute_condition_evaluator.evaluate`](https://github.com/optimizely/javascript-sdk/blob/bd49de091af215c867bb1017ec3b77337cc33b33/packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.js#L48-L77) and its unit tests. I think this is the best time to take this on - we're starting to move towards a public API, and we're prepare to introduce many more evaluators than `custom_attribute_condition_evaluator`

lpappone:
@mikeng13 @mjc1283 @jamesopti 
Can you guys (at least a couple of you) weigh in on @nchilada's suggestion to pass an object with named properties to the `custom_attribute_condition_evaluator`? I'm not familiar enough with the SDK to have a strong opinion, and I don't want to put in the time to do the work unless there's some consensus that it's the right thing to do.

PR #289, opened 2019-06-10
Dont use pendingEventsDispatcher with user defined eventDispatcher

jordangarcia:
## Summary
The is an edge case where the `sendPendingEvents()` mechanism keeps sending events because a custom eventDispatcher never invoked the `callback()` denoting the request completed.  This leads to events piling up in localStorage and sending every event each time the SDK is initialized.  

To prevent this only use the pendingEventDispatcher if the default event dispatcher is used, since this guarantees the callback is properly called.

## Test plan
- Added unit tests

## Issues
- OASIS-4887

### DIFF
--- a/packages/optimizely-sdk/lib/index.browser.js
+++ b/packages/optimizely-sdk/lib/index.browser.js
@@ -85,18 +85,26 @@ module.exports = {
         config.skipJSONValidation = true;
       }
 
-      var wrappedEventDispatcher = new eventProcessor.LocalStoragePendingEventsDispatcher({
-        eventDispatcher: config.eventDispatcher || defaultEventDispatcher,
-      });
-      if (!hasRetriedEvents) {
-        wrappedEventDispatcher.sendPendingEvents();
-        hasRetriedEvents = true;
+      var eventDispatcher;
+      // prettier-ignore
+      if (config.eventDispatcher == null) { // eslint-disable-line eqeqeq
+        // only wrap the event dispatcher with pending events retry if the user didnt override
+        eventDispatcher = new eventProcessor.LocalStoragePendingEventsDispatcher({
+          eventDispatcher: defaultEventDispatcher,
+        });
+
+        if (!hasRetriedEvents) {
+          eventDispatcher.sendPendingEvents();
+          hasRetriedEvents = true;
+        }
+      } else {
+        eventDispatcher = config.eventDispatcher;
       }
 
       config = fns.assignIn({
         clientEngine: enums.JAVASCRIPT_CLIENT_ENGINE,
       }, config, {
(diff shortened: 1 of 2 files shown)


packages/optimizely-sdk/lib/index.browser.js
```diff
@@ -85,18 +85,26 @@ module.exports = {
...
-      });
-      if (!hasRetriedEvents) {
-        wrappedEventDispatcher.sendPendingEvents();
-        hasRetriedEvents = true;
+      var eventDispatcher;
+      // prettier-ignore
+      if (config.eventDispatcher == null) { // eslint-disable-line eqeqeq
+        // only wrap the event dispatcher with pending events retry if the user didnt override
+        eventDispatcher = new eventProcessor.LocalStoragePendingEventsDispatcher({
+          eventDispatcher: defaultEventDispatcher,
```
mjc1283: Given that we we no longer allow a custom `eventDispatcher` to be used with `LocalStoragePendingEventsDispatcher`, does passing `eventDispatcher` to `LocalStoragePendingEventsDispatcher` still make sense?

jordangarcia: Hm, you mean the idea of having an event dispatcher that wraps another event dispatcher.

Probably isn't the most ideal design, but I dont hate it.  If we just made this the default event dispatcher, without wrapping it, then we'd be replacing the existing `defaultEventDispatcher` implementation. That could be a bit strange for people using the eventDispatcher in a custom manner.

When we made the decision to only wrap the defaultEventDispatcher, I did feel a bit safer about this pattern then replacement.

WDYT?

packages/optimizely-sdk/lib/index.browser.tests.js
```diff
@@ -73,11 +73,39 @@ describe('javascript-sdk', function() {
...
+          });
+          // Invalid datafile causes onReady Promise rejection - catch this error
+          optlyInstance.onReady().catch(function() {});
+
+          sinon.assert.calledOnce(LocalStoragePendingEventsDispatcher.prototype.sendPendingEvents);
+        });
+      });
+
+      describe('when an eventDispatcher is passed in', function() {
+        it('should wrap the default eventDispatcher and invoke sendPendingEvents', function() {
```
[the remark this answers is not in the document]
jordangarcia: good catch

packages/optimizely-sdk/lib/index.browser.js
mjc1283: That makes sense to me. Initially, I thought this wrapper pattern was weird given that it only ever wraps one thing, but it is safe, and I like that pending-events-related logic is isolated.

PR #298, opened 2019-06-28
Implement getFeatureVariable and create tests

brandonhudavid:
## Summary
- Implement `getFeatureVariable` method to return value of variable with any type 
- Create new unit tests to ensure functionality of `getFeatureVariable`

Since Javascript is a dynamically-typed language, it suffices to have a single method return the value of a feature variable rather than have a separate method for each possible feature variable type.

## Test plan
- Create new unit tests to ensure functionality of `getFeatureVariable`
  - New unit tests similar to tests for `getFeatureVariable` with specific type
- Create gherkin integration tests using FSC and javascript-testapp

## Issues
- Fulfills one of the [SDK convenience APIs](https://docs.google.com/document/d/1MCsMioa4TNImryaDdOJJsZwL9j5zFvFIXODArHF3ORs/edit#) (design doc [here](https://docs.google.com/document/d/1CZ0iappDujwgQzYR8oiU9LRVzLho9jGQt2zaB1rpB48/edit#heading=h.6jynaot9cbnq))

### DIFF
--- a/packages/optimizely-sdk/lib/optimizely/index.js
+++ b/packages/optimizely-sdk/lib/optimizely/index.js
@@ -719,8 +719,10 @@ Optimizely.prototype._getFeatureVariableForType = function(featureKey, variableK
   if (!variable) {
     return null;
   }
-  
-  if (variableType && variable.type !== variableType) {
+
+  if (!variableType) {
+    variableType = variable.type;
+  } else if (variable.type !== variableType) {
     this.logger.log(
       LOG_LEVEL.WARNING,
       sprintf(LOG_MESSAGES.VARIABLE_REQUESTED_WITH_WRONG_TYPE, MODULE_NAME, variableType, variable.type)
@@ -759,7 +761,7 @@ Optimizely.prototype._getFeatureVariableForType = function(featureKey, variableK
     }
   }
 
-  var typeCastedValue = projectConfig.getTypeCastValue(variableValue, variable.type, this.logger);
+  var typeCastedValue = projectConfig.getTypeCastValue(variableValue, variableType, this.logger);
   this.notificationCenter.sendNotifications(
     NOTIFICATION_TYPES.DECISION,
     {
@@ -772,7 +774,7 @@ Optimizely.prototype._getFeatureVariableForType = function(featureKey, variableK
         source: decision.decisionSource,
         variableKey: variableKey,
         variableValue: typeCastedValue,
-        variableType: variable.type,
+        variableType: variableType,
         sourceInfo: sourceInfo,
       }
     }
--- a/packages/optimizely-sdk/lib/optimizely/index.js
+++ b/packages/optimizely-sdk/lib/optimizely/index.js
@@ -648,23 +648,49 @@ Optimizely.prototype.getEnabledFeatures = function(userId, attributes) {
 };
 
 /**
- * Returns value of the variable attached to the given feature flag.
- * Returns null if the feature key or variable key is invalid.
+ * Returns dynamically-typed value of the variable attached to the given
+ * feature flag. Returns null if the feature key or variable key is invalid.
  *
- * @param {string} featureKey   Key of the feature whose variable's value is
- *                              being accessed
- * @param {string} variableKey  Key of the variable whose value is being
- *                              accessed
- * @param {string} userId       ID for the user
- * @param {Object} attributes   Optional user attributes
- * @return {*}                  Value of the variable cast to the appropriate
- *                              type, or null if the feature key is invalid or
- *                              the variable key is invalid
+ * @param {string} featureKey           Key of the feature whose variable's
+ *                                      value is being accessed
+ * @param {string} variableKey          Key of the variable whose value is
+ *                                      being accessed
+ * @param {string} userId               ID for the user
+ * @param {Object} attributes           Optional user attributes
+ * @return {string|boolean|number|null} Value of the variable cast to the appropriate
+ *                                      type, or null if the feature key is invalid or
+ *                                      the variable key is invalid
  */
 
 Optimizely.prototype.getFeatureVariable = function(featureKey, variableKey, userId, attributes) {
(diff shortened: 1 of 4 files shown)


packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -647,6 +647,99 @@ Optimizely.prototype.getEnabledFeatures = function(userId, attributes) {
...
+ * @param {string} variableKey  Key of the variable whose value is being
+ *                              accessed
+ * @param {string} userId       ID for the user
+ * @param {Object} attributes   Optional user attributes
+ * @return {*}                  Value of the variable cast to the appropriate
+ *                              type, or null if the feature key is invalid or
+ *                              the variable key is invalid
+ */
+
+Optimizely.prototype.getFeatureVariable = function(featureKey, variableKey, userId, attributes) {
```
mikeng13: This seems like it duplicates the functionality in the private method `_getFeatureVariableForType`. Let's reuse that method instead of duplicating here by calling it from this method with a `null` value for the type to that private method and having that method skip this check if it sees the given type is null:

```
if (variable.type !== variableType) {
    this.logger.log(	    this.logger.log(
      LOG_LEVEL.WARNING,	      LOG_LEVEL.WARNING,
      sprintf(LOG_MESSAGES.VARIABLE_REQUESTED_WITH_WRONG_TYPE, MODULE_NAME, variableType, variable.type)	      sprintf(LOG_MESSAGES.VARIABLE_REQUESTED_WITH_WRONG_TYPE, MODULE_NAME, variableType, variable.type)
    );	    );
    return null;	    return null;
  }	  }```

File: packages/optimizely-sdk/lib/optimizely/index.js [hunk 2]
```diff
@@ -647,29 +647,50 @@ Optimizely.prototype.getEnabledFeatures = function(userId, attributes) {
...
+ *                                      being accessed
+ * @param {string} userId               ID for the user
+ * @param {Object} attributes           Optional user attributes
+ * @return {string|boolean|number|null} Value of the variable cast to the appropriate
+ *                                      type, or null if the feature key is invalid or
+ *                                      the variable key is invalid
+ */
+
+Optimizely.prototype.getFeatureVariable = function(featureKey, variableKey, userId, attributes) {
+  return this._getFeatureVariableForType(featureKey, variableKey, null, userId, attributes);
```
mikeng13: You should wrap this in a `try/catch` like we do with the other public APIs: https://github.com/optimizely/javascript-sdk/pull/298/files#diff-742c29e45e027dd9699a4d4d2e76c03bR617

This ensures that we also catch any synchronous exceptions that could crash the user's app.

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -692,13 +713,15 @@ Optimizely.prototype._getFeatureVariableForType = function(featureKey, variableK
...
   }
-
-  if (variable.type !== variableType) {
-    this.logger.log(
-      LOG_LEVEL.WARNING,
-      sprintf(LOG_MESSAGES.VARIABLE_REQUESTED_WITH_WRONG_TYPE, MODULE_NAME, variableType, variable.type)
-    );
-    return null;
+  
+  if (variableType) {
```
mikeng13: You can collapse these two ifs together since you don't have an else condition for the inner `if` clause:

```
if (variableType && variable.type !== variableType) 
```

mikeng13:
Also, please merge the latest master into your branch and push that with your next commit.

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -732,7 +759,7 @@ Optimizely.prototype._getFeatureVariableForType = function(featureKey, variableK
     }
   }
 
-  var typeCastedValue = projectConfig.getTypeCastValue(variableValue, variableType, this.logger);
+  var typeCastedValue = projectConfig.getTypeCastValue(variableValue, variable.type, this.logger);
```
aliabbasrizvi: Why was this change needed?

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -647,29 +647,56 @@ Optimizely.prototype.getEnabledFeatures = function(userId, attributes) {
...
+ *                                      in lib/utils/enums/index.js), or null to return the
+ *                                      value of the variable cast to the appropriate type
+ * @param {string} userId               ID for the user
+ * @param {Object} attributes           Optional user attributes
+ * @return {string|boolean|number|null} Value of the variable cast to the appropriate
+ *                                      type, or null if the feature key is invalid, the
+ *                                      variable key is invalid, or there is a mismatch
+ *                                      with the type of the variable
  */
 Optimizely.prototype._getFeatureVariableForType = function(featureKey, variableKey, variableType, userId, attributes) {
```
aliabbasrizvi: You should determine and set `variableType` (if it is null) at the beginning of this method instead of changing too much code downstream.

mikeng13: 👍  I like this approach

packages/optimizely-sdk/lib/optimizely/index.js
brandonhudavid: This change was needed in the case where `getFeatureVariable` was called. In this case, `variableType == null`. Thus, we replace `variableType` with `variable.type`. This change should be fine since line 723 performs a check to ensure that either `variableType == null` or that `variableType == variable.type`.

brandonhudavid: Just read your comment below, makes sense to set `variableType == variable.type` if `variableType == null` initially. I'll add those changes.

mikeng13:
Passing build here: https://travis-ci.com/optimizely/fullstack-sdk-compatibility-suite/builds/119044040 using FSC PR: https://github.com/optimizely/fullstack-sdk-compatibility-suite/pull/303

@aliabbasrizvi please unblock this when you get a chance so we can merge this in and then merge in the FSC PR.

PR #339, opened 2019-08-08
feat(event processor): Send log event notification upon event dispatch when notification center available

mjc1283:
## Summary

With this change, the event processor can be constructed with a notification center. If a notification center is provided, it triggers a log event notification after the event is sent to the event dispatcher.

## Test plan

Added new unit test. Manually tested

## Issues
https://optimizely.atlassian.net/browse/OASIS-4976

### DIFF
--- a/packages/event-processor/package.json
+++ b/packages/event-processor/package.json
@@ -38,7 +38,7 @@
   },
   "dependencies": {
     "@optimizely/js-sdk-logging": "^0.1.0",
-    "@optimizely/js-sdk-utils": "^0.1.0"
+    "@optimizely/js-sdk-utils": "^0.2.0"
   },
   "devDependencies": {
     "@types/jest": "^24.0.9",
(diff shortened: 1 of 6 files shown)


packages/event-processor/package.json
```diff
@@ -38,7 +38,7 @@
   },
   "dependencies": {
     "@optimizely/js-sdk-logging": "^0.1.0",
-    "@optimizely/js-sdk-utils": "^0.1.0"
+    "@optimizely/js-sdk-utils": "^0.2.0"
```
mikeng13: I think these changes are not supposed to be in this PR

mjc1283: I did intend this. I'm updating to 0.2.0 so I can get access to `NOTIFICATION_TYPES` and `NotificationCenter`, new exports from js-sdk-utils. But did you want this update itself to be separated into its own PR?

PR #341, opened 2019-08-09
fix(event processor): Flush current batch when event with different context received

mjc1283:
## Summary

Prior to this change, event processor's buffer could contain events with different contexts. At flush time, the buffered events would be grouped by context and a request would be dispatched for each group.

With this change, we introduce an invariant: all events in the buffer share the same context. As soon as an event with different context is received, the queue is flushed. With this restriction, at most one request is dispatched per flush.

- A new comparator function (`batchComparator`) is provided to `DefaultEventQueue`. When enqueuing an event, `batchComparator` is used to determine whether the incoming event can be included into the current batch. If not, we flush.
- `LogTierV1EventProcessor` provides a `batchComparator` when constructing a `DefaultEventQueue` that checks equality of all context properties.
- In `AbstractEventProcessor`, the grouping step is removed from the flush process.

## Test plan

New unit tests. Manually tested.

## Issues
[OASIS-5160](https://optimizely.atlassian.net/browse/OASIS-5160)

### DIFF
--- a/packages/event-processor/__tests__/eventQueue.spec.ts
+++ b/packages/event-processor/__tests__/eventQueue.spec.ts
@@ -149,7 +149,7 @@ describe('eventQueue', () => {
       queue.stop()
     })
 
-    it('should invoke the sink function when an event incompatable with the current batch (according to batchComparator) is received', () => {
+    it('should invoke the sink function when an item incompatable with the current batch (according to batchComparator) is received', () => {
       const sinkFn = jest.fn()
       const queue = new DefaultEventQueue<number>({
         flushInterval: 100,
--- a/packages/event-processor/__tests__/eventQueue.spec.ts
+++ b/packages/event-processor/__tests__/eventQueue.spec.ts
@@ -56,6 +56,7 @@ describe('eventQueue', () => {
         flushInterval: 100,
         maxQueueSize: -1,
         sink: sinkFn,
+        batchComparator: () => true
       })
 
       queue.start()
@@ -76,6 +77,7 @@ describe('eventQueue', () => {
         flushInterval: 100,
         maxQueueSize: 0,
         sink: sinkFn,
+        batchComparator: () => true
       })
 
       queue.start()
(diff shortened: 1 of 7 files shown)


packages/event-processor/__tests__/eventQueue.spec.ts
```diff
@@ -145,12 +149,33 @@ describe('eventQueue', () => {
       queue.stop()
     })
 
+    it('should invoke the sink function when an item incompatable with the current batch (according to batchComparator) is received', () => {
+      const sinkFn = jest.fn()
+      const queue = new DefaultEventQueue<number>({
+        flushInterval: 100,
+        maxQueueSize: 100,
+        sink: sinkFn,
+        batchComparator: (n1, n2) => (n1 % 2) === (n2 % 2)
```
aliabbasrizvi: nit. Can you add a comment about what this thing is doing? It will be useful for future maintainers.

packages/event-processor/__tests__/v1EventProcessor.spec.ts
```diff
@@ -292,27 +292,63 @@ describe('LogTierV1EventProcessor', () => {
       })
     })
 
-    it("should flush two batches if the event context isn't the same", () => {
+    it('should flush the current batch when it receives an event with a different context revision than the current batch', async () => {
       const impressionEvent1 = createImpressionEvent()
-      const impressionEvent2 = createImpressionEvent()
       const conversionEvent = createConversionEvent()
+      const impressionEvent2 = createImpressionEvent()
```
aliabbasrizvi: nit. Can you add a note that `createImpressionEvent` and `createConversionEvent` create events with revision '1' and so line 300 makes sense with that context.

Or leave a comment for line 300 i.e. setting different revision to be able to test flush.

packages/event-processor/__tests__/eventQueue.spec.ts
mjc1283: I changed it to a more straightforward test checking the first character of a string, and added comments.

PR #342, opened 2019-08-09
fix(event processor): If batchSize = 0 or flushInterval = 0, log that values are invalid and use default values instead

mjc1283:
## Summary

Add/change validation for `maxQueueSize` and `flushInterval` inputs to `AbstractEventProcessor`. When invalid, ignore and use default values.

## Test plan

New unit tests

## Issues
[OASIS-5167](https://optimizely.atlassian.net/browse/OASIS-5167)

### DIFF
--- a/packages/event-processor/src/eventProcessor.ts
+++ b/packages/event-processor/src/eventProcessor.ts
@@ -16,10 +16,7 @@
 // TODO change this to use Managed from js-sdk-models when available
 import { Managed } from './managed'
 import { ConversionEvent, ImpressionEvent } from './events'
-import {
-  EventDispatcher,
-  EventV1Request,
-} from './eventDispatcher'
+import { EventDispatcher, EventV1Request } from './eventDispatcher'
 import { EventQueue, DefaultEventQueue, SingleEventQueue } from './eventQueue'
 import { getLogger } from '@optimizely/js-sdk-logging'
 import { NOTIFICATION_TYPES, NotificationCenter } from '@optimizely/js-sdk-utils'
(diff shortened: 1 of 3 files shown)


packages/event-processor/src/eventProcessor.ts
```diff
@@ -33,26 +33,39 @@ export interface EventProcessor extends Managed {
   process(event: ProcessableEvents): void
 }
 
-const MIN_FLUSH_INTERVAL = 100
+const DEFAULT_FLUSH_INTERVAL = 30000
```
aliabbasrizvi: Can you add a comment about what this value is? i.e. this does not talk about unit so a simple comment about this being 30 seconds will be useful I feel.

packages/event-processor/src/eventProcessor.ts
```diff
@@ -33,26 +33,39 @@ export interface EventProcessor extends Managed {
...
     this.dispatcher = dispatcher
 
+    if (flushInterval <= 0) {
+      logger.warn(`Invalid flushInterval ${flushInterval}, defaulting to ${DEFAULT_FLUSH_INTERVAL}`)
+      flushInterval = DEFAULT_FLUSH_INTERVAL
+    }
+
+    maxQueueSize = Math.floor(maxQueueSize)
+    if (maxQueueSize < 1) {
+      logger.warn(`Invalid maxQueueSize, defaulting to ${DEFAULT_MAX_QUEUE_SIZE}`)
```
aliabbasrizvi: Let's log the invalid `maxQueueSize` value as well?

PR #346, opened 2019-08-15
feat(event processor): Validation and defaults for event processor configuration

mjc1283:
## Summary

This PR adds validation and defaults for event batch size and flush interval. There are different defaults for the browser & node entry points:

Node: batch size 10, flush interval 30000
Browser: batch size 10, flush interval 1000

I also added validation that ignores invalid values and uses these defaults.

## Test plan

Updated unit tests

## Issues
[OASIS-5163](https://optimizely.atlassian.net/browse/OASIS-5163)


packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -38,8 +38,6 @@ var FEATURE_VARIABLE_TYPES = enums.FEATURE_VARIABLE_TYPES;
 var DECISION_NOTIFICATION_TYPES = enums.DECISION_NOTIFICATION_TYPES;
 var NOTIFICATION_TYPES = enums.NOTIFICATION_TYPES;
 
-var DEFAULT_EVENT_MAX_QUEUE_SIZE = 1;
```
mikeng13: Yeah makes sense to move this out into the platform-specific entry-points

PR #347, opened 2019-08-16
feat (event processor/pending events dispatcher): In browser entry point, add event listener for pagehide or unload that calls optimizely.close

mjc1283:
## Summary
With event batching, `optimizely.close()` needs to be called to ensure that events are flushed before the user navigates away.  This PR ensures that optimizely.close() is called when the page unloads by hooking into the `pagehide` event or the `unload` event.

## Test plan
Added a UMD bundle test. Manually tested.

### DIFF
--- a/packages/optimizely-sdk/lib/index.browser.js
+++ b/packages/optimizely-sdk/lib/index.browser.js
@@ -117,17 +117,19 @@ module.exports = {
 
       var optimizely = new Optimizely(config);
 
-      try {
-        var unloadEvent = 'onpagehide' in window ? 'pagehide' : 'unload';
-        window.addEventListener(
-          unloadEvent,
-          function() {
-            optimizely.close();
-          },
-          false
-        );
-      } catch (e) {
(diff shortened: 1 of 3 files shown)


packages/optimizely-sdk/lib/index.browser.js
```diff
@@ -125,7 +131,24 @@ module.exports = {
         config.eventFlushInterval = DEFAULT_EVENT_FLUSH_INTERVAL;
       }
 
-      return new Optimizely(config);
+      var optimizely = new Optimizely(config);
```
mikeng13: I wonder if we should bother making this turnoffable in case the user is really adamant about doing it themselves. I don't see it as a hard blocker for this PR thought. Just good for thought.

mjc1283: That's a fair question. I think this is the right default given that batching will be enabled by default.  I'm willing to wait and see if there's any interest in turning it off from users.

PR #359, opened 2019-09-27
fix(event processor / pending events store): Added a check before accessing localstorage to make it work for React Native

zashraf1985:
## Summary
React Native does not have localStorage. Added a check to make sure it does not throw error in react native. This is a temporary fix to make sure error is not thrown in React Native. Will implement a better fix in the future.

## Test plan
Manually tested thoroughly.


zashraf1985:
@mjc1283 I have added description to changelog. I dont have rights to merge. It would be great if you can go ahead and do the honours on my behalf :slightly_smiling_face:

PR #362, opened 2019-10-18
feat: getOptimizelyConfig API to get static experiments and features data

zashraf1985:
## Summary
Added a `getOptimizelyConfig` API which returns static experiments and features data.

## Test plan
1. Manually tested thoroughly with a node JS client and a browser client.
2. Added Unit Tests.


packages/optimizely-sdk/lib/core/optimizely_config/index.js
```diff
@@ -0,0 +1,103 @@
...
+          variablesMap: variation.variables.reduce(function(variables, variable) {
+            variables[variable.id] = fns.assign({}, {
+              id: variable.id,
+              value: variable.value,
+            }, mergeVariables[variable.id]);
+            return variables;                
+          }, {}),
+        };
+        return variations;
+      }, {}),
```
mjc1283: If the variation has the corresponding feature disabled, the variables array will be empty. To produce the `variationsMap`, we need to iterate over the feature flag's variables. This area may need some refactoring to allow efficient access to the feature flag associated with a given experiment (if such a feature flag exists).

Also, I don't think we necessarily need to use `assign` here. The properties we are interested in are fixed. From the feature flag variable, we need `id`, `key`, `defaultValue`, and `type`. From the variation variable, we need `value`.

packages/optimizely-sdk/lib/core/optimizely_config/index.js
```diff
@@ -0,0 +1,103 @@
...
+      experimentsMap: feature.experimentIds.reduce(function(experiments, experimentId) {
+        experiments[experimentId] = allExperiments[experimentId];
+        return experiments;
+      }, {}),
+      variablesMap: feature.variables.reduce(function(variables, variable) {
+        variables[variable.id] = {
+          id: variable.id,
+          key: variable.key,
+          type: variable.type,
+          value: variable.value,
```
mjc1283: I thin variables inside feature flags have `defaultValue`, not `value`. I recommend developing this against real datafiles generated by the app, to confirm it's working.

packages/optimizely-sdk/lib/core/optimizely_config/index.js
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Checks if an experiment is part of a rollout
+function isRollout(configObj, experimentId) {
+  return configObj.rollouts.some(function(rollout) {
+    return rollout.experiments.some(function(experiment) {
+      return experiment.id === experimentId;
+    });
+  });
+}
```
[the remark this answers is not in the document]
zashraf1985: Done. I have actually refactored the whole file.

packages/optimizely-sdk/lib/core/optimizely_config/index.js
zashraf1985: Refactored the whole file. Modified feature varible merging logic to add default values for variations that are not enabled

packages/optimizely-sdk/lib/core/optimizely_config/index.js
zashraf1985: I tested it against the real datafille generated from an app.optimizely project. I just missed this bit.

packages/optimizely-sdk/lib/core/optimizely_config/index.js
```diff
@@ -0,0 +1,121 @@
...
+        id: experiment.id,
+        key: experiment.key,
+        variationsMap: experiment.variations.reduce(function(variations, variation) {
+          variations[variation.id] = {
+            id: variation.id,
+            key: variation.key,
+            variablesMap: variation.variables.reduce(function(variables, variable) {
+              // return empty variable map when feature is disabled
+              if (!variation.featureEnabled) {
+                return {};
```
mjc1283: Nit: there's no need to iterate over all the variables when `featureEnabled` is false. We could put this check outside the `reduce` and only do the `reduce` when it's `true`.

packages/optimizely-sdk/lib/core/optimizely_config/index.js
```diff
@@ -0,0 +1,121 @@
...
+  }, {});
+}
+
+// Merges feature varibles in variations of passed in experiment
+// Modifies experiment object.
+function mergeFeatureVariables(experiment, featureVariables) {
+  var variations = fns.values(experiment.variationsMap);
+  variations.forEach(function(variation) {
+    featureVariables.forEach(function(featureVariable) {
+      var variable = variation.variablesMap[featureVariable.id];
```
mjc1283: Nit: suggest renaming this
```suggestion
      var variationVariable = variation.variablesMap[featureVariable.id];
```
To clearly distinguish this from `featureVariable`.

packages/optimizely-sdk/lib/core/optimizely_config/index.js
```diff
@@ -0,0 +1,121 @@
...
+
+// Gets map of all experiments
+function getFeaturesMap(configObj, allExperiments) {
+  return configObj.featureFlags.reduce(function(features, feature) {
+    features[feature.id] = {
+      id: feature.id,
+      key: feature.key,
+      experimentsMap: feature.experimentIds.reduce(function(experiments, experimentId) {
+        // Update experiment object with merged feature variables
+        mergeFeatureVariables(allExperiments[experimentId], feature.variables);
```
mjc1283: `mergeFeatureVariables` makes sense to me, but it could be confusing that it's only being called inside `getFeaturesMap`. Until we call `mergeFeatureVariables`, the `experimentsMap` is incomplete, but that's not obvious at a glance. So, I suggest doing `mergeFeatureVariables` on each experiment inside `getExperimentsMap`.

packages/optimizely-sdk/lib/core/optimizely_config/index.js
```diff
@@ -0,0 +1,121 @@
...
+    }));
+  }, []);
+}
+
+// Gets Map of all experiments except rollouts
+function getExperimentsMap(configObj) {
+  var rolloutExperimentsIds = getRolloutExperimentIds(configObj.rollouts);
+  return configObj.experiments.reduce(function(experiments, experiment) {
+    // skip experiments that are part of a rollout
+    if (!rolloutExperimentsIds.includes(experiment.id)) {
```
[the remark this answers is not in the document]
zashraf1985: Done

packages/optimizely-sdk/lib/core/optimizely_config/index.js
```diff
@@ -0,0 +1,121 @@
...
+      };
+    }
+    return experiments;
+  }, {});
+}
+
+// Merges feature varibles in variations of passed in experiment
+// Modifies experiment object.
+function mergeFeatureVariables(experiment, featureVariables) {
+  var variations = fns.values(experiment.variationsMap);
```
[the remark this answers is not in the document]
zashraf1985: Done

packages/optimizely-sdk/lib/core/optimizely_config/index.js
zashraf1985: Done

packages/optimizely-sdk/lib/core/optimizely_config/index.js
zashraf1985: Done

packages/optimizely-sdk/lib/core/optimizely_config/index.js
zashraf1985: Done

packages/optimizely-sdk/lib/core/optimizely_config/index.js
zashraf1985: Done

packages/optimizely-sdk/lib/core/optimizely_config/index.js
```diff
@@ -0,0 +1,128 @@
...
+function getExperimentsMap(configObj) {
+  var rolloutExperimentIds = getRolloutExperimentIds(configObj.rollouts);
+  var featureVariablesMap = configObj.featureFlags.reduce(function(resultMap, feature){
+    resultMap[feature.id] = feature.variables;
+    return resultMap;
+  }, {});
+  return configObj.experiments.reduce(function(experiments, experiment) {
+    // skip experiments that are part of a rollout
+    if (!rolloutExperimentIds[experiment.id]) {
+      experiments[experiment.id] = {
```
jaeopt: All maps in OptimizelyConfig should be mapped by "key" not "id". All data values look good. Just replace the keys.

packages/optimizely-sdk/lib/core/optimizely_config/index.js
```diff
@@ -0,0 +1,128 @@
...
+                id: variable.id,
+                value: variable.value,
+              };
+              return variables;
+            }, {});
+          }
+          variations[variation.id] = {
+            id: variation.id,
+            key: variation.key,
+            variablesMap,
```
jaeopt: Spec changed: need to add "featureEnabled: Bool" for all variations. Please see the spec doc.

zashraf1985: done

packages/optimizely-sdk/lib/core/optimizely_config/index.js
zashraf1985: Done

PR #366, opened 2019-11-06
fix: log error message when user allocated to vacated space in group or experiment

fayyazarshad:
**Summary**

When a user is allocated to space in a group that used to be occupied by an experiment, but is now unallocated, the JS SDK logs an error message. This is not an error, so it shouldn't log an error message.

**Expected:**
An info level message saying “User not in any experiment in group” should be logged

**Actual:**
An error level message saying 'Experiment ID " " is not in the datafile' is logged.

### DIFF
--- a/packages/optimizely-sdk/lib/core/bucketer/index.js
+++ b/packages/optimizely-sdk/lib/core/bucketer/index.js
@@ -87,7 +87,7 @@ module.exports = {
     bucketerParams.logger.log(LOG_LEVEL.DEBUG, bucketedUserLogMessage);
 
     var entityId = module.exports._findBucket(bucketValue, bucketerParams.trafficAllocationConfig);
-    if (entityId === null || entityId === '') {
+    if (!entityId) {
       var userHasNoVariationLogMessage = sprintf(LOG_MESSAGES.USER_HAS_NO_VARIATION, MODULE_NAME, bucketerParams.userId, bucketerParams.experimentKey);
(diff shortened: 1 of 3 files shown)


packages/optimizely-sdk/lib/core/bucketer/index.js
```diff
@@ -87,10 +87,10 @@ module.exports = {
     bucketerParams.logger.log(LOG_LEVEL.DEBUG, bucketedUserLogMessage);
 
     var entityId = module.exports._findBucket(bucketValue, bucketerParams.trafficAllocationConfig);
-    if (entityId === null) {
+    if (entityId === null || entityId === '') {
```
zashraf1985: This whole check can probably be changed to `if (!entityId)` as null and empty both evaluate to false. Do test it thoroughly though

packages/optimizely-sdk/lib/core/decision_service/index.js
```diff
@@ -349,7 +349,7 @@ DecisionService.prototype._getVariationForFeatureExperiment = function(configObj
 
 DecisionService.prototype._getExperimentInGroup = function(configObj, group, userId) {
   var experimentId = bucketer.bucketUserIntoExperiment(group, userId, userId, this.logger);
-  if (experimentId !== null) {
+  if (experimentId !== null && experimentId !== '') {
```
zashraf1985: Same as above. This can also be checked as `!experimentId`

PR #372, opened 2019-11-25
feat: triggering react native tests on FSC

zashraf1985:
todo: Add details here


zashraf1985:
This is not needed as react-native-testapp will not be triggered through FSC. creating a separate PR to trigger react-native-testapp directly from this sdk

PR #374, opened 2019-12-13
feat: Run tests for sub packages in CI

zashraf1985:
## Summary
Travis CI Only run tests for `optimizely-sdk` package. This PR makes Travis CI run unit tests for other sub packages on `node 12`

## Test Plan
Tested on Travis CI


.travis.yml
```diff
@@ -52,3 +53,14 @@ jobs:
     - stage: Cross-browser and umd unit tests
       node_js: '8'
       script: npm run test-ci
+    - &packagetest
+      stage: 'Test sub packages'
+      node_js: '12'
+      before_install: cd packages/datafile-manager
+      before_script: tsc
```
mjc1283: Is this `tsc` necessary?

packages/utils/src/index.ts
```diff
@@ -62,7 +62,7 @@ export function objectValues<K>(obj: { [key: string]: K }): K[] {
   return Object.keys(obj).map(key => obj[key])
 }
 
-export function objectEntries<K>(obj: { [key: string]: K }): [string, K][] {
+export function objectEntries<K>(obj: { [key: string]: K }): (string | K)[][] {
```
mjc1283: What is this change?

zashraf1985: This is fixing a tsc compilation error. return type was not correct.

.travis.yml
zashraf1985: not needed. i just checked. looks like jest automatically compiles typescript before use

packages/logging/package.json
```diff
@@ -42,6 +42,7 @@
   "devDependencies": {
     "@types/jest": "^23.3.12",
     "jest": "^23.6.0",
-    "ts-jest": "^23.10.5"
+    "ts-jest": "^23.10.5",
+    "typescript": "^3.7.3"
```
mjc1283: Please change typescript dependency versions to `"3.3.3333"` in all packages

PR #376, opened 2019-12-17
feat: Added caching for optimizelyConfig object

fayyazarshad:
## Summary
Updated the `OptimizelyConfig` implementation to always return cached object. `OptimizelyConfig` object will only be updated when new `ProjectConfig` is received.

## Test Plan
Added Unit tests

### DIFF
--- a/packages/optimizely-sdk/lib/core/project_config/project_config_manager.tests.js
+++ b/packages/optimizely-sdk/lib/core/project_config/project_config_manager.tests.js
@@ -420,7 +420,6 @@ describe('lib/core/project_config/project_config_manager', function() {
         // validate it should return the existing optimizely config
         manager.getOptimizelyConfig();
         sinon.assert.calledOnce(optimizelyConfig.getOptimizelyConfig);
-
         // create config with new revision
         var newConfig = testData.getTestProjectConfig();
         newConfig.revision = "43";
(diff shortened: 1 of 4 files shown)


packages/optimizely-sdk/lib/core/project_config/project_config_manager.tests.js
```diff
@@ -407,5 +408,26 @@ describe('lib/core/project_config/project_config_manager', function() {
...
+        // creating optimizely config once project config manager for the first time
+        sinon.assert.calledOnce(optimizelyConfig.getOptimizelyConfig);
+        // validate it should return the existing optimizely config
+        manager.getOptimizelyConfig();
+        sinon.assert.calledOnce(optimizelyConfig.getOptimizelyConfig);
+        // create config with new revision
+        var newConfig = testData.getTestProjectConfig();
+        newConfig.revision = "43";
+        // verify the optimizely config is updated
+        manager.__handleNewConfigObj(newConfig);
```
mjc1283: This violates the project config manager abstraction, `__handleNewConfigObj` should not be called directly

packages/optimizely-sdk/lib/core/project_config/project_config_manager.tests.js
```diff
@@ -407,5 +408,26 @@ describe('lib/core/project_config/project_config_manager', function() {
         });
       });
     });
+
+    describe('test caching of optimizely config', function() {
+      it('should return the same config until revision is changed', function() {
+        sinon.stub(optimizelyConfig, "getOptimizelyConfig");
```
mjc1283: Stub setup and restoration should be in `beforeEach` and `afterEach` callbacks.

zashraf1985:
@mjc1283 Thanks for approving. Can you also merge it because i dont have rights to do so.

PR #377, opened 2019-12-17
fix: Lint errors and global variable declaration in optimizely_config

mjc1283:
## Summary
Fix lint errors and global variable declaration in optimizely_config

## Test plan
Existing tests


packages/optimizely-sdk/lib/core/optimizely_config/index.js
```diff
@@ -111,9 +111,9 @@ module.exports = {
     // Fetch all feature variables from feature flags to merge them with variation variables
     var experimentsMap = getExperimentsMap(configObj);
     return {
-      experimentsMap,
+      experimentsMap: experimentsMap,
```
mikeng13: Oh wow this is actually a bug if run in the browser right?

mjc1283: Most browsers now support it, but if we are sticking to our es5-only requirement until the next major version, we can't do this.

PR #392, opened 2020-01-30
chore(notification_center)[OASIS-5764]: Add all unit tests for the notification_center module

alibaker-optimizely:
## Summary
- The notification_center module did not have a unit test associated with it.
- This PR addresses that to ensure 100% coverage of this module.

## Test plan
addNotificationListener
- only valid listener types should have listeners registered
- the listenerId should be returned upon successful registration
- cannot register the same callback twice

removeNotificationListener
- should return true if the listener is successfully removed
- should only remove the specified listener using the listenerId

clearAllNotificationListeners
- should remove all listeners of all types

clearNotificationListeners
- should remove only the specified notification type when there are no listeners registered
- should remove only the specified notification type and leave all other notification type listeners still registered

sendNotifications
- checks that each notification type listeners executes the callback with the passed in notification data from the user

## Issues
- Addresses [OASIS-5764](https://optimizely.atlassian.net/browse/OASIS-5764)


packages/optimizely-sdk/lib/core/notification_center/index.tests.js
```diff
@@ -0,0 +1,962 @@
...
+                it('should return -1 if notification type is not a valid one', function(){
+                    
+                    listenerId = notificationCenterInstance.addNotificationListener(
+                        INVALID_LISTENER_TYPE, 
+                        genericCallbackSpy
+                    );
+                    assert.strictEqual(listenerId, -1);
+                });
+
+                it('should not call the callback function if notification type is not a valid one', function(){
```
mjc1283: I'm not sure we need this test case. Even if type was valid, the function would not be called anyway, right?

packages/optimizely-sdk/lib/core/notification_center/index.tests.js
```diff
@@ -0,0 +1,962 @@
...
+                });
+
+                // should return an id of notification listener (listenerId) if ACTIVATE listener is not already added 
+                it('should return an id (listenerId) > 0 of the notification listener if ACTIVATE callback is not already added', function(){
+                    listenerId = notificationCenterInstance.addNotificationListener(
+                        enums.NOTIFICATION_TYPES.ACTIVATE,
+                        activateCallbackSpy1
+                    );
+
+                    assert.isAbove(listenerId, 0);
```
mjc1283: Nice, I didn't know about `isAbove` 👍

packages/optimizely-sdk/lib/core/notification_center/index.tests.js
```diff
@@ -0,0 +1,962 @@
...
+                        decisionCallbackSpy1
+                    );
+                    
+                    assert.strictEqual(notificationCenterInstance.addNotificationListener(
+                        enums.NOTIFICATION_TYPES.DECISION,
+                        decisionCallbackSpy1
+                    ), -1);
+                });
+
+                // should return an id of notification listener (listenerId) if DECISION listener is not already added
```
mjc1283: Let's remove these comments that are the same as the string passed to `it`.

packages/optimizely-sdk/lib/core/notification_center/index.tests.js
```diff
@@ -0,0 +1,962 @@
...
+                it('should return an id (listenerId) > 0 of the notification listener if LOG_EVENT callback is not already added', function(){
+                    listenerId = notificationCenterInstance.addNotificationListener(
+                        enums.NOTIFICATION_TYPES.LOG_EVENT,
+                        logEventCallbackSpy1
+                    );
+
+                    assert.isAbove(listenerId, 0);
+                });
+
+                // LOG_EVENT: should return callback with object argument properties of: url {str}, httpVerb {str}, params {obj} when an impression or conversion is sent
```
mjc1283: Out of place comment?

packages/optimizely-sdk/lib/core/notification_center/index.tests.js
```diff
@@ -0,0 +1,962 @@
...
+                        enums.NOTIFICATION_TYPES.DECISION,
+                        decisionCallbackSpy1
+                    );
+
+                    assert.isAbove(listenerId, 0);
+                });
+            });
+
+            context('add notification listener for type LOG_EVENT', function(){
+                it('should return -1 if that same LOG_EVENT callback is already added', function(){
```
mjc1283: Nice work covering all notification types 👍 that's very valuable. But I think it might be better to structure test cases according to functionality rather than notification type. We could have a single `it('should return -1 if that same callback is already added`, ...)` which can exercise all notification types.

packages/optimizely-sdk/lib/core/notification_center/index.tests.js
```diff
@@ -0,0 +1,962 @@
...
+var enums = require('../../utils/enums');
+var fns = require('../../utils/fns');
+var sprintf = require('@optimizely/js-sdk-utils').sprintf;
+var LOG_LEVEL = enums.LOG_LEVEL;
+var LOG_MESSAGES = enums.LOG_MESSAGES;
+var chai = require('chai');
+var sinon = require('sinon');
+var assert = chai.assert;
+
+describe('lib/core/notification_center', function() {
```
mjc1283: The standard in optimizely-sdk is 2 space indent. We are not enforcing this with the linter or talking about it in README or CONTRIBUTING, so we need to address that. Apologies that I didn't think to mention this in our previous discussions. Let's discuss how to fix this.

alibaker-optimizely: Thanks for letting me know! I did run the linter, so +1 on adding that information somewhere -- should I add that when I make these adjustments? I'll update my IDE as well.

packages/optimizely-sdk/lib/core/notification_center/index.tests.js
```diff
@@ -0,0 +1,962 @@
+/****************************************************************************
+ * Copyright 2017-2020, Optimizely, Inc. and contributors                        *
```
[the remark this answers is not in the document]
alibaker-optimizely: Ohh that makes sense! Will be fixed in my next push of changes.

packages/optimizely-sdk/lib/core/notification_center/index.tests.js
alibaker-optimizely: Very much so!

packages/optimizely-sdk/lib/core/notification_center/index.tests.js
alibaker-optimizely: created a task for the linter rule: https://optimizely.atlassian.net/browse/OASIS-5917

packages/optimizely-sdk/lib/core/notification_center/index.tests.js
alibaker-optimizely: **I'd like to review my reasoning here so we can get at the root cause of why I thought this test made sense at the time** 
Since a callback is passed in as one of the args, I thought it made sense as a test case on this basis. We are handing some piece of data to the function so it feels like we should check what happens to it. In other words, ensuring the function isn't somehow handling the callback in some weird way. Does this make sense? This perspective is not informed by anything outside of my own interpretation of what the function is expected to do, so if this is not really a convention/necessity in testing, I definitely want to fix this! 

**Again, want to check my assumptions here. Question in response to**

> Even if type was valid, the function would not be called anyway, right?

Even though the callback isn't expected to be called, it is expected to be registered and eventually triggered by `sendNotifications`, but the creation of the callback for the `NotificationCenter` instance does happen here. The only way I could think to fully test that the function properly sets the callback to the type in the `__notificationListeners` object (which is something expected of this function) is to call sendNotifications and check if the callback was executed. Or is it sufficient or more proper to defer the callback testing to `sendNotifications` (which it currently does do by checking the callback and the args passed in?

packages/optimizely-sdk/lib/core/notification_center/index.tests.js
```diff
@@ -0,0 +1,962 @@
...
+            activateCallbackSpy1.reset();
+            activateCallbackSpy2.reset();
+            decisionCallbackSpy1.reset();
+            decisionCallbackSpy2.reset();
+            logEventCallbackSpy1.reset();
+            logEventCallbackSpy2.reset();
+            configUpdateCallbackSpy1.reset();
+            configUpdateCallbackSpy2.reset();
+            trackCallbackSpy1.reset();
+            trackCallbackSpy2.reset();
```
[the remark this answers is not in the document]
alibaker-optimizely: Oh man that is awesome! Thanks for showing this to me! I'm going to refactor and use the sandbox for sure.

packages/optimizely-sdk/lib/core/notification_center/index.tests.js
alibaker-optimizely: Having just finished refactoring, this does look much better and makes a lot more sense. Thank you for the direction on this!

packages/optimizely-sdk/lib/core/notification_center/index.tests.js
```diff
@@ -0,0 +1,962 @@
...
+        beforeEach(function(){
+            mockLoggerStub = sinon.stub(mockLogger, 'log');
+            mockErrorHandlerStub = sinon.stub(mockErrorHandler, 'handleError');
+    
+            notificationCenterInstance = NotificationCenter.createNotificationCenter({
+                logger: mockLoggerStub,
+                errorHandler: mockErrorHandlerStub
+            }); 
+
+            activateCallbackSpy1 = sinon.spy();
```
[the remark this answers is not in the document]
alibaker-optimizely: done! I've removed global spies and created local ones in each it() statement as needed. This does improve readability. Thanks!

PR #395, opened 2020-02-14
Removed dependency on lodash.assign

zashraf1985:
## Summary
to reduce package size, we are trying to gradually get rid of lodash library. This PR replaces `assign` function with `Object.assign`. Also added a polyfill method to support old browsers. Polyfill was taken from MDN. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Polyfill

## Test Plan
All tests pass after this change.

### DIFF
--- a/packages/optimizely-sdk/lib/utils/fns/index.js
+++ b/packages/optimizely-sdk/lib/utils/fns/index.js
@@ -18,7 +18,10 @@ var _isFinite = require('lodash/isFinite');
 var MAX_NUMBER_LIMIT = Math.pow(2, 53);
 
 module.exports = {
-  assign: require('lodash/assign'),
+  assign: function (target) {
+    // Object.assign crashes if target object is undefined or null
+    return !!target ? Object.assign.apply(Object, arguments) : {};
+  },
   assignIn: require('lodash/assignIn'),
   cloneDeep: require('lodash/cloneDeep'),
   currentTimestamp: function() {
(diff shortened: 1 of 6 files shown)


packages/optimizely-sdk/lib/utils/fns/index.js
```diff
@@ -17,8 +17,42 @@ var uuid = require('uuid');
 var _isFinite = require('lodash/isFinite');
 var MAX_NUMBER_LIMIT = Math.pow(2, 53);
 
+// Polyfills for older browsers
+var applyPolyfills = function() {
+  if (typeof Object.assign !== 'function') {
+    // Must be writable: true, enumerable: false, configurable: true
+    Object.defineProperty(Object, "assign", {
```
mjc1283: We should not mutate global objects. I suggest to compare these 3 options and choose the one that results in smallest bundle size:

1. Include this same helper function, but just call it if `Object.assign` isn't available, instead of mutating the global `Object`.
2. Use object-assign library from npm (https://www.npmjs.com/package/object-assign)
3. Use TypeScript with object spread syntax

zashraf1985: Using typescript for only this looked like an overkill to me. I went for option 1 which will have no implact on the bundle size. I also modified all lodash.assignIn usages to use assign instead because it looks like we are not dealing with inherited properties here.

PR #396, opened 2020-02-15
refactor: Removed dependency on lodash.values

zashraf1985:
## Summary
Replaced lodash values function with `objectValues` from js-sdk-utils. This is part of an effort to remove lodash dependency to reduce bundle size.

## Test Plan
All Unit tests and Full Stack SDK compatibility passed


zashraf1985:
@mjc1283 I removed values from fns altogether. there was only one usage, i used objectValues from js-sdk-utils directly there

PR #402, opened 2020-02-18
Removed dependency on lodash.keyBy

zashraf1985:
## Summary
Replaced lodash values function with `keyBy` from js-sdk-utils and added missing functionality. This is part of an effort to remove lodash dependency to reduce bundle size.

## Test Plan
1) Added Unit tests
2) All Unit tests and Full Stack SDK compatibility passed

### DIFF
--- a/packages/optimizely-sdk/lib/utils/fns/index.js
+++ b/packages/optimizely-sdk/lib/utils/fns/index.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017, 2019, Optimizely
+ * Copyright 2017, 2019-2020, Optimizely
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


packages/optimizely-sdk/lib/utils/fns/index.js
```diff
@@ -29,7 +30,15 @@ module.exports = {
   isFinite: function(number) {
     return _isFinite(number) && Math.abs(number) <= MAX_NUMBER_LIMIT;
   },
-  keyBy: require('lodash/keyBy'),
+  keyBy: function(arr, callback) {
+    if (!arr) return {};
+    if (typeof callback === 'string' || callback instanceof String) {
```
mjc1283: In our usages of `keyBy` we are always passing string as 2nd argument. We have no need to support callback as a second argument.

zashraf1985: This is Done

zashraf1985:
This PR was stuck because of a github issue. I created a new one from the same commit and got that merged. This one is no longer needed now. closing it.

PR #407, opened 2020-02-20
Removed dependency on lodash.isNumber

fayyazarshad:
## Summary
to reduce package size, we are trying to gradually get rid of lodash library. This PR replaces isNumber function with simple implementation for numbers and strings.

## Test plan
Added unit tests

### DIFF
--- a/packages/optimizely-sdk/lib/utils/fns/index.js
+++ b/packages/optimizely-sdk/lib/utils/fns/index.js
@@ -35,6 +35,6 @@ module.exports = {
   },
   values: require('lodash/values'),
   isNumber: function(value) {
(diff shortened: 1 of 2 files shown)


packages/optimizely-sdk/lib/utils/fns/index.js
```diff
@@ -38,5 +38,7 @@ module.exports = {
     return uuid.v4();
   },
   values: require('lodash/values'),
-  isNumber: require('lodash/isNumber'),
+  isNumber: function(value) {
+    return typeof value == 'number';
```
mjc1283:
```suggestion
    return typeof value === 'number';
```

PR #408, opened 2020-02-20
Removed dependency on lodash.isFinite

fayyazarshad:
## Summary
to reduce package size, we are trying to gradually get rid of lodash library. This PR replaces isFinite function with simple implementation numbers. As we are using this methods for numbers or strings only

## Test plan
Update Unit Tests

### DIFF
--- a/packages/optimizely-sdk/lib/utils/fns/index.js
+++ b/packages/optimizely-sdk/lib/utils/fns/index.js
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 var uuid = require('uuid');
-var MAX_NUMBER_LIMIT = Math.pow(2, 53);
+var MAX_SAFE_INTEGER_LIMIT = Math.pow(2, 53);
 var keyBy = require('@optimizely/js-sdk-utils').keyBy;
 module.exports = {
   assign: function (target) {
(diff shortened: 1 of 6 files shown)


packages/optimizely-sdk/lib/utils/fns/index.js
```diff
@@ -27,7 +26,7 @@ module.exports = {
   isArray: require('lodash/isArray'),
   isEmpty: require('lodash/isEmpty'),
   isFinite: function(number) {
```
mjc1283: Please rename this to `isSafeInteger`.

packages/optimizely-sdk/lib/utils/fns/index.js
```diff
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 var uuid = require('uuid');
-var _isFinite = require('lodash/isFinite');
 var MAX_NUMBER_LIMIT = Math.pow(2, 53);
```
mjc1283: Please rename this to `MAX_SAFE_INTEGER_LIMIT`.

packages/optimizely-sdk/lib/utils/fns/index.js
zashraf1985: @fayyazarshad

packages/optimizely-sdk/lib/utils/fns/index.js
zashraf1985: @fayyazarshad

PR #409, opened 2020-02-20
Removed dependency on lodash.map

zashraf1985:
## Summary
To reduce package size, we are trying to gradually get rid of lodash library. This PR removes the `map` function.

## Test plan
All tests pass after this change.

### DIFF
--- a/packages/optimizely-sdk/lib/utils/fns/index.js
+++ b/packages/optimizely-sdk/lib/utils/fns/index.js
@@ -33,7 +33,17 @@ module.exports = {
   filter: require('lodash/filter'),
   forEach: require('lodash/forEach'),
   forOwn: require('lodash/forOwn'),
-  map: require('lodash/map'),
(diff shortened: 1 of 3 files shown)


packages/optimizely-sdk/lib/utils/fns/index.js
```diff
@@ -33,7 +33,17 @@ module.exports = {
   filter: require('lodash/filter'),
   forEach: require('lodash/forEach'),
   forOwn: require('lodash/forOwn'),
-  map: require('lodash/map'),
+  map: function(obj, callback) {
```
mjc1283: This `map` function seems unnecessary. I suggest we just update all the call sites to do what's necessary.

zashraf1985: There was only one use. I updated that to use Object.keys and removed map method from fns altogether

PR #412, opened 2020-02-21
Removed dependency on lodash.forEach

fayyazarshad:
## Summary
to reduce package size, we are trying to gradually get rid of lodash library. This PR removes the implementation of fns.forEach and replaced it with default forEach method of arrays and Object.Keys for maps

## Test plan
All tests are passing

### DIFF
--- a/packages/optimizely-sdk/lib/core/notification_center/index.js
+++ b/packages/optimizely-sdk/lib/core/notification_center/index.js
@@ -66,7 +66,7 @@ NotificationCenter.prototype.addNotificationListener = function (notificationTyp
     }
 
     var callbackAlreadyAdded = false;
-    fns.forEach(this.__notificationListeners[notificationType], function (listenerEntry) {
+    (this.__notificationListeners[notificationType] || []).forEach(function(listenerEntry) {
       if (listenerEntry.callback === callback) {
         callbackAlreadyAdded = true;
         return false;
@@ -102,7 +102,7 @@ NotificationCenter.prototype.removeNotificationListener = function (listenerId) 
     var indexToRemove;
     var typeToRemove;
     fns.forOwn(this.__notificationListeners, function (listenersForType, notificationType) {
-      fns.forEach(listenersForType, function (listenerEntry, i) {
+      (listenersForType || []).forEach(function(listenerEntry, i) {
         if (listenerEntry.id === listenerId) {
           indexToRemove = i;
           typeToRemove = notificationType;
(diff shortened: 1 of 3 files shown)


packages/optimizely-sdk/lib/utils/fns/index.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
     return _isFinite(number) && Math.abs(number) <= MAX_NUMBER_LIMIT;
   },
   keyBy: require('lodash/keyBy'),
-  forEach: require('lodash/forEach'),
+  filter: require('lodash/filter'),
```
mjc1283: Filter was removed in another PR, please update from master.

packages/optimizely-sdk/lib/core/notification_center/index.js
```diff
@@ -102,7 +102,7 @@ NotificationCenter.prototype.removeNotificationListener = function (listenerId)
     var indexToRemove;
     var typeToRemove;
     fns.forOwn(this.__notificationListeners, function (listenersForType, notificationType) {
-      fns.forEach(listenersForType, function (listenerEntry, i) {
+      (listenersForType || []).forEach(function(listenerEntry, i) {
```
mjc1283: Missed this earlier - this changes behavior as `forEach` does not exit loop when callback returns `false`.

packages/optimizely-sdk/lib/core/notification_center/index.js
```diff
@@ -102,7 +102,7 @@ NotificationCenter.prototype.removeNotificationListener = function (listenerId)
     var indexToRemove;
     var typeToRemove;
     fns.forOwn(this.__notificationListeners, function (listenersForType, notificationType) {
-      fns.forEach(listenersForType, function (listenerEntry, i) {
+      (listenersForType || []).every(function(listenerEntry, i) {
```
mjc1283: `every` will stop iteration as soon as the callback returns a falsy value. Here, it will always stop after the first item, because the callback either returns `false` or `undefined`. To continue iteration, `true` must be returned. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

fayyazarshad: Good catch, I am wondering how the tests were getting passed. I will looking into it further for tests

fayyazarshad: Good catch, I am wondering how the tests were getting passed. I will looking into it further for tests

PR #415, opened 2020-02-25
refactor: Removed dependency on lodash.cloneDeep

fayyazarshad:
## Summary
To reduce package size, we are gradually removing lodash library. This PR removes `lodash.cloneDeep` from optimizely-sdk package. It is still part of the repository but it is only being used to run tests. lodash has been removed from dependencies and moved to dev-dependencies in package.json

## Test plan
All unit tests and Full Stack compatibility tests pass after this change

### DIFF
--- a/packages/optimizely-sdk/lib/core/project_config/index.js
+++ b/packages/optimizely-sdk/lib/core/project_config/index.js
@@ -35,7 +35,7 @@ module.exports = {
    * @return {Object} Object representing project configuration
    */
   createProjectConfig: function(datafile) {
-    var projectConfig = fns.cloneDeep(datafile);
+    var projectConfig = fns.assign({}, datafile);
 
     /*
      * Conditions of audiences in projectConfig.typedAudiences are not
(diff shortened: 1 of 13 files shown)


packages/optimizely-sdk/lib/utils/fns/index.js
```diff
@@ -20,7 +20,9 @@ var MAX_NUMBER_LIMIT = Math.pow(2, 53);
 module.exports = {
   assign: require('lodash/assign'),
   assignIn: require('lodash/assignIn'),
-  cloneDeep: require('lodash/cloneDeep'),
+  cloneDeep: function(object) {
+    return JSON.parse(JSON.stringify(object));
```
zashraf1985: This is not a true alternative of deepClone because
1. It will not copy functions.
2. It can mess up some datatypes.
But in my opinion, it works for our case because we are merely using it to clone objects created from JSON strings or for objects where we are only concerned with value properties and not functions. My suggestion is to use this implementation but give it a less fancy / non-misleading name. May be cloneJsonObject or cloneDeepProperties etc. What do you suggest @mjc1283

packages/optimizely-sdk/lib/core/project_config/index.js
```diff
@@ -35,7 +35,7 @@ module.exports = {
    * @return {Object} Object representing project configuration
    */
   createProjectConfig: function(datafile) {
-    var projectConfig = fns.cloneDeep(datafile);
+    var projectConfig = datafile;
```
mjc1283:
```suggestion
    var projectConfig = fns.assign({}, datafile);
```

PR #417, opened 2020-02-26
refactor: Removed Promise Polyfill to reduce bundle size

zashraf1985:
## Summary
Removed Promise Polyfill from browser entry point to reduce bundle size. A user can still use the SDK in older browsers by including the promise polyfill themselves.

## Test plan
All tests pass by providing promise polyfill as part of the test setup.


zashraf1985:
> I think we need to remove the `promise-polyfill` library from our dependencies or at least move it to `dev-dependencies` since we use it for testing: https://github.com/optimizely/javascript-sdk/blob/master/packages/optimizely-sdk/package.json#L43

Yes you are right. This should have been moved to dev dependencies. This is done now.

PR #418, opened 2020-03-03
feat(core)[OASIS-6054]: Change Error to LOG_LEVEL.WARNING for unfound event

circAssimilate:
## Summary
- Log a warning when an event key is not found in the datafile instead of throwing an Error
- Log a warning mentioning that user tracking is not happening
- Update actual error logger in catch to be an ERROR instead of INFO
- Updated CONTRIBUTING.md instructions with CHANGELOG.md update instructions

In a similar vein as https://github.com/optimizely/java-sdk/pull/361, the JavaScript SDK noise could be more configurable for use with integrations like Segment.

_Note:_ Please pardon the prettier noise. I tried to do a prep PR in https://github.com/optimizely/javascript-sdk/pull/419 but there are still changes when I save. The actual functionality changes are in https://github.com/optimizely/javascript-sdk/pull/418/commits/646665d0e8275dfa07dd67153a4b78485e865715.

## Test plan
- Updated unit test to cover new functionality
- Ensure other tests pass

## Issues
- [OASIS-6054](https://optimizely.atlassian.net/browse/OASIS-6054)


packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -57,7 +57,10 @@ var DEFAULT_ONREADY_TIMEOUT = 30000;
 function Optimizely(config) {
   var clientEngine = config.clientEngine;
   if (enums.VALID_CLIENT_ENGINES.indexOf(clientEngine) === -1) {
-    config.logger.log(LOG_LEVEL.INFO, jsSdkUtils.sprintf(LOG_MESSAGES.INVALID_CLIENT_ENGINE, MODULE_NAME, clientEngine));
+    config.logger.log(
```
circAssimilate: Most of this file's updates are Prettier changes. Here's the only updated part:

https://github.com/optimizely/javascript-sdk/pull/418/files#diff-742c29e45e027dd9699a4d4d2e76c03bR277-R318

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -46,11 +46,11 @@ var DECISION_SOURCES = enums.DECISION_SOURCES;
 var DECISION_NOTIFICATION_TYPES = enums.DECISION_NOTIFICATION_TYPES;
 var FEATURE_VARIABLE_TYPES = enums.FEATURE_VARIABLE_TYPES;
 
-describe('lib/optimizely', function() {
+describe('lib/optimizely', function () {
```
circAssimilate: Most of this file's updates are Prettier changes. Here's the only updated part:

https://github.com/optimizely/javascript-sdk/pull/418/files#diff-f6d6b958bf1d05fe2016671164857766R1331-R1349

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -285,7 +291,10 @@ Optimizely.prototype.track = function (eventKey, userId, attributes, eventTags)
     }
 
     if (!projectConfig.eventWithKeyExists(configObj, eventKey)) {
-      throw new Error(jsSdkUtils.sprintf(ERROR_MESSAGES.INVALID_EVENT_KEY, MODULE_NAME, eventKey));
+      this.logger.log(
```
jamesopti: So we still need to skip the rest of this block (the part that actually sends the event) when the event key is missing.

We can either:

- return early here
- continue to throw but catch this specific error and use a different log level

I'm inclined to do the latter since, in the case of a missing event OR a real error with the track call, we still log an info level message (see `failedTrackLogMessage` below).


@mjc1283 I dont really understand the need to log the aforementioned `failedTrackLogMessage` at info level when the error was logged right before. Can we consolidate?

Track success-> INFO
Event missing -> WARNING
Track error -> ERROR

circAssimilate: Oh, so we should actually skip sending the event? Wouldn't we want to send it anyway to support tracking events with a backdoor like approach?

mjc1283: @circAssimilate : I agree with @jamesopti, we should not send the event.

@jamesopti 
> I dont really understand the need to log the aforementioned failedTrackLogMessage at info level when the error was logged right before. Can we consolidate?

It may not be obvious that the event failed to be tracked just from whatever error was thrown. However I do agree with changing the level of failedTrackLogMessage to error.

jamesopti: If we consolidated, we could just return early on the missing key and this code would be cleaner/simpler.

circAssimilate: Alright. Gonna resolve this with the latest push but re-open if I'm still missing something.

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -46,11 +46,11 @@ var DECISION_SOURCES = enums.DECISION_SOURCES;
 var DECISION_NOTIFICATION_TYPES = enums.DECISION_NOTIFICATION_TYPES;
 var FEATURE_VARIABLE_TYPES = enums.FEATURE_VARIABLE_TYPES;
 
-describe('lib/optimizely', function () {
+describe('lib/optimizely', function() {
```
circAssimilate: Most of this file's updates are Prettier changes. Here's the only updated part:

https://github.com/optimizely/javascript-sdk/pull/418/files#diff-f6d6b958bf1d05fe2016671164857766R1533-R1552

PR #424, opened 2020-03-09
Converting util/fns from CJS to ES module

fayyazarshad:
## Summary
This activity is part of reducing bundle size, 5to6codemod is used for the automatic transformation but the conversion only include import/export transforms.

## Test plan
All unit test and FSC are getting passed


packages/optimizely-sdk/package.json
```diff
@@ -7,15 +7,16 @@
   "react-native": "lib/index.react_native.js",
   "typings": "lib/index.d.ts",
   "scripts": {
-    "test": "mocha ./lib/*.tests.js ./lib/**/*.tests.js ./lib/**/**/*tests.js --recursive --exit --require lib/tests/exit_on_unhandled_rejection.js",
+    "build": "./node_modules/.bin/babel src --out-dir lib --presets=@babel/env",
```
mjc1283: Why do we need babel? I think we only need a bundler. We can use Rollup, like we do in React SDK.

PR #426, opened 2020-03-12
feat: Key Value cache implementation for React Native

zashraf1985:
## Summary
This is a Key Value Pair cache implementation for react native. This will be used when developing a Separate Datafile Manager for React native which will support datafile caching.

## Test plan
Added Unit Tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,10 +82,11 @@ jobs:
     - &packagetest
       stage: 'Test sub packages'
       node_js: '12'
-      before_install: cd packages/datafile-manager
+      before_install: cd packages/utils
     - <<: *packagetest
       before_install: cd packages/event-processor
     - <<: *packagetest
       before_install: cd packages/logging
     - <<: *packagetest
-      before_install: cd packages/utils
+      before_script: npm install "@react-native-community/async-storage"
+      before_install: cd packages/datafile-manager
--- a/packages/datafile-manager/__mocks__/@react-native-community/async-storage.ts
+++ b/packages/datafile-manager/__mocks__/@react-native-community/async-storage.ts
@@ -25,7 +25,7 @@ export default class AsyncStorage {
           resolve(null)
           break;
         case 'keyWithInvalidJsonObject':
-          resolve('asdfa }')
+          resolve('bad json }')
           break;
       }
     })
--- a/packages/datafile-manager/__mocks__/@react-native-community/async-storage.ts
+++ b/packages/datafile-manager/__mocks__/@react-native-community/async-storage.ts
@@ -18,20 +18,20 @@ export default class AsyncStorage {
   static getItem(key: string, callback?: (error?: Error, result?: string) => void): Promise<string | null> {
     return new Promise((resolve, reject) => {
       switch (key) {
-        case 'keyThatExists': 
+        case 'keyThatExists':
           resolve('{ "name": "Awesome Object" }')
-          break;
+          break
         case 'keyThatDoesNotExist':
           resolve(null)
-          break;
+          break
         case 'keyWithInvalidJsonObject':
           resolve('bad json }')
-          break;
+          break
       }
     })
   }
 
   static setItem(key: string, value: string, callback?: (error?: Error) => void): Promise<void> {
-    return new Promise(resolve => resolve())
+    return Promise.resolve()
   }
 }
--- a/packages/datafile-manager/__mocks__/@react-native-community/async-storage.ts
+++ b/packages/datafile-manager/__mocks__/@react-native-community/async-storage.ts
@@ -1,0 +1,37 @@
+/**
+ * Copyright 2020, Optimizely
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export default class AsyncStorage {
+  static getItem(key: string, callback?: (error?: Error, result?: string) => void): Promise<string | null> {
+    return new Promise((resolve, reject) => {
+      switch (key) {
+        case 'keyThatExists': 
+          resolve('{ "name": "Awesome Object" }')
+          break;
+        case 'keyThatDoesNotExist':
+          resolve(null)
+          break;
+        case 'keyWithInvalidJsonObject':
+          resolve('asdfa }')
+          break;
+      }
+    })
+  }
+
+  static setItem(key: string, value: string, callback?: (error?: Error) => void): Promise<void> {
+    return new Promise(resolve => resolve())
+  }
+}
--- a/packages/datafile-manager/__test__/reactNativeAsyncStorageCache.spec.ts
+++ b/packages/datafile-manager/__test__/reactNativeAsyncStorageCache.spec.ts
@@ -1,0 +1,79 @@
+/**
+ * Copyright 2020, Optimizely
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import ReactNativeAsyncStorageCache from '../src/reactNativeAsyncStorageCache'
+
+describe('reactNativeAsyncStorageCache', () => {
+  let cacheInstance: ReactNativeAsyncStorageCache
+
+  beforeEach(() => {
+    cacheInstance = new ReactNativeAsyncStorageCache()
+  })
+
+  describe('get', function() { 
+    it('should return correct object when item is found in cache', function(done) {
+      cacheInstance.get('keyThatExists')
+        .then(v => { 
+          expect(v).toEqual({ name: "Awesome Object" })
+          done()
+        })
+    })
+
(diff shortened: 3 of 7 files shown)


packages/datafile-manager/src/persistentKeyValueCache.ts
```diff
@@ -0,0 +1,51 @@
...
+  
+  /**
+   * Returns value stored against a key or null if not found.
+   * @param key 
+   * @returns Promise which resolves with a
+   * 1. String if the value found is a string.
+   * 2. Object if value found was stored as a JSON Object.
+   * 3. null if the key does not exist in the cache.
+   */
+  get(key: string): Promise<string | Object | null>
```
mjc1283: Why does it need to support storing both strings and objects?

packages/datafile-manager/src/reactNativeAsyncStorageCache.ts
```diff
@@ -0,0 +1,61 @@
...
+ * limitations under the License.
+ */
+
+import AsyncStorage from '@react-native-community/async-storage';
+import PersistentKeyValueCache from './persistentKeyValueCache';
+
+export default class ReactNativeAsyncStorageCache implements PersistentKeyValueCache {
+  // If value is a valid JSON string, it converts it to object before returning
+  get(key: string): Promise<Object | string | null> {
+    return new Promise((resolve, reject) => {
```
mjc1283: I don't think we need to construct a new promise. We can use the one returned from `AsyncStorage.getItem`.

packages/datafile-manager/src/reactNativeAsyncStorageCache.ts
```diff
@@ -0,0 +1,61 @@
...
+      AsyncStorage.getItem(key).then((val: String | null) => {
+        let finalVal = val;
+        if (val) {
+          try {
+            finalVal = JSON.parse(val.valueOf());
```
mjc1283: Do we need to parse the value here? Datafile manager already does this.

packages/datafile-manager/src/reactNativeAsyncStorageCache.ts
```diff
@@ -0,0 +1,61 @@
...
+          } catch (ex) {
+            // couldnt parse, its a string
```
mjc1283: I am wary of catching and ignoring exceptions. Almost always, we should be logging an error.

packages/datafile-manager/src/reactNativeAsyncStorageCache.ts
```diff
@@ -0,0 +1,61 @@
...
+        }
+        resolve(finalVal);
+      })
+      .catch(reject);
+    });
+  }
+  
+  // If value is an Object, it stringifies it before storing.
+  set(key: string, val: string | Object): Promise<void> {
+    return new Promise((resolve, reject) => {
```
mjc1283: Same comment about constructing a promise

packages/datafile-manager/src/persistentKeyValueCache.ts
zashraf1985: I was trying to make it a bit generic so that someone can use it to store string values too if required. But i can change it to object only for now.

packages/datafile-manager/__mocks__/@react-native-community/async-storage.ts
```diff
@@ -0,0 +1,37 @@
...
+          break;
+        case 'keyWithInvalidJsonObject':
+          resolve('bad json }')
+          break;
+      }
+    })
+  }
+
+  static setItem(key: string, value: string, callback?: (error?: Error) => void): Promise<void> {
+    return new Promise(resolve => resolve())
```
mjc1283:
```suggestion
    return Promise.resolve();
```

packages/datafile-manager/package.json
```diff
@@ -38,6 +38,9 @@
     "@optimizely/js-sdk-logging": "^0.1.0",
     "@optimizely/js-sdk-utils": "^0.1.0"
   },
+  "peerDependencies": {
+    "@react-native-community/async-storage": "^1.8.1"
```
mjc1283: Should we do something very general like `"1.8.x"`? Depending on what features of this we use, can we even get away with `"1.x.x"`?

packages/datafile-manager/src/persistentKeyValueCache.ts
```diff
@@ -0,0 +1,50 @@
...
+export default interface PersistentKeyValueCache {
+  
+  /**
+   * Returns value stored against a key or null if not found.
+   * @param key 
+   * @returns Promise which resolves with a
+   * 1. Object if value found was stored as a JSON Object.
+   * 2. null if the key does not exist in the cache.
+   */
+  get(key: string): Promise<Object | null>
```
mjc1283: It's not recommended to use `Object` as a type in TypeScript (see: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#general-types)

packages/datafile-manager/src/reactNativeAsyncStorageCache.ts
```diff
@@ -0,0 +1,63 @@
...
+            resolve(JSON.parse(val))
+          } catch (ex) {
+            logger.error('Error Parsing Object from cache - %s', ex)
+            reject(ex)
+          }
+        } else {
+          resolve(null)
+        }
+      }).catch(reject)
+    })
```
mjc1283: I still don't think constructing promise is necessary, better to chain off the `getItem` promise. I think this behaves the same:
```suggestion
   return AsyncStorage.getItem(key).then((val: string | null) => {
     if (val) {
       return JSON.parse(val);
     }
     return null;
   }).catch((ex) => {
     logger.error('Error Parsing Object from cache - %s', ex);
     throw ex;
   });
```
Usually it is best to chain off existing promises, instead of creating them.

packages/datafile-manager/__test__/reactNativeAsyncStorageCache.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+describe('reactNativeAsyncStorageCache', () => {
+  let cacheInstance: ReactNativeAsyncStorageCache
+
+  beforeEach(() => {
+    cacheInstance = new ReactNativeAsyncStorageCache()
+  })
+
+  describe('get', function() { 
+    it('should return correct object when item is found in cache', function(done) {
```
mjc1283: Instead of using a `done` argument, I believe you can return a promise from your test method.

packages/datafile-manager/src/reactNativeAsyncStorageCache.ts
```diff
@@ -0,0 +1,63 @@
...
+        AsyncStorage.setItem(key, JSON.stringify(val)).then(resolve).catch(reject)
+      } catch (e) {
+        logger.error('Error stringifying Object to Json - %s', e)
+        reject(e)
+      }
+    })
+  }
+  
+  contains(key: string): Promise<Boolean> {
+    return new Promise((resolve, reject) =>
```
mjc1283: Same comment about constructing a promise.

packages/datafile-manager/src/persistentKeyValueCache.ts
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * An Interface to implement a persistent key value cache which supports strings as keys
+ * and JSON Object as value.
+ */
+export default interface PersistentKeyValueCache {
+  
+  /**
```
mjc1283: For all methods in this interface, the docs need more detail about the returned promise, what value it is fulfilled with, and what causes the promise to reject.

zashraf1985: I just added more details to each method.

packages/datafile-manager/src/reactNativeAsyncStorageCache.ts
```diff
@@ -0,0 +1,52 @@
...
+const logger = getLogger('DatafileManager')
+
+export default class ReactNativeAsyncStorageCache implements PersistentKeyValueCache {
+
+  get(key: string): Promise<any | null> {
+    try {
+      return AsyncStorage.getItem(key)
+        .then((val: string | null) => val ? JSON.parse(val) : null)
+    } catch (ex) {
+      logger.error('Error Parsing Object from cache - %s', ex)
```
mjc1283: We won't reach this line if there's an error thrown by `JSON.parse` on line 29. If you want to log an error upon JSON parse error, the `try`/`catch` block should be located inside the callback passed to `then`:
```js
AsyncStorage.getItem(key).then((val: string | null) => {
  if (!val) {
    return null;
  }
  try {
    return JSON.parse(val);
  } catch (ex) {
    logger.error('Error ...');
    throw ex;
  }
});
```

.travis.yml
```diff
@@ -82,10 +82,11 @@ jobs:
...
       node_js: '12'
-      before_install: cd packages/datafile-manager
+      before_install: cd packages/utils
     - <<: *packagetest
       before_install: cd packages/event-processor
     - <<: *packagetest
       before_install: cd packages/logging
     - <<: *packagetest
-      before_install: cd packages/utils
+      before_script: npm install "@react-native-community/async-storage"
```
mjc1283: Is this necessary? The tests should not actually call into `AsyncStorage` - it should be mocked.

packages/datafile-manager/__test__/reactNativeAsyncStorageCache.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+      return cacheInstance.get('keyThatExists')
+        .then(v => expect(v).toEqual({ name: "Awesome Object" }))
+    })
+
+    it('should return null if item is not found in cache', function() {
+      return cacheInstance.get('keyThatDoesNotExist').then(v => expect(v).toBeNull())
+    })
+
+    it('should reject promise error if string has an incorrect JSON format', function() {
+      return cacheInstance.get('keyWithInvalidJsonObject').catch(() => {})
```
mjc1283: I think this test will pass even if the promise does not reject. We need to do something in the `catch` callback, like set a flag to true, then check it after. Same goes for the other test  that begins with `should reject...`

.travis.yml
zashraf1985: Yes but typescript compiles first and that needs this dependency. It eventually uses mock

packages/datafile-manager/__test__/reactNativeAsyncStorageCache.spec.ts
zashraf1985: Made changes accordingly. Chained a `then` after the `catch` which checks for a value that is only available if it has been chained through `catch`.

PR #427, opened 2020-03-13
refactor: Convert lib/plugins from CJS to ES module

fayyazarshad:
## Summary
This PR converts contents inside `lib/plugins` to ESModule.

## Test plan
All unit test and Full stack compatibility suite tests pass.

### DIFF
--- a/packages/optimizely-sdk/lib/index.browser.tests.js
+++ b/packages/optimizely-sdk/lib/index.browser.tests.js
@@ -55,6 +55,7 @@ describe('javascript-sdk', function() {
         sinon.stub(configValidator, 'validate');
 
         xhr = sinon.useFakeXMLHttpRequest();
+        global.XMLHttpRequest = xhr;
         requests = [];
         xhr.onCreate = function(req) {
           requests.push(req);
--- a/packages/optimizely-sdk/lib/index.browser.tests.js
+++ b/packages/optimizely-sdk/lib/index.browser.tests.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016-2020 Optimizely
+ * Copyright 2016-2020, Optimizely
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -70,8 +70,8 @@ jobs:
       before_script:
         - mkdir $HOME/travisci-tools && pushd $HOME/travisci-tools && git init && git pull https://$CI_USER_TOKEN@github.com/optimizely/travisci-tools.git && popd
       script:
-        - CLIENT=node $HOME/travisci-tools/trigger-script-with-status-update.sh
-        - CLIENT=browser $HOME/travisci-tools/trigger-script-with-status-update.sh
+        - CLIENT=node TESTAPP_BRANCH=fayyaz/added-support-for-es6 $HOME/travisci-tools/trigger-script-with-status-update.sh
```
mikeng13: Is this something that needs to be reverted before merging in? Or is it something we need to clean up once the refactor is done?

zashraf1985: From this repo, it will be reverted before merging. In Testapp, this branch `fayyaz/added-support-for-es6` will be merged because now we will be building the code using rollup before using it in the testapp.

packages/optimizely-sdk/lib/index.browser.tests.js
```diff
@@ -55,6 +55,7 @@ describe('javascript-sdk', function() {
         sinon.stub(configValidator, 'validate');
 
         xhr = sinon.useFakeXMLHttpRequest();
+        global.XMLHttpRequest = xhr;
```
mjc1283, line 58: Hmm, wasn't this removed in an earlier PR? Why add it back?

zashraf1985: This was not causing any problem last time. but in this PR, when we made changes to the event dispatcher. Its somehow failing the tests. I still dont know the root cause but adding it back resolves the issue

PR #428, opened 2020-03-13
feature (optimizelyConfig): added some properties to the optimizely config for the vscode plugin

thomaszurkan-optimizely:
## Summary
- Added project id
- Add events
- Added attributes
These were added to be consumed by the optimizely vscode plugin


mikeng13:
Going to close now until we can pick back up.

PR #430, opened 2020-03-23
feat: React native datafile manager that supports Datafile caching.

zashraf1985:
## Summary
This overrides existing `HttpPollingDatafileManager` to support Datafile Caching for React Native. Browser and Node implementations can also override the same methods to support datafile caching.

## Test plan
Unit Tests Pending

### DIFF
--- a/packages/datafile-manager/__test__/httpPollingDatafileManager.spec.ts
+++ b/packages/datafile-manager/__test__/httpPollingDatafileManager.spec.ts
@@ -83,7 +83,7 @@ describe('httpPollingDatafileManager', () => {
     })
 
     it('resolves onReady immediately', async () => {
-      manager.start()
+      await manager.start()
       await manager.onReady()
       expect(manager.get()).toEqual({ foo: 'abcd' })
     })
@@ -103,7 +103,7 @@ describe('httpPollingDatafileManager', () => {
       )
       const updateFn = jest.fn()
       manager.on('update', updateFn)
-      manager.start()
+      await manager.start()
       expect(manager.responsePromises.length).toBe(1)
       await manager.responsePromises[0]
       expect(updateFn).toBeCalledTimes(1)
@@ -135,7 +135,7 @@ describe('httpPollingDatafileManager', () => {
     })
 
     it('after being started, resolves onReady immediately', async () => {
-      manager.start()
+      await manager.start()
       await manager.onReady()
       expect(manager.get()).toEqual({ foo: 'abcd' })
     })
@@ -148,7 +148,7 @@ describe('httpPollingDatafileManager', () => {
       })
       const updateFn = jest.fn()
       manager.on('update', updateFn)
-      manager.start()
+      await manager.start()
       expect(manager.responsePromises.length).toBe(1)
       await manager.responsePromises[0]
       expect(updateFn).toBeCalledTimes(1)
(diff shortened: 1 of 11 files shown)


packages/datafile-manager/__test__/httpPollingDatafileManager.spec.ts
```diff
@@ -106,10 +100,6 @@ describe('httpPollingDatafileManager', () => {
       manager.start()
       expect(manager.responsePromises.length).toBe(1)
       await manager.responsePromises[0]
-      expect(updateFn).toBeCalledTimes(1)
-      expect(updateFn).toBeCalledWith({
-        datafile: { foo: 'bar' }
-      })
```
mjc1283: Why is this deleted?

mjc1283: Ok, I see why this behavior changed. We are only emitting update events after ready is resolved. Deleting this is OK.

packages/datafile-manager/src/httpPollingDatafileManager.ts
```diff
@@ -314,4 +345,15 @@ export default abstract class HttpPollingDatafileManager implements DatafileMana
       logger.debug('Saved last modified header value from response: %s', this.lastResponseLastModified)
     }
   }
+
+  setDatafileFromCacheIfAvailable(): void {
+    this.cache.get(this.cacheKey)
+      .then(datafile => {
+        if (!this.isReadyPromiseSettled && datafile) {
```
mjc1283: The datafile manager can be stopped at any time. When it's stopped, we want to cancel all pending timers, abort requests, and cut short any already-scheduled promise reactions. So let's add a check for `this.isStarted` in here and return early if it's false (just like other promise reactions in this class).

packages/datafile-manager/__test__/httpPollingDatafileManager.spec.ts
```diff
@@ -634,4 +644,91 @@ describe('httpPollingDatafileManager', () => {
...
+      manager.queuedResponses.push(new Error('Connection Error'))
+      const updateFn = jest.fn()
+      manager.on('update', updateFn)
+      manager.start()
+      await manager.onReady()
+      expect(manager.get()).toEqual({name: 'keyThatExists'})
+      await advanceTimersByTime(50)
+      expect(manager.get()).toEqual({name: 'keyThatExists'})
+      expect(updateFn).toBeCalledTimes(0)
+      updateFn.mockReset()
```
mjc1283: Does calling `mockReset` matter? `updateFn` is just thrown away.

packages/datafile-manager/src/httpPollingDatafileManager.ts
```diff
@@ -197,6 +226,8 @@ export default abstract class HttpPollingDatafileManager implements DatafileMana
     if (datafile !== null) {
       logger.info('Updating datafile from response')
       this.currentDatafile = datafile
+      logger.debug('Adding Datafile to Cache')
```
mjc1283: Let's remove this log message. It will be confusing for users to see this with the no-op cache.

packages/datafile-manager/__test__/httpPollingDatafileManager.spec.ts
zashraf1985: you are right. I naively copied without thinking. The places where i copied it from was verifying it twice hence resetting after first use. Pifalls of copy/pasting :)

PR #438, opened 2020-04-02
[OASIS-6102]: changed functionality of the JSON schema validator module

yavorona:
## Summary
- Changed functionality of the JSON schema validator module to only validate against projectConfigSchema
- Updated existing unit tests to reflect this change

## Test plan

Existing unit tests
## Issues

OASIS-6102

### DIFF
--- a/packages/optimizely-sdk/lib/core/project_config/index.js
+++ b/packages/optimizely-sdk/lib/core/project_config/index.js
@@ -17,7 +17,6 @@ var fns = require('../../utils/fns');
 var enums = require('../../utils/enums');
 var jsSdkUtils = require('@optimizely/js-sdk-utils');
 var configValidator = require('../../utils/config_validator');
-var projectConfigSchema = require('./project_config_schema');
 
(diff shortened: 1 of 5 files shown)


mikeng13:
@yavorona we should also include a section on the readme to explain when/how to use the jsonSchemaValidator --> users should include in their `config` only if they are not using our project config manager and are using a datafile that has been modified.

yavorona:
Talked offline with @mikeng13 and decided to document this in our developer docs instead.

mjc1283:
@mikeng13 This PR shouldn't have any impact on users, it's just a refactor that removes the project config schema from being pulled into the browser bundle. Let me know if I overlooked something that needs to be documented or updated in docs.

mikeng13:
> @mikeng13 This PR shouldn't have any impact on users, it's just a refactor that removes the project config schema from being pulled into the browser bundle. Let me know if I overlooked something that needs to be documented or updated in docs.

You are right. Looks like we already don't bring in the validator in the browser bundle and instead are removing the schema from the project config. SGTM

PR #445, opened 2020-04-07
refactor: Convert entry points to es module and create minified bundles

fayyazarshad:
## Summary
This is the first step in an effort to convert javascript SDK from Common JS to ES Module. The changes include
1. Convert all the entry point files to ES Module.
2. Minified Common JS bundles are being created for each entry point using Rollup and Terser
`main: dist/optimizely.node.min.js`
`browser: dist/optimizely.browser.min.js`
`react-native: dist/optimizely.react_native.min.js`
3. All the bundles including umd bundle are generated using Rollup and terser. Webpack is only being used to keep the Karma tests running.

### Why Webpack for Karma?
We tried using Rollup with Karma for the same purpose but couldnt run it despite trying many times. To not block progress on the ES module conversion, we recommend moving forward with using webpack to keep running karma tests when all the bundles are being created by Rollup and terser. We can revisit and modify Karma to use Rollup later. 

### Note about Testapp
This PR uses a specific branch of javascript-testapp. This change is required because now testapp has to run an additional command to build the bundles which was not the case earlier. This testapp PR needs to be merged before merging this SDK branch.
https://github.com/optimizely/javascript-testapp/pull/83

## Test plan
All unit tests and Full stack compatibility suite tests pass.


packages/optimizely-sdk/karma.base.conf.js
```diff
@@ -24,6 +24,14 @@ module.exports = {
 
   webpack: {
     mode: 'production',
+    module: {
+      rules: [{
+        exclude: /\.json$/,
+        use: {
+          loader: 'buble-loader',
```
mjc1283, line 31: What is buble-loader used for?

packages/optimizely-sdk/lib/index.browser.tests.js
```diff
@@ -57,7 +55,6 @@ describe('javascript-sdk', function() {
         sinon.stub(configValidator, 'validate');
 
         xhr = sinon.useFakeXMLHttpRequest();
-        global.XMLHttpRequest = xhr;
```
mjc1283, line 60: Is this line not necessary?

packages/optimizely-sdk/rollup.config.js
```diff
@@ -0,0 +1,83 @@
...
+      plugins: [ terser() ],
+    },
+  ],
+};
+
+export default [
+  BUILD_ALL && getConfigForPlatform('node'),
+  BUILD_ALL && getConfigForPlatform('browser'),
+  BUILD_ALL && getConfigForPlatform('react_native'),
+  (BUILD_ALL || BUILD_UMD_BUNDLE) && configForUMD,
```
mjc1283, line 82: Can you add another config for an ES-module browser output? The `input` should be `lib/index.browser.js`, but the `output.format` should be `es` and the `output.file` should be `dist/optimizely.browser.es.min.js`. Then in `package.json`, we should add `"module": "dist/optimizely.browser.es.min.js"`.

This can be implemented in a follow up PR if you want.

packages/optimizely-sdk/package.json
```diff
@@ -42,27 +43,31 @@
...
     "chai": "^4.2.0",
     "coveralls": "^3.0.2",
     "eslint": "^6.7.2",
+    "esm": "^3.2.25",
     "istanbul": "^0.4.5",
     "json-loader": "^0.5.4",
     "karma": "^4.4.1",
     "karma-browserstack-launcher": "^1.5.1",
     "karma-chai": "^0.1.0",
-    "karma-chrome-launcher": "^2.1.1",
```
mjc1283, line 54: Why is this removed? Is this used for running umd or xbrowser tests locally?

packages/optimizely-sdk/package.json
```diff
@@ -42,27 +43,31 @@
...
     "karma-mocha": "^1.3.0",
-    "karma-sinon": "^1.0.5",
```
mjc1283, line 56: Why is this removed?

packages/optimizely-sdk/package.json
zashraf1985: I tried removing modules that are not being used. This is in the devDependencies but is not being used in karma file. It looks like somebody added it to try karma tests locally.

packages/optimizely-sdk/package.json
zashraf1985: I tried removing modules which were not being used by removing and running tests. Removing this one did not fail any tests so looks like its no longer required

packages/optimizely-sdk/lib/index.browser.tests.js
zashraf1985: It looked redundant and removing it did not affect any tests.

packages/optimizely-sdk/karma.base.conf.js
zashraf1985: Used it for transpilation. I was confused and thought we need to transpile the code so that relatively new JS features also work. But we are not traspiling the bundles so its not required here. I will remove it.

packages/optimizely-sdk/rollup.config.js
zashraf1985: Done

packages/optimizely-sdk/package.json
mjc1283: If this is necessary to run it locally, let's leave it in, otherwise OK.

packages/optimizely-sdk/lib/index.browser.js
```diff
@@ -13,146 +13,166 @@
...
+  setLogLevel,
+  createInstance,
+  __internalResetRetryState,  
+}
+
+export default {
+  logging: loggerPlugin,
+  errorHandler: defaultErrorHandler,
+  eventDispatcher: defaultEventDispatcher,
+  enums,
```
mjc1283, line 173: Oops, I missed this. @zeeshan don't use the property shorthand here. This is not ES5-compatible. Needs to be `enums: enums`.

zashraf1985: This is Done

PR #458, opened 2020-04-17
fix: Fixed Project config update logic

zashraf1985:
## Summary
Project config manager creates project config from datafile everytime it recieves an update from datafile manager. Generally, the redundant processing is blocked at the datafile manager level where it ignores 304 response. In some cases http library handles that interenally and returns 200 ok response with cached datafile body. In this case project config manger creates a new project config object everytime from datafile and then throws it away before using because of the revision check. I have now moved this config creation logic after the revision check. This way, config will only be created if the datafile that is received has a new revision number.

## Test plan
All Existing tests pass


packages/optimizely-sdk/lib/core/project_config/project_config_manager.js
```diff
@@ -251,14 +220,33 @@ ProjectConfigManager.prototype.__validateDatafileOptions = function(datafileOpti
...
-ProjectConfigManager.prototype.__handleNewConfigObj = function(newConfigObj) {
-  var oldConfigObj = this.__configObj;
+ProjectConfigManager.prototype.__handleNewDatafile = function() {
+  var newDatafile = this.datafileManager.get();
 
-  var oldRevision = oldConfigObj ? oldConfigObj.revision : 'null';
-  if (oldRevision === newConfigObj.revision) {
-    return;
+  // Only proceed if the datafile has a different revision
+  if (!newDatafile || (this.__configObj && newDatafile.revision === this.__configObj.revision)) {
```
mjc1283, line 233: If `newDatafile` is falsy, we should not return `{ success: true }`.  Datafile manager shouldn't return a falsy value in these cases, but if it does, we shouldn't be treating it like a success.

I am not sure I like the combination of the functions for handling the update and the fulfillment of datafile manager `onReady`. They are very close, but the return value doesn't matter in one case, so it's a little confusing and a subtle error like this can happen. Putting this revision check in both places might be a little verbose/duplicative, but also might be easier to think about and prevents doing unnecessary work of creating a return object (with an error message) just for it to be ignored.

If we are going to do this consolidation, we should get rid of the intermediate `__onDatafileManagerUpdate` and `__onDatafileManagerReadyFulfill ` functions, and just have the caller call `_handleNewDatafile` directly.

PR #462, opened 2020-04-21
fix(project config): Don't mutate the datafile object

mjc1283:
## Summary
`createProjectConfig` was destructively mutating its `datafile` argument. After that mutation, the datafile object could not be re-used to create another instance.
With this change, `createProjectConfig` copies the objects it needs to mutate and mutates only those copies. It's not a full deep clone; it only shallowly clones audiences, experiments, feature flags, groups, rollouts, group experiments, and rollout experiments.

## Test plan

- Updated & new unit tests
- Compatibility suite should pass

## Issues
https://optimizely.atlassian.net/browse/OASIS-6301

### DIFF
--- a/packages/optimizely-sdk/lib/core/project_config/index.js
+++ b/packages/optimizely-sdk/lib/core/project_config/index.js
@@ -33,6 +33,12 @@ function createMutationSafeDatafileCopy(datafile) {
   datafileCopy.audiences = (datafile.audiences || []).map(function(audience) {
     return fns.assign({}, audience);
   });
+  datafileCopy.experiments = (datafile.experiments || []).map(function(experiment) {
+    return fns.assign({}, experiment);
+  });
+  datafileCopy.featureFlags = (datafile.featureFlags || []).map(function(featureFlag) {
+    return fns.assign({}, featureFlag);
+  });
   datafileCopy.groups = (datafile.groups || []).map(function(group) {
     var groupCopy = fns.assign({}, group);
     groupCopy.experiments = (group.experiments || []).map(function(experiment) {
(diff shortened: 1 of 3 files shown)


mjc1283:
@mikeng13 Yea, that'd probably work. But we know what the datafile schema is, we don't need to handle arbitrary objects. The datafile schema does not change very often, and when it does, we don't necessarily need to change this code. Overall I think the project config module could be a lot cleaner, but it would require careful refactoring of many other modules, so not trying to tackle it now. I think we can set it up so nothing is mutated, and we create fast lookup objects/Maps separately from the datafile.

mjc1283:
Put another way, copying/cloning is only required because we are mutating the datafile object. But there is no inherent need to mutate the datafile object. I think we can refactor so that it's nothing is mutated, and nothing is copied.

mikeng13:
Hmm.. I think even if we don't mutate the datafile, customers could end up inadvertently mutating it and causing issues. I think cloning it for safe use makes sense.

mjc1283:
That's a fair point (cloning for safety). I didn't look too closely at the Medium post. Earlier I had my eye on this npm library [`rfdc`](https://www.npmjs.com/package/rfdc). I think we could include a forked snippet from that (don't need all its features) and would only add about 50-100 LOC.

jamesopti:
> That's a fair point (cloning for safety). I didn't look too closely at the Medium post. Earlier I had my eye on this npm library [`rfdc`](https://www.npmjs.com/package/rfdc). I think we could include a forked snippet from that (don't need all its features) and would only add about 50-100 LOC.

This would be my preference in the medium term, but the current fix is OK by me for now!
Could that utility also replace our [`Object.assign` polyfill](https://github.com/optimizely/javascript-sdk/blob/master/packages/optimizely-sdk/lib/utils/fns/index.js#L21-L42)?

PR #463, opened 2020-04-22
feat: Added proxy support for node

zashraf1985:
## Summary
Added support to provide proxy information. Users can now provide a `proxyUrl` in the optimizely config options.

```
const instance = optimizely.createInstance({
  sdkKey: '{sdkKey}',
  proxyUrl: 'http://{proxyhost}:{port}',
});
```
url can also contain basic authentication
`http://{username}:{password}@{proxyhost}:{port}`

## Test plan
1. All existing tests pass.
2. New Unit tests pending


mikeng13:
@mjc1283 this can be separately, but what do you think about having a compat suite test for this? Seems like at some point we wanna support in all SDKs anyway. Agent already supports this and can be tested via FSC.

packages/datafile-manager/__test__/httpPollingDatafileManager.spec.ts
```diff
@@ -310,7 +310,7 @@ describe('httpPollingDatafileManager', () => {
             await advanceTimersByTime(1000);
             expect(makeGetRequestSpy).toBeCalledTimes(1);
 
-            resolveResponsePromise!({
+            await resolveResponsePromise!({
```
mjc1283, line 313: Why are these `await`s` added? Was something wrong with these tests?

packages/datafile-manager/package.json
```diff
@@ -38,13 +38,15 @@
     "nock": "^10.0.6",
     "prettier": "^1.19.1",
     "ts-jest": "^24.0.0",
-    "typescript": "3.3.3333",
+    "typescript": "3.8.3",
```
mjc1283, line 41: Why upgrade the TypeScript version? If we need to do this, can we do it in a separate PR?

packages/datafile-manager/src/nodeRequest.ts
```diff
@@ -115,15 +119,22 @@ function getResponseFromRequest(request: http.ClientRequest): Promise<Response>
   });
 }
 
-export function makeGetRequest(reqUrl: string, headers: Headers): AbortableRequest {
+export function makeGetRequest(reqUrl: string, headers: Headers, proxyUrl?: string): AbortableRequest {
```
mjc1283, line 122: To avoid creating the agent on every request, I think `NodeDatafileManager` should have a private `agent` field that's initialized in the constructor. Then `makeGetRequest` here accepts an optional agent.

packages/datafile-manager/src/nodeRequest.ts
```diff
@@ -115,15 +119,22 @@ function getResponseFromRequest(request: http.ClientRequest): Promise<Response>
...
   // TODO: Use non-legacy URL parsing when we drop support for Node 6
   const parsedUrl = url.parse(reqUrl);
+  let proxyAgent = null;
 
   let requester: ClientRequestCreator;
   if (parsedUrl.protocol === 'http:') {
     requester = http.request;
+    if (proxyUrl) {
+      proxyAgent = new (HttpProxyAgent as any)(proxyUrl);
```
mjc1283, line 131: I think you did this just as a first pass to get it working :) but we can't cast it to `any`. We should install the types for the new libraries from DefinitelyTyped, or write them ourselves if none exist.

mjc1283:
@mikeng13 Agree, FSC coverage for this seems valuable.

mjc1283:
@zashraf1985 Could you also update the credits section in the root README as well as the README inside optimizely-sdk. We need to add all new 3rd-party dependencies.

JordanAdams:
Hey @zashraf1985, what's your feeling with picking up these changes? We have some work depending on this so I'd be happy to help out if need be 🙂

PR #467, opened 2020-04-30
feat: Added a new getFeatureVariableJson Api

zashraf1985:
## Summary
Implemented getFeatureVariableJson API

## Test plan
Added unit tests for getFeatureVariableJson.

### DIFF
--- a/packages/optimizely-sdk/lib/core/optimizely_config/index.tests.js
+++ b/packages/optimizely-sdk/lib/core/optimizely_config/index.tests.js
@@ -81,6 +81,9 @@ describe('lib/core/optimizely_config', function() {
         });
         var variablesMap = featuresMap[featureFlag.key].variablesMap;
         featureFlag.variables.forEach(function(variable) {
+          if (variable.type === 'string' && variable.subType === 'json') {
+            variable.type = 'json';
+          }
           assert.include(variablesMap[variable.key], {
             id: variable.id,
             key: variable.key,
--- a/packages/optimizely-sdk/lib/core/optimizely_config/index.tests.js
+++ b/packages/optimizely-sdk/lib/core/optimizely_config/index.tests.js
@@ -81,13 +81,13 @@ describe('lib/core/optimizely_config', function() {
         });
         var variablesMap = featuresMap[featureFlag.key].variablesMap;
         featureFlag.variables.forEach(function(variable) {
-          if (variable.type === 'string' && variable.subType === 'json') {
-            variable.type = 'json';
-          }
+          // json is represented as sub type of string to support backwards compatibility in datafile.
+          // project config treats it as a first-class type.
+          var expectedVariableType = (variable.type === "string" && variable.subType === "json") ? "json" : variable.type;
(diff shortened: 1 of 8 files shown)


packages/optimizely-sdk/lib/core/optimizely_config/index.tests.js
```diff
@@ -81,6 +81,9 @@ describe('lib/core/optimizely_config', function() {
         });
         var variablesMap = featuresMap[featureFlag.key].variablesMap;
         featureFlag.variables.forEach(function(variable) {
+          if (variable.type === 'string' && variable.subType === 'json') {
+            variable.type = 'json';
+          }
```
mjc1283, line 86: Can we add a comment explaining why this is overwritten? Also I think it's a bad idea to mutate `datafile`; it is assigned once at the top of the file and could be used in other tests.

Instead could we do something like this? 
```js
var expectedVariableType =
  variable.type === "string" && variable.subType === "json"
    ? "json"
    : variable.type;
assert.include(variablesMap[variable.key], {
  id: variable.id,
  key: variable.key,
  type: expectedVariableType,
  value: variable.defaultValue,
});
```

packages/optimizely-sdk/lib/core/project_config/index.js
```diff
@@ -119,6 +119,13 @@ export var createProjectConfig = function(datafile) {
 
   projectConfig.featureKeyMap = fns.keyBy(projectConfig.featureFlags || [], 'key');
   objectValues(projectConfig.featureKeyMap || {}).forEach(function(feature) {
+    // Convert type:string and subType:json to type:json to generalize implementation
```
mjc1283, line 122: I think this comment needs a little more information. We can mention that json variables in the datafile have a `type` of `"string"` for backwards compatibility, but we want to erase the concept of `subType`, and represent them as a first-class type within the project config.

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -3000,6 +3000,36 @@ describe('lib/optimizely', function() {
                   });
                 });
 
+                it('returns the right value from getFeatureVariable and send notification with featureEnabled true', function() {
```
mjc1283, line 3003: These strings passed to `it()` should mention the type of variable being tested (I know you are following what was already here). It's weird that they all have the same description string.

packages/optimizely-sdk/lib/core/project_config/index.js
zashraf1985: Done

packages/optimizely-sdk/lib/optimizely/index.tests.js
zashraf1985: Done

packages/optimizely-sdk/lib/core/optimizely_config/index.tests.js
zashraf1985: Done

PR #470, opened 2020-05-01
feat: Added a new getAllFeatureVariables Api

zashraf1985:
## Summary
Implemented getAllFeatureVariables API

## Test plan
Added new unit tests.

### DIFF
--- a/packages/optimizely-sdk/lib/index.d.ts
+++ b/packages/optimizely-sdk/lib/index.d.ts
@@ -94,12 +94,12 @@ declare module '@optimizely/optimizely-sdk' {
       variableKey: string,
       userId: string,
       attributes?: UserAttributes
-    ): any | null;
+    ): unknown;
     getAllFeatureVariables(
       featureKey: string,
       userId: string,
       attributes?: UserAttributes
-    ): any | null;
+    ): unknown;
     getOptimizelyConfig(): OptimizelyConfig | null;
     onReady(options?: { timeout?: number }): Promise<{ success: boolean; reason?: string }>;
     close(): Promise<{ success: boolean; reason?: string }>;
--- a/packages/optimizely-sdk/lib/index.d.ts
+++ b/packages/optimizely-sdk/lib/index.d.ts
@@ -89,6 +89,17 @@ declare module '@optimizely/optimizely-sdk' {
       userId: string,
       attributes?: UserAttributes
     ): string | null;
+    getFeatureVariableJson(
+      featureKey: string,
+      variableKey: string,
+      userId: string,
+      attributes?: UserAttributes
+    ): any;
+    getAllFeatureVariables(
+      featureKey: string,
+      userId: string,
+      attributes?: UserAttributes
+    ): any;
     getOptimizelyConfig(): OptimizelyConfig | null;
     onReady(options?: { timeout?: number }): Promise<{ success: boolean; reason?: string }>;
     close(): Promise<{ success: boolean; reason?: string }>;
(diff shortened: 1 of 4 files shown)


packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -957,6 +957,127 @@ Optimizely.prototype.getFeatureVariableJson = function(featureKey, variableKey,
   }
 };
 
+/**
+ * Returns values for all the json variables attached to the given feature
```
mjc1283, line 961: This returns all variable values, not just values for JSON variables, right?

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -957,6 +957,127 @@ Optimizely.prototype.getFeatureVariableJson = function(featureKey, variableKey,
...
+    var decision = this.decisionService.getVariationForFeature(configObj, featureFlag, userId, attributes);    
+    var featureEnabled = decision.variation !== null ? decision.variation.featureEnabled : false;
+    var logger = this.logger;
+    var allVariables = {};
+    
+    featureFlag.variables.forEach(function (variable) {
+      var variableValue = variable.defaultValue;      
+      if (decision.variation !== null) {
+        var value = projectConfig.getVariableValueForVariation(configObj, variable, decision.variation, logger);
+        if (value !== null) {
```
mjc1283, line 1000: Seems like you are following `_getFeatureVariableForType`, can we refactor & call that from here instead of duplicating this logic? 

Aside, reading this now I find it a little hard to follow with the nested `if`s. Can we flatten this out while preserving the same behavior? `decision.variation !== null`, `value !== null`, `featureEnabled` determine the log message. `decision.variation !== null` and `featureEnabled` determine the variable value.

zashraf1985: Thanks for the suggestion. since i dont understand all the behaviour in detail, i followed whatever was already there. The tests already written are extremely thorough. I will go ahead and simplify the stuff and rely on the tests to ensure correctness

packages/optimizely-sdk/lib/optimizely/index.js
zashraf1985: You are right. It was a copy/paste mistake.

packages/optimizely-sdk/lib/optimizely/index.js
zashraf1985: @mjc1283 I refactored this.
1. Moved common code in a separate function
2. Separated logs logic and variable assignment.
3. Simplified checks for logs to make them more readable and added comments

packages/optimizely-sdk/lib/index.d.ts
```diff
@@ -89,6 +89,17 @@ declare module '@optimizely/optimizely-sdk' {
       userId: string,
       attributes?: UserAttributes
     ): string | null;
+    getFeatureVariableJson(
```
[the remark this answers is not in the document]
zashraf1985: Done

packages/optimizely-sdk/lib/optimizely/index.js
```diff
@@ -744,33 +744,62 @@ Optimizely.prototype._getFeatureVariableForType = function(featureKey, variableK
...
+      sourceInfo: sourceInfo,
+    },
+  });
+  return variableValue;
+};
 
-  if (decision.variation !== null) {
-    featureEnabled = decision.variation.featureEnabled;
-    var value = projectConfig.getVariableValueForVariation(configObj, variable, decision.variation, this.logger);
+Optimizely.prototype._getFeatureVariableValueFromVariation = function(featureKey, featureEnabled, variation, variable, userId) {
```
mjc1283, line 784: Can you add a doc comment for `_getFeatureVariableValueFromVariation`?

zashraf1985: Done

PR #471, opened 2020-05-06
Configure repository.directory for monorepo packages

jasonkarns:
## Summary 

Configure each package's `package.json#repository.directory` 
property to properly reflect the location of the package within 
the repository.

This is the standard configuration value for deep-linking into a
package's directory within a monorepo.

npm RFC: https://github.com/npm/rfcs/blob/latest/implemented/0010-monorepo-subdirectory-declaration.md

This will ensure the packages' pages on npmjs.org will link to the
correct directory.

Also, _eventually_, it will ensure the `npm repo` command opens to the
correct location: https://github.com/npm/cli/pull/163

Lastly, this is _required_ for monorepo packages to be successfully
published to the GitHub Package Registry (if that is ever considered in
the future).
https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-npm-for-use-with-github-packages#publishing-multiple-packages-to-the-same-repository

## Test plan
n/a

### DIFF
--- a/packages/datafile-manager/package.json
+++ b/packages/datafile-manager/package.json
@@ -2,6 +2,11 @@
   "name": "@optimizely/js-sdk-datafile-manager",
   "version": "0.5.0",
   "description": "Optimizely Full Stack Datafile Manager",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/optimizely/javascript-sdk.git",
+    "directory": "packages/datafile-manager"
+  },
   "license": "Apache-2.0",
   "engines": {
     "node": ">=6.0.0"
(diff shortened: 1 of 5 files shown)


jasonkarns:
@mjc1283 same! I only learned about it when I tried to publish our fork to github package registry. Since the package registry ties packages to a repo, having multiple packages published to a _single_ repo is problematic. So they _require_ this configuration to make it work (somehow). All the rest I only stumbled on afterwards!

(And it's a bummer that the npm CLI is still pending the PR to respect it. I use the `npm repo` command heavily and it's annoying that it doesn't yet work well with monorepos. 😞  )

PR #472, opened 2020-05-06
Updates engines.node stanza to reflect actual requirements

jasonkarns:
## Summary
The version spec in package.json#engines.node is out of date from what is specified in the corresponding READMEs. This updates the package to declare the correct minimum version of node.

While it may seem unimportant, the package.json#engines.node spec is respected by some tools, like nodenv, for node version selection. So this being incorrect can potentially activate the wrong version of node for someone developing locally.

### DIFF
--- a/packages/datafile-manager/package.json
+++ b/packages/datafile-manager/package.json
@@ -4,7 +4,7 @@
   "description": "Optimizely Full Stack Datafile Manager",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=6.0.0"
+    "node": ">=8.0.0"
   },
   "main": "lib/index.node.js",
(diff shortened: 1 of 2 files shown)


mjc1283:
@jasonkarns Could you merge the latest master from this repo into your branch? Then I can go ahead and merge this.

jasonkarns:
> Could you merge the latest master from this repo into your branch? Then I can go ahead and merge this.

I merged using the update-branch... but might be worth re-evaluating whatever process requires this... shouldn't be any reason for that to be necessary unless a branch is significantly out of date. This branch was cut from latest master as of yesterday, so it wasn't significantly stale.

PR #474, opened 2020-05-07
Removes direct UUID dep

### LINKED ISSUE #479: Update UUID dependency for ESM
## How would the enhancement work?

As of v8, uuid exposes ESM to support tree shaking.
https://github.com/uuidjs/uuid/blob/master/CHANGELOG.md#800-2020-04-29

Now we can import the v4 algorithm directly without pulling the entire library into our bundle.

#474 was initially opened to bump the UUID dep in optimizely-sdk. However, it was later discovered that the UUID dep is also imported by js-sdk-utils. To properly update the dep and to avoid duplicating the dep:

1. [ ] optimizely-sdk should import the UUID function from js-sdk-utils
2. [ ] optimizely-sdk should remove the UUID dependency from its package.json
3. [ ] js-sdk-utils should update its UUID dep to latest (v8.0) and import via ESM

## When would the enhancement be useful?

Minimize optimizely bundles by excluding portions of the UUID library that are unnecessary.

potential **before**

potential **after**

jasonkarns:
## Summary

UUID is already exposed in the js-sdk-utils package, so we should use it directly.
That will ensure de-duping, as well as automatic tree shaking of the uuid dep (because the utils package imports UUID via ESM).

**Before**
image

**After**
image

closes #479

### DIFF
--- a/packages/optimizely-sdk/lib/utils/fns/index.js
+++ b/packages/optimizely-sdk/lib/utils/fns/index.js
@@ -17,7 +17,7 @@ import { generateUUID as uuid, keyBy as keyByUtil } from '@optimizely/js-sdk-uti
 
 var MAX_SAFE_INTEGER_LIMIT = Math.pow(2, 53);
 
-export var assign = function(target) {
+export var assign = function (target) {
   if (!target) {
     return {};
   }
@@ -40,24 +40,24 @@ export var assign = function(target) {
   }
 };
 
-export var currentTimestamp = function() {
+export var currentTimestamp = function () {
   return Math.round(new Date().getTime());
 };
 
-export var isSafeInteger = function(number) {
+export var isSafeInteger = function (number) {
   return typeof number == 'number' && Math.abs(number) <= MAX_SAFE_INTEGER_LIMIT;
 };
 
-export var keyBy = function(arr, key) {
+export var keyBy = function (arr, key) {
   if (!arr) return {};
-  return keyByUtil(arr, function(item) {
+  return keyByUtil(arr, function (item) {
     return item[key];
   });
 };
 
-export { uuid }
+export { uuid };
 
-export var isNumber = function(value) {
+export var isNumber = function (value) {
   return typeof value === 'number';
 };
 
--- a/packages/optimizely-sdk/lib/utils/fns/index.js
+++ b/packages/optimizely-sdk/lib/utils/fns/index.js
@@ -62,10 +62,10 @@ export var isNumber = function (value) {
 };
 
 export default {
-  assign,
-  currentTimestamp,
-  isSafeInteger,
-  keyBy,
-  uuid,
-  isNumber,
+  assign: assign,
+  currentTimestamp: currentTimestamp,
+  isSafeInteger: isSafeInteger,
+  keyBy: keyBy,
+  uuid: uuid,
+  isNumber: isNumber,
 };
--- a/packages/optimizely-sdk/lib/utils/fns/index.js
+++ b/packages/optimizely-sdk/lib/utils/fns/index.js
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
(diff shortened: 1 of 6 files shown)


jasonkarns:
Just realized the js-sdk-utils package exposes this same function... So why is it imported directly in optimizely-sdk? The versions now need bumped in both packages so that they'll de-dupe. But the bigger question is: if this is exposed in the utils package, why isn't the utils version being used instead of importing a second copy?

mjc1283:
Yes, good points. I think we need to remove this dependency from optimizely-sdk, use the version in utils, and update the version in utils to 8.0

jasonkarns:
PR now updated to remove the direct UUID dep. It's pulled in from js-sdk-utils. Updated tests to stub the correct function.

Also, the js-sdk-utils package is already using the ESM import of UUID, so no version bump is necessary. Indeed, keeping the version at v3* is probably preferable for a while because it increases the chances that downstream consumers can de-dupe UUID itself in _their_ bundles.

packages/optimizely-sdk/lib/utils/fns/index.js
```diff
@@ -41,34 +40,32 @@ export var assign = function(target) {
...
-  isSafeInteger: isSafeInteger,
-  keyBy: keyBy,
-  uuid: uuid,
-  isNumber: isNumber,
+  assign,
+  currentTimestamp,
+  isSafeInteger,
+  keyBy,
+  uuid,
+  isNumber,
```
mjc1283, line 70: Could you revert this part? We have actually held back on introducing more new language features (apart from modules). We have had some discussions about addressing this wholesale in the near future, potentially via the introduction of TypeScript or Babel in optimizely-sdk.

jasonkarns: done.

Does this SDK officially support IE? That's the only browser that doesn't support this feature. (node has support since 4.0)

mjc1283: If the user includes a `Promise` polyfill, the SDK will work in IE.

jasonkarns: ah, good to know

jasonkarns:
As a side-note, something I encountered while updating the tests... because all the tests across this suite use the same fake UUID for stubbing the uuid function, when a sinon stub was not properly unstubbed, a stub from one test would poison other subsequent tests. I would strongly suggest using different dummy values between files/suites. That way when a test bleeds over to affect another test, the root cause can be identified much quicker (because the error message will be accurate and not masked by the fact that the UUID is "correct" but for the wrong reason).

Alternatively, since UUID generation is not an expensive operation, I would suggest that the stubbing should be updated to return a different value for each test. That is, remove the static stubbing entirely. Generate a UUID during setup and use _that_ uuid as the stubbed value for the test. That way, a poisoned test run will more quickly expose the root cause; because the UUIDs won't be shared across tests.

mjc1283:
Everything LGTM, but I overlooked a process issue in my first review. Our packages are published independently, so we can't update `optimizely-sdk` to use the new stuff in `utils` until a new version of `utils` is published. If you could split up the `utils` changes into its own PR, I can merge that and get a new version published. Sorry for the extra overhead. The fixes here are much appreciated.

jasonkarns:
It probably got lost in the evolution of this PR, but back [here](https://github.com/optimizely/javascript-sdk/pull/474#issuecomment-625834700) I discovered that the version of UUID being used by the utils package is already ES compliant and tree-shakeable. This PR _does_ contain a tweak to the utils package, but its purely stylistic. The export of the generateUUID function can be exported directly without being wrapped in another function: https://github.com/optimizely/javascript-sdk/pull/474/files#diff-234d37a781c939ff5538faf0110c211c

However, that change does not need to be released for the optimizely-sdk package to consume the function from utils instead of UUID.

In other words, this PR is entirely focused on optimizely-sdk and can be merged as-is. The change within the utils package can be merged but can remain an unreleased tweak on master.

PR #475, opened 2020-05-07
Fix broken test globs and coverage scripts

jasonkarns:
## Summary

**globs**
It is not safe to use unquoted globstars within npm scripts because they
are not expanded consistently on all versions of bash. Specifically,
globstar isn't even available until version 4 of bash; and macOS still
ships with v3. (Ergo, globstar won't work on stock macOS)
Further, even if it _is_ available, it's only an option. Nothing ensures
the option is enabled.

Clearly, some prior developer recognized that the tests weren't all
running and they "duplicated" the glob pattern to ensure it matched all
files. However, they did 3 things wrong:

1. mocha (like virtually all *nix utilities) expects file arguments
   _last_
2. they still didn't properly quote the globs, merely worked around the
   globstar issue by manually matching 3 directory levels
3. the workaround still doesn't robustly match recursive directories
   because it only goes three levels deep

The proper fix is to quote the glob pattern such that mocha itself (by
means of a dependency) will expand the globstar pattern consistently
regardless of shell version.

**istanbul**
Istanbul coverage is broken (https://travis-ci.org/github/optimizely/javascript-sdk/jobs/683996183#L1612) but istanbul is also deprecated.

This replaces istanbul with nyc and cleans up the coverage scripts and flow. The coverage is _still_ broken (needs some ES or TS plugins for nyc) but it's an improvement.

**npm scripts** 
Also sorts the npm scripts a bit more logically and auto-runs lint after npm-test.


jasonkarns:
I also considered updating mocha's version in this PR because it's 2 major versions behind latest. But I decided against it. However, the test suite did run SUCCESSFULLY 🎉  against the latest version (v7.1.2) without any further modifications! So presumably it would be safe to bump.

packages/optimizely-sdk/package.json
```diff
@@ -8,15 +8,18 @@
   "react-native": "dist/optimizely.react_native.min.js",
   "typings": "lib/index.d.ts",
   "scripts": {
-    "test": "mocha ./lib/*.tests.js ./lib/**/*.tests.js ./lib/**/**/*tests.js --recursive --exit --require esm --require lib/tests/exit_on_unhandled_rejection.js",
+    "lint": "eslint 'lib/**/*.js'",
+    "test": "nyc -s mocha --exit -r esm -r lib/tests/exit_on_unhandled_rejection.js 'lib/**/*.tests.js'",
```
mjc1283, line 12: @jasonkarns I'm not too familiar with `nyc`, but what is the motivation for running it here within the 'test' package.json script? It seems more natural to only run it as part of `npm run cover`, and perhaps do this within the `precover` script you added. `nyc` seems to add a little extra run time here. If I don't care about coverage and just want to run tests, this is forcing me to wait that little extra period.

jasonkarns: There isn't a strong motivation other than simplification. `nyc`'s documentation and examples demonstrate both uses. The benefit of instrumenting all test runs, is that when a coverage report is desired, the test suite doesn't need run _again_. (The scripts prior to this PR cause the test suite to be run twice on CI.)

However, I acknowledge that running the suite twice on CI is not a problem at the cost of slowing down _local_ test runs. I'm totally cool splitting those up.

jasonkarns:
Updated the scripts to not run coverage as part of npm-test.

To summarize, the istanbul coverage was broken prior to this PR. And _after_ this PR, coverage is still broken, but is configured to use the istanbul cli replacement: nyc.

nyc is reporting 0 coverage after this PR (which is an improvement from before this PR, where the coverage script actually errored).  I haven't figured out why nyc is reporting 0 coverage, but I suspect it has something to do with the mocha --exit flag or requiring esm. (Or perhaps even the unhandled rejection hook.)

Perhaps the fix for #487 will fix the coverage. Regardless, this PR doesn't leave the test suite or coverage in a _worse_ place than before, but it still needs work.

packages/optimizely-sdk/package.json
```diff
@@ -9,16 +9,19 @@
   "typings": "lib/index.d.ts",
   "scripts": {
     "clean": "rm -rf dist",
-    "test": "mocha ./lib/*.tests.js ./lib/**/*.tests.js ./lib/**/**/*tests.js --recursive --exit --require esm --require lib/tests/exit_on_unhandled_rejection.js",
+    "lint": "eslint 'lib/**/*.js'",
+    "test": "mocha --exit -r esm -r lib/tests/exit_on_unhandled_rejection.js 'lib/**/*tests.js'",
```
mjc1283, line 13: After running this locally, I noticed this glob pattern picks up a file that it shouldn't: `lib/index.browser.umdtests.js`. That one is intended to be run via Karma (`npm run test-umdbrowser`) and doesn't work with a plain mocha run. Can we use this pattern instead? 
```suggestion
    "test": "mocha --exit -r esm -r lib/tests/exit_on_unhandled_rejection.js 'lib/**/*.tests.js'",
```

jasonkarns: Figures... that was the pattern I had before, until I noticed that the original code had this:

`mocha ./lib/*.tests.js ./lib/**/*.tests.js ./lib/**/**/*tests.js`

(The last pattern was `*tests.js`.)

If I could be so bold, I would suggest this indicates a problem with the way the tests are organized. If the umdtests file shouldn't be run as part of the mocha suite, shouldn't it be placed elsewhere? (Like `test/umd-integration.js` or something?)

jasonkarns: I've updated the PR by removing that last commit.

packages/optimizely-sdk/package.json
```diff
@@ -9,16 +9,19 @@
...
     "build": "rollup -c",
     "build-browser-umd": "rollup -c --config-umd",
-    "test-ci": "npm run test-xbrowser && npm run test-umdbrowser",
-    "lint": "eslint 'lib/**/*.js'",
-    "cover": "istanbul cover _mocha ./lib/*.tests.js ./lib/**/*.tests.js ./lib/**/**/*tests.js",
-    "coveralls": "npm run cover -- --report lcovonly && cat ./coverage/lcov.info | coveralls",
+    "precover": "nyc npm test",
+    "cover": "nyc report -r lcov",
+    "precoveralls": "npm run cover",
+    "coveralls": "< coverage/lcov.info coveralls",
```
jasonkarns, line 24: To clarify the weirdness of putting the file redirection in front: this allows extra flags to be passed when running via npm:

`npm run coveralls -- --foo` will pass `--foo` along to the coveralls command. If input/output redirection is tacked on at the _end_ of the script as is common in shell, then that would cause a malformed command.

`

packages/optimizely-sdk/package.json
mjc1283: Yea that's fair, we should move it elsewhere. I'll follow up on this.

jasonkarns: Regardless, the "restored" glob is back to excluding umdtests, so that extraction need not affect this PR :D

PR #476, opened 2020-05-07
Updates engines.node stanza to reflect actual requirements

mjc1283:


jasonkarns:
tangential question? Why is this duplicate PR necessary? I checked the option to allow maintainer edits, so if anything needs tweaked in my PRs, that can be done directly on them.

mjc1283:
@jasonkarns No asks from our end other than keeping them up to date with master. This is necessary due to the way our CI implementation handles PRs from other repositories. We are looking at ways to make this less awkward.

PR #481, opened 2020-05-08
Fix package build scripts for git-deps and npm-link symlinks

jasonkarns:
## Summary
Main change: `prepublishOnly` is the wrong lifecycle script for building/compiling. prepublishOnly does not run during install for git-deps; nor does it run when a package is being symlinked via npm-link. The correct script is `prepare`.

This PR:
- ensures the build/bundle/compile script is accurately defined as `build` so as to be run properly by the top level `lerna run build` script.
- ensures "clean" is run before building
- ensures `clean` is defined and exposed such that `lerna run clean` works properly
- ensures `build` is invoked during `prepare` lifecycle hook (packing, publishing, git-installing, npm-linking)
- preserves `npm test` being run during `prepublishOnly`
- adds package tarball pattern to gitignore so npm-pack can be run for inspection
- removes unnecessary files from `package.json#files` array (license, changelog, readme, and package.json files are included automatically)
- renames CHANGELOG.MD to be cased properly

## Test plan

No code changes

**note** It's probably also worth cleaning up the lerna publish script. Since building prior to publishing is handled by npm, it's redundant for the lerna publish script to _also_ trigger a build. However, since lerna publish does not support dry-run, I opted not to change this since I can't verify for sure.

But if (during the next `lerna publish` phase) it is observed that packages are being built twice, I would recommend changing: https://github.com/optimizely/javascript-sdk/blob/990245f234f277b2a781a170f7982b81918c50f4/package.json#L8 to be just: `"publish": "lerna publish",`


packages/datafile-manager/package.json
```diff
@@ -55,9 +51,13 @@
     "@react-native-community/async-storage": "^1.2.0"
   },
   "scripts": {
+    "clean": "rm -rf lib",
     "lint": "tsc --noEmit && eslint --fix 'src/**/*.ts' '__test__/**/*.ts'",
     "test": "jest",
-    "tsc": "rm -rf lib && tsc",
-    "prepublishOnly": "npm run lint && npm test && npm run tsc"
+    "posttest": "npm run lint",
+    "prebuild": "npm run clean",
```
mjc1283, line 58: Is `"prebuild"` a thing? I don't see it documented [here](https://docs.npmjs.com/misc/scripts). But it does appear to be working as expected when I run locally.

jasonkarns: > Pre and post commands with matching names will be run for those as well (e.g. premyscript, myscript, postmyscript)

(That's in the docs linked)

Any pre* script or post* script will run before (or after, respectively) the same-named script.

With the long-standing bug/quirk that a pre* script won't work if the '*' begins with pre. Was meant to disallow `preprefoo` running before `prefoo` which runs before `foo`. However, it has the unfortunate side effect of disallowing `preprepare`. Regardless, this is the mechanism that makes preinstall, postinstall, prepublish, postpublish,  work. Those aren't special scripts. They work because any pre/post script will run before/after the corresponding script.

PR #484, opened 2020-05-12
Ship unminified file as ESM entrypoint

jasonkarns:
## Summary

ESM entrypoints are very very likely to be consumed by bundlers, not loaded directly into browsers. Therefore it is preferable that the bundlers have access to the unminified source so that consumers can have more control over the final output.

This change _adds_ an ESM output bundle that is not minified with terser.
Notably, the unminified bundle is created _in addition to_ the minified bundle;
so the minified bundle is still distributed with the package (at the same output location).

This way users who do actually want the minified bundle may still use it.

However, the `package.json#module` entrypoint is changed to reference the unminified bundle, as this is most likely what users will want when consuming from a bundler. (And bundlers are virtually the exclusive consumers of the `module` entrypoint.)


zashraf1985:
Moved the work to #752 . Closing this

Issue #487, opened 2020-05-21
optimizely-sdk test suite is not cleaning up properly (requires dirty --exit flag for mocha)


jasonkarns:
The optimizely-sdk test suite requires the `--exit` flag to mocha. This means the test suite is not cleaning up after itself properly.

Mocha [provides some details](https://mochajs.org/#-exit) about the flag, why it is indicative of problems, and some ways to find/fix the root cause.

I ran the following to uncover _which_ of the many test files is causing node to hang:

```
set -x; for f in lib/**/*.tests.js; do npx mocha -r esm "$f"; done
```

The problematic file is `lib/optimizely/index.tests.js` (luckily only one file is causing node to hang).

**`@optimizely/optimizely-sdk` version:** 
tested at 764135b5e8e76dcc59a1a0e8d62d523a0520cd59

mjc1283:
Fixed, and removed `--exit` flag, in #497.

PR #497, opened 2020-06-02
fix: Stub event dispatcher and mock timers in tests

mjc1283:
## Summary
- Several unit tests weren't properly stubbing the default event dispatcher - add these stubs
- The test suite wasn't exiting due to active timers not being cleaned up. Fixed this by mocking `setTimeout` using `sinon.useFakeTimers`
- One test for Node default event dispatcher was relying on a real connection error from attempting to request to an invalid domain. Added a `nock` mock for this to prevent the real request.

## Test plan
Run `npm test`. Should exit immediately, and not make any outgoing HTTP requests.

## Issues
https://github.com/optimizely/javascript-sdk/issues/487

### DIFF
--- a/packages/optimizely-sdk/lib/index.browser.tests.js
+++ b/packages/optimizely-sdk/lib/index.browser.tests.js
@@ -414,7 +414,6 @@ describe('javascript-sdk', function() {
             eventDispatcher: fakeEventDispatcher,
             logger: silentLogger,
           });
-
           sinon.assert.calledWithExactly(
             eventProcessorSpy,
             sinon.match({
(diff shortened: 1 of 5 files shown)


packages/optimizely-sdk/lib/index.browser.tests.js
```diff
@@ -166,19 +177,6 @@ describe('javascript-sdk', function() {
...
-          datafile: {},
-          errorHandler: fakeErrorHandler,
-          eventDispatcher: fakeEventDispatcher,
-          logger: silentLogger,
-        });
-        // Invalid datafile causes onReady Promise rejection - catch this error
-        optlyInstance.onReady().catch(function() {});
-        assert.equal('react-sdk', optlyInstance.clientEngine);
-      });
-
```
mjc1283, line 181: Deleted this because it's identical to the test above.

PR #498, opened 2020-06-02
feat: Added support for Authenticated Datafiles

zashraf1985:
## Summary
Added support for Authenticated Datafiles

## Test plan
Will Add Unit Tests.

### DIFF
--- a/packages/datafile-manager/CHANGELOG.md
+++ b/packages/datafile-manager/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Changes that have landed but are not yet released.
 
 ### Changed
-- Added support for authenticated datafiles. `NodeDatafileManager` now accepts `datafileAccessToken` to be able to fetch authentication datafiles.
+- Added support for authenticated datafiles. `NodeDatafileManager` now accepts `datafileAccessToken` to be able to fetch authenticated datafiles.
 
 ## [0.5.0] - April 17, 2020
 
--- a/packages/datafile-manager/CHANGELOG.md
+++ b/packages/datafile-manager/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 Changes that have landed but are not yet released.
 
+### Changed
+- Added support for authenticated datafiles. `NodeDatafileManager` now accepts `datafileAccessToken` to be able to fetch authentication datafiles.
+
 ## [0.5.0] - April 17, 2020
 
 ### Breaking Changes
--- a/packages/datafile-manager/src/config.ts
+++ b/packages/datafile-manager/src/config.ts
@@ -20,7 +20,7 @@ export const MIN_UPDATE_INTERVAL = 1000;
 
 export const DEFAULT_URL_TEMPLATE = `https://cdn.optimizely.com/datafiles/%s.json`;
 
-export const DEFAULT_AUTHENTICATED_DATAFILE_URL_TEMPLATE = `https://www.optimizely-cdn.com/datafiles/auth/%s.json`;
+export const DEFAULT_AUTHENTICATED_URL_TEMPLATE = `https://www.optimizely-cdn.com/datafiles/auth/%s.json`;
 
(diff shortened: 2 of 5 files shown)


packages/datafile-manager/src/config.ts
```diff
@@ -20,6 +20,8 @@ export const MIN_UPDATE_INTERVAL = 1000;
 
 export const DEFAULT_URL_TEMPLATE = `https://cdn.optimizely.com/datafiles/%s.json`;
 
+export const DEFAULT_AUTHENTICATED_DATAFILE_URL_TEMPLATE = `https://www.optimizely-cdn.com/datafiles/auth/%s.json`;
```
mjc1283, line 23: Suggest `DEFAULT_AUTHENTICATED_URL_TEMPLATE` - everything pertains to datafiles.

packages/datafile-manager/src/datafileManager.ts
```diff
@@ -43,4 +43,5 @@ export interface DatafileManagerConfig {
   updateInterval?: number;
   urlTemplate?: string;
   cache?: PersistentKeyValueCache;
+  authDatafileToken?: string
```
mjc1283, line 46: If we want to isolate this functionality to Node, this shouldn't be added here. `DatafileManagerConfig` is used by the browser datafile manager. We should create a new interface accepted only by the Node datafile manager.

packages/datafile-manager/src/nodeDatafileManager.ts
```diff
@@ -18,9 +18,28 @@ import { makeGetRequest } from './nodeRequest';
...
+    super({
+      ... config,
+      urlTemplate: config.urlTemplate || defaultUrlTemplate,
+    });
+    this.authToken = config.authDatafileToken;
+  }
+
   protected makeGetRequest(reqUrl: string, headers: Headers): AbortableRequest {
+    if (this.authToken) {
+      headers['Authorization'] = `Bearer ${this.authToken}`;
```
mjc1283, line 41: - Let's not mutate the `headers` object, rather make a shallow copy and add this header.
- Let's log at debug level, of course NOT included the secret in the log message, but the message should indicate that we added an `Authorization: Bearer` header.

packages/datafile-manager/src/config.ts
zashraf1985: Done

packages/datafile-manager/src/datafileManager.ts
zashraf1985: Done

packages/datafile-manager/src/nodeDatafileManager.ts
zashraf1985: Done

packages/datafile-manager/src/datafileManager.ts
```diff
@@ -44,3 +44,7 @@ export interface DatafileManagerConfig {
   urlTemplate?: string;
   cache?: PersistentKeyValueCache;
 }
+
+export interface NodeDatafileManagerConfig extends DatafileManagerConfig {
```
pawels-optimizely, line 48: not sure if NodeDatafileManagerConfig is the right name for this.
Also, not sure if it would be easier to make changes in DatafileManagerConfig and HttpPollingDatafileManager directly instead of extending them.  It looks clean, but for instance, it does not make sense to do something like that in go-sdk, so maybe we can have language specific implementation.

I would like to have @mjc1283  input on that

mjc1283: @pawels-optimizely It is the right name. This class gets pulled in via a specific entry point which is intended for use in Node. We already had these browser/node-specific subclasses as a way of implementing the datafile GET request differently in each environment. This is another behavior that's different for browser vs. Node, so I think it's appropriate. I'm not strongly opposed to putting this in HttpPollingDatafileManager, and just advising against using this in client-side. But since we already have these classes in place, I don't mind having it here either.

packages/datafile-manager/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 Changes that have landed but are not yet released.
 
+### Changed
+- Added support for authenticated datafiles. `NodeDatafileManager` now accepts `datafileAccessToken` to be able to fetch authentication datafiles.
```
mjc1283, line 11:
```suggestion
- Added support for authenticated datafiles. `NodeDatafileManager` now accepts `datafileAccessToken` to be able to fetch authenticated datafiles.
```

PR #503, opened 2020-06-09
refactor: Move Async Storage Implementation to utils package

zashraf1985:
## Summary
This is first PR in an effort to implement support for Offline Events persistence for React Native Apps. Async Storage implementation was already done in `datafile-manager` package to support datafile caching. This PR moves the generic caching implementation to `utils` so that it can be used by both `datafile-manager` and `event-processor`.

## Test plan
All Existing tests should pass

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -38,8 +38,7 @@
     "access": "public"
   },
   "dependencies": {
-    "uuid": "^3.3.2",
-    "@optimizely/js-sdk-logging": "^0.1.0"
+    "uuid": "^3.3.2"
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


packages/utils/package.json
```diff
@@ -38,13 +38,18 @@
     "access": "public"
   },
   "dependencies": {
-    "uuid": "^3.3.2"
+    "uuid": "^3.3.2",
+    "@optimizely/js-sdk-logging": "^0.1.0"
```
mjc1283, line 42: I realized that this will create a circular dependency between our logging and utils packages. Please remove logging from the utils package.

zashraf1985: Done

PR #507, opened 2020-06-12
chore(utils): Prepare for js-sdk-utils v0.3.1 release

mjc1283:
Prepare for js-sdk-utils v0.3.1 release

### DIFF
--- a/packages/utils/CHANGELOG.MD
+++ b/packages/utils/CHANGELOG.MD
@@ -7,10 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 Changes that have landed but are not yet released.
 
-## [0.2.0] - June 11, 2020
+## [0.3.1] - June 12, 2020
+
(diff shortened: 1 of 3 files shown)


packages/utils/CHANGELOG.MD
```diff
@@ -7,10 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 Changes that have landed but are not yet released.
 
-## [0.2.0] - June 11, 2020
+## [0.3.1] - June 12, 2020
+
+### Bug Fixes
+- Fix exports of `PersistentKeyValueCache` and `ReactNativeAsyncStorageCache` to be named, not default
```
aliabbasrizvi, line 13: Can you tag PR also? Here and in the update for 0.3.0? Or we don't do that here?

mjc1283: We haven't been doing it, but agree it's a good practice. Will start it now.

PR #513, opened 2020-06-18
refactor: Made PendingEventsStore interface asynchronous

zashraf1985:
## Summary
Made PendingEventsStore interface asynchronous. This change is done to implement React Native Async Storage which is asynchronous using the same interface.

## Test plan
Fixed the failing unit tests

### DIFF
--- a/packages/event-processor/__tests__/pendingEventsDispatcher.spec.ts
+++ b/packages/event-processor/__tests__/pendingEventsDispatcher.spec.ts
@@ -72,10 +72,12 @@ describe('LocalStoragePendingEventsDispatcher', () => {
       expect((originalEventDispatcher.dispatchEvent as unknown) as jest.Mock).toBeCalledTimes(1)
       expect(internalDispatchCall[0]).toEqual(eventV1Request)
 
-      // assert that the passed in callback to pendingEventsDispatcher was called
-      expect(callback).toHaveBeenCalledTimes(1)
-      expect(callback).toHaveBeenCalledWith({ statusCode: 200 })
-      done()
+      setTimeout(() => {
+        // assert that the passed in callback to pendingEventsDispatcher was called
+        expect(callback).toHaveBeenCalledTimes(1)
+        expect(callback).toHaveBeenCalledWith({ statusCode: 200 })
+        done()
+      })
     })
   })
 
@@ -100,10 +102,12 @@ describe('LocalStoragePendingEventsDispatcher', () => {
       expect((originalEventDispatcher.dispatchEvent as unknown) as jest.Mock).toBeCalledTimes(1)
       expect(internalDispatchCall[0]).toEqual(eventV1Request)
 
-      // assert that the passed in callback to pendingEventsDispatcher was called
-      expect(callback).toHaveBeenCalledTimes(1)
-      expect(callback).toHaveBeenCalledWith({ statusCode: 400})
-      done()
+      setTimeout(() => {
+        // assert that the passed in callback to pendingEventsDispatcher was called
+        expect(callback).toHaveBeenCalledTimes(1)
+        expect(callback).toHaveBeenCalledWith({ statusCode: 400})
+        done()
+      })
     })
   })
 })
@@ -135,7 +139,7 @@ describe('PendingEventsDispatcher', () => {
 
   describe('dispatch', () => {
     describe('when the dispatch is successful', () => {
-      it('should save the pendingEvent to the store and remove it once dispatch is completed', async () => {
+      it('should save the pendingEvent to the store and remove it once dispatch is completed', async (done) => {
         const callback = jest.fn()
         const eventV1Request: EventV1Request = {
           url: 'http://cdn.com',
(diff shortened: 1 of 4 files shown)


packages/event-processor/__tests__/pendingEventsDispatcher.spec.ts
```diff
@@ -61,22 +61,25 @@ describe('LocalStoragePendingEventsDispatcher', () => {
...
-    internalDispatchCall[1]({ statusCode: 200 })
-
-    // assert that the original dispatch function was called with the request
-    expect((originalEventDispatcher.dispatchEvent as unknown) as jest.Mock).toBeCalledTimes(1)
-    expect(internalDispatchCall[0]).toEqual(eventV1Request)
-
-    // assert that the passed in callback to pendingEventsDispatcher was called
-    expect(callback).toHaveBeenCalledTimes(1)
-    expect(callback).toHaveBeenCalledWith({ statusCode: 200 })
+    setTimeout(() => {
```
mjc1283, line 64: Why did you add this `setTimeout`?

packages/event-processor/src/pendingEventsDispatcher.ts
```diff
@@ -53,23 +53,22 @@ export class PendingEventsDispatcher implements EventDispatcher {
...
   sendPendingEvents(): void {
-    const pendingEvents = this.store.values()
-
-    logger.debug('Sending %s pending events from previous page', pendingEvents.length)
-
-    pendingEvents.forEach(item => {
-      try {
-        this.send(item, () => {})
-      } catch (e) {}
+    this.store.values().then((pendingEvents) => {
```
mjc1283, line 56: - Did you test this in the browser to see how its behavior compares to the old version? I'm curious if the additional async step would affect how much of this is allowed by the browser to execute during a page close.
- If `close` is called while `values()` is in progress, I think we need to stop and not send the events. To the extend possible, we should stop all additional work after `close` is called.

packages/event-processor/src/pendingEventsDispatcher.ts
```diff
@@ -53,23 +53,22 @@ export class PendingEventsDispatcher implements EventDispatcher {
...
     })
   }
 
   protected send(entry: DispatcherEntry, callback: EventDispatcherCallback): void {
-    this.store.set(entry.uuid, entry)
-
-    this.dispatcher.dispatchEvent(entry.request, response => {
-      this.store.remove(entry.uuid)
-      callback(response)
+    this.store.set(entry.uuid, entry).then(() => {
```
mjc1283, line 67: Same comment as above RE: doing things after close: If `close` is called while `set(...)` is in progress, we need to stop and not send the events.

packages/event-processor/src/pendingEventsDispatcher.ts
```diff
@@ -53,23 +53,22 @@ export class PendingEventsDispatcher implements EventDispatcher {
...
+      this.dispatcher.dispatchEvent(entry.request, response => {
+        this.store.remove(entry.uuid)
```
mjc1283, line 69: I think we need to wait for remove to complete before calling the callback. This callback eventually feeds into the `close` promise status. We don't want `close` to resolve while an async operation is still pending.

packages/event-processor/src/pendingEventsDispatcher.ts
mjc1283: Discussed this offline w/Zeeshan - this part is OK as-is, disregard.

packages/event-processor/__tests__/pendingEventsDispatcher.spec.ts
zashraf1985: Because i want to detect call `dispatchEvent` on the original event dispatcher which was called synchronously earlier but now its being called asynchronously after the `set` promise resolves.

packages/event-processor/src/pendingEventsDispatcher.ts
zashraf1985: Done

packages/event-processor/src/pendingEventsDispatcher.ts
```diff
@@ -53,23 +53,21 @@ export class PendingEventsDispatcher implements EventDispatcher {
...
-
-    pendingEvents.forEach(item => {
-      try {
-        this.send(item, () => {})
-      } catch (e) {}
+    this.store.values().then((pendingEvents) => {
+      logger.debug('Sending %s pending events from previous page', pendingEvents.length)
+      pendingEvents.forEach(item => {
+        try {
+          this.send(item, () => {})
```
jaeopt, line 60: Does MAP guarantee the fifo order in JS? Otherwise, we lose the order of events dispatched.

jaeopt: Also, orders are concerned in other cases as well:
- if we have any queued events, all the following events dispatched should be queued as well instead of sending them first. We may have queued events from the previous session or new events queued from the last failed dispatch.

packages/event-processor/src/pendingEventsDispatcher.ts
```diff
@@ -53,23 +53,21 @@ export class PendingEventsDispatcher implements EventDispatcher {
...
 
   protected send(entry: DispatcherEntry, callback: EventDispatcherCallback): void {
-    this.store.set(entry.uuid, entry)
-
-    this.dispatcher.dispatchEvent(entry.request, response => {
-      this.store.remove(entry.uuid)
-      callback(response)
+    this.store.set(entry.uuid, entry).then(() => {
+      this.dispatcher.dispatchEvent(entry.request, response => {
+        this.store.remove(entry.uuid).then(() => callback(response))
```
jaeopt, line 69: We also need a protection from one packet stuck in the queue forever (cannot send to the server successfully for JSON format error etc). This can be an issue when we change them in FIFO order.

zashraf1985:
This is no longer needed now. I am now leaving the existing localstorge implementation intact and implementing event caching for react native separately in a new PR.
https://github.com/optimizely/javascript-sdk/pull/517

PR #515, opened 2020-06-30
fix(logs): Fixing log messages for Targeted Rollouts

fayyazarshad:
## Summary
With this change we are introducing targeted rollout specific log messages.

## Test plan
All Unit and FSC Tests are getting passed

### DIFF
--- a/packages/optimizely-sdk/lib/core/bucketer/index.js
+++ b/packages/optimizely-sdk/lib/core/bucketer/index.js
@@ -104,7 +104,7 @@ export var bucket = function(bucketerParams) {
   var bucketValue = this._generateBucketValue(bucketingId);
 
   var bucketedUserLogMessage = sprintf(
-    LOG_MESSAGES.USER_ASSIGNED_TO_VARIATION_BUCKET,
+    LOG_MESSAGES.USER_ASSIGNED_TO_EXPERIMENT_BUCKET,
     MODULE_NAME,
     bucketValue,
     bucketerParams.userId
--- a/packages/optimizely-sdk/lib/core/bucketer/index.js
+++ b/packages/optimizely-sdk/lib/core/bucketer/index.js
@@ -112,28 +112,11 @@ export var bucket = function(bucketerParams) {
   bucketerParams.logger.log(LOG_LEVEL.DEBUG, bucketedUserLogMessage);
 
   var entityId = this._findBucket(bucketValue, bucketerParams.trafficAllocationConfig);
-  if (!entityId) {
-    var userHasNoVariationLogMessage = sprintf(
-      LOG_MESSAGES.USER_HAS_NO_VARIATION,
-      MODULE_NAME,
-      bucketerParams.userId,
-      bucketerParams.experimentKey
-    );
-    bucketerParams.logger.log(LOG_LEVEL.DEBUG, userHasNoVariationLogMessage);
-  } else if (!bucketerParams.variationIdMap.hasOwnProperty(entityId)) {
+
+  if (!bucketerParams.variationIdMap.hasOwnProperty(entityId)) {
     var invalidVariationIdLogMessage = sprintf(LOG_MESSAGES.INVALID_VARIATION_ID, MODULE_NAME);
     bucketerParams.logger.log(LOG_LEVEL.WARNING, invalidVariationIdLogMessage);
     return null;
-  } else {
-    var variationKey = bucketerParams.variationIdMap[entityId].key;
-    var userInVariationLogMessage = sprintf(
-      LOG_MESSAGES.USER_HAS_VARIATION,
-      MODULE_NAME,
-      bucketerParams.userId,
-      variationKey,
-      bucketerParams.experimentKey
-    );
-    bucketerParams.logger.log(LOG_LEVEL.INFO, userInVariationLogMessage);
   }
 
   return entityId;
--- a/packages/optimizely-sdk/lib/core/bucketer/index.tests.js
+++ b/packages/optimizely-sdk/lib/core/bucketer/index.tests.js
@@ -76,7 +76,7 @@ describe('lib/core/bucketer', function() {
 
           var bucketedUser_log1 = createdLogger.log.args[0][1];
           expect(bucketedUser_log1).to.equal(
-            sprintf(LOG_MESSAGES.USER_ASSIGNED_TO_VARIATION_BUCKET, 'BUCKETER', '50', 'ppid1')
+            sprintf(LOG_MESSAGES.USER_ASSIGNED_TO_EXPERIMENT_BUCKET, 'BUCKETER', '50', 'ppid1')
           );
 
           var bucketerParamsTest2 = cloneDeep(bucketerParams);
@@ -86,7 +86,7 @@ describe('lib/core/bucketer', function() {
           var notBucketedUser_log1 = createdLogger.log.args[1][1];
 
           expect(notBucketedUser_log1).to.equal(
-            sprintf(LOG_MESSAGES.USER_ASSIGNED_TO_VARIATION_BUCKET, 'BUCKETER', '50000', 'ppid2')
+            sprintf(LOG_MESSAGES.USER_ASSIGNED_TO_EXPERIMENT_BUCKET, 'BUCKETER', '50000', 'ppid2')
           );
         });
       });
@@ -153,7 +153,7 @@ describe('lib/core/bucketer', function() {
 
             var log3 = createdLogger.log.args[2][1];
             expect(log3).to.equal(
-              sprintf(LOG_MESSAGES.USER_ASSIGNED_TO_VARIATION_BUCKET, 'BUCKETER', '50', 'testUser')
+              sprintf(LOG_MESSAGES.USER_ASSIGNED_TO_EXPERIMENT_BUCKET, 'BUCKETER', '50', 'testUser')
             );
           });
 
@@ -247,7 +247,7 @@ describe('lib/core/bucketer', function() {
             sinon.assert.calledOnce(createdLogger.log);
 
             var log1 = createdLogger.log.args[0][1];
-            expect(log1).to.equal(sprintf(LOG_MESSAGES.USER_ASSIGNED_TO_VARIATION_BUCKET, 'BUCKETER', '0', 'testUser'));
+            expect(log1).to.equal(sprintf(LOG_MESSAGES.USER_ASSIGNED_TO_EXPERIMENT_BUCKET, 'BUCKETER', '0', 'testUser'));
           });
 
           it('should return null when a user does not fall into an experiment within an overlapping group', function() {
--- a/packages/optimizely-sdk/lib/core/bucketer/index.tests.js
+++ b/packages/optimizely-sdk/lib/core/bucketer/index.tests.js
@@ -75,27 +75,18 @@ describe('lib/core/bucketer', function() {
           expect(bucketer.bucket(bucketerParamsTest1)).to.equal('111128');
 
           var bucketedUser_log1 = createdLogger.log.args[0][1];
-          var bucketedUser_log2 = createdLogger.log.args[1][1];
-
           expect(bucketedUser_log1).to.equal(
             sprintf(LOG_MESSAGES.USER_ASSIGNED_TO_VARIATION_BUCKET, 'BUCKETER', '50', 'ppid1')
-          );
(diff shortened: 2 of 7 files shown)


packages/optimizely-sdk/lib/core/bucketer/index.js
```diff
@@ -112,28 +112,11 @@ export var bucket = function(bucketerParams) {
...
-    var userHasNoVariationLogMessage = sprintf(
-      LOG_MESSAGES.USER_HAS_NO_VARIATION,
-      MODULE_NAME,
-      bucketerParams.userId,
-      bucketerParams.experimentKey
-    );
-    bucketerParams.logger.log(LOG_LEVEL.DEBUG, userHasNoVariationLogMessage);
-  } else if (!bucketerParams.variationIdMap.hasOwnProperty(entityId)) {
+
+  if (!bucketerParams.variationIdMap.hasOwnProperty(entityId)) {
```
oakbani, line 116: @zashraf1985 Can you check if we really need this condition? I don't see invalid_variation handling in other SDKs.

packages/optimizely-sdk/lib/core/decision_service/index.js
```diff
@@ -456,14 +475,14 @@ DecisionService.prototype._getVariationForRollout = function(configObj, feature,
...
+  var rolloutRule;
   var bucketerParams;
   var variationId;
   var variation;
   for (index = 0; index < endIndex; index++) {
-    experiment = configObj.experimentKeyMap[rollout.experiments[index].key];
+    rolloutRule = configObj.experimentKeyMap[rollout.experiments[index].key];
 
-    if (!this.__checkIfUserIsInAudience(configObj, experiment.key, userId, attributes)) {
+    if (!this.__checkIfUserIsInAudience(configObj, rolloutRule.key, AUDIENCE_EVALUATION_TYPES.RULE, userId, attributes)) {
```
oakbani, line 485: We need to send the log index not rollout key. I think kry for rollout is some system geenrated string. 
We would like our logs to print like ' Evaluating rollout rule 1 ... '. where 1 is an integer. Only for the last rule, we print 'EveryoneElse'

packages/optimizely-sdk/lib/core/bucketer/index.js
fayyazarshad: It turns out there is a test for it as well, also it make sense to have this check as well

packages/optimizely-sdk/lib/core/decision_service/index.js
fayyazarshad: Done

fayyazarshad:
Resolved feedback from Owais, @zashraf1985 can you please have a look?

fayyazarshad:
get_feature_variable_json tests of FSC are failing, due to its name is changed in https://github.com/optimizely/javascript-sdk/pull/516 which is not merged yet but testapp PR is merged with the updated name

packages/optimizely-sdk/lib/core/decision_service/index.js
```diff
@@ -171,21 +194,24 @@ DecisionService.prototype.__getWhitelistedVariation = function(experiment, userI
...
+ * @param  {Object}  configObj            The parsed project configuration object
+ * @param  {string}  experimentKey        Key of experiment being validated
+ * @param  {string}  evaluationAttribute  String representing experiment key or rule
+ * @param  {string}  userId               ID of user
+ * @param  {Object}  attributes           Optional parameter for user's attributes
+ * @param  {string}  loggingKey           String representing experiment key or rollout rule. To be used in log messages only.
  * @return {boolean} True if user meets audience conditions
  */
-DecisionService.prototype.__checkIfUserIsInAudience = function(configObj, experimentKey, userId, attributes) {
+DecisionService.prototype.__checkIfUserIsInAudience = function(configObj, experimentKey, evaluationAttribute, userId, attributes, loggingKey) {
```
msohailhussain, line 205: Why sequence of args are changed? We don't need to rename `attributes` to evaluationAttribute.

msohailhussain: @mjc1283 In other SDKs, this method name is changed. We should rename here?

packages/optimizely-sdk/lib/utils/enums/index.js
```diff
@@ -96,7 +96,7 @@ export var LOG_MESSAGES = {
   SHOULD_NOT_DISPATCH_ACTIVATE: '%s: Experiment %s is not in "Running" state. Not activating user.',
   SKIPPING_JSON_VALIDATION: '%s: Skipping JSON schema validation.',
   TRACK_EVENT: '%s: Tracking event %s for user %s.',
-  USER_ASSIGNED_TO_VARIATION_BUCKET: '%s: Assigned variation bucket %s to user %s.',
+  USER_ASSIGNED_TO_VARIATION_BUCKET: '%s: Assigned bucket %s to user with bucketing ID %s.',
   USER_ASSIGNED_TO_EXPERIMENT_BUCKET: '%s: Assigned experiment bucket %s to user %s.',
```
aliabbasrizvi, line 100: These two should ideally be the same message. Wdyt @mjc1283 ?

packages/optimizely-sdk/lib/core/bucketer/index.tests.js
```diff
@@ -308,7 +284,7 @@ describe('lib/core/bucketer', function() {
         it('should return null', function() {
           var bucketerParamsTest1 = cloneDeep(bucketerParams);
           bucketerParamsTest1.userId = 'ppid1';
-          expect(bucketer.bucket(bucketerParamsTest1)).to.equal('');
+          expect(bucketer.bucket(bucketerParamsTest1)).to.equal(null);
```
aliabbasrizvi, line 287: How did this change?

packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -6021,27 +6021,27 @@ describe('lib/optimizely', function() {
...
               createdLogger.log,
               LOG_LEVEL.INFO,
-              'OPTIMIZELY: Feature "test_feature" is not enabled for user user1. Returning default value for variable "new_content".'
+              'OPTIMIZELY: Feature "test_feature" is not enabled for user user1. Returning the default variable value "new_content".'
             );
             sinon.assert.calledWith(
               createdLogger.log,
               LOG_LEVEL.INFO,
-              'OPTIMIZELY: Feature "test_feature" is not enabled for user user1. Returning default value for variable "price".'
+              'OPTIMIZELY: Feature "test_feature" is not enabled for user user1. Returning the default variable value "price".'
```
aliabbasrizvi, line 6029: The usage of `FEATURE_NOT_ENABLED_RETURN_DEFAULT_VARIABLE_VALUE` needs to be fixed in code. What I am seeing here are variable names and not variable values. Please fix and apply to this PR.

aliabbasrizvi: See https://github.com/optimizely/javascript-sdk/blob/8ca4242b6e3806083f6bbcdeb7e6d898a3b220e4/packages/optimizely-sdk/lib/optimizely/index.js#L820-L828. It is incorrect. Please fix the error there and adjust message to specify variable value for variable key.

packages/optimizely-sdk/lib/core/decision_service/index.js
```diff
@@ -171,21 +194,24 @@ DecisionService.prototype.__getWhitelistedVariation = function(experiment, userI
 
 /**
  * Checks whether the user is included in experiment audience
- * @param  {Object}  configObj     The parsed project configuration object
- * @param  {string}  experimentKey Key of experiment being validated
- * @param  {string}  userId        ID of user
- * @param  {Object}  attributes    Optional parameter for user's attributes
+ * @param  {Object}  configObj            The parsed project configuration object
+ * @param  {string}  experimentKey        Key of experiment being validated
```
aliabbasrizvi, line 198: What is the purpose of `experimentKey`? Isn't `loggingKey` already fulfilling the requirement?

packages/optimizely-sdk/lib/core/bucketer/index.tests.js
fayyazarshad: As removed these conditions from bucketer/index.js
if (!entityId) {
    var userHasNoVariationLogMessage = sprintf(
      LOG_MESSAGES.USER_HAS_NO_VARIATION,
      MODULE_NAME,
      bucketerParams.userId,
      bucketerParams.experimentKey
    );
    bucketerParams.logger.log(LOG_LEVEL.DEBUG, userHasNoVariationLogMessage);
} else {
    var variationKey = bucketerParams.variationIdMap[entityId].key;
    var userInVariationLogMessage = sprintf(
      LOG_MESSAGES.USER_HAS_VARIATION,
      MODULE_NAME,
      bucketerParams.userId,
      variationKey,
      bucketerParams.experimentKey
    );
    bucketerParams.logger.log(LOG_LEVEL.INFO, userInVariationLogMessage);
} 
and moved these logs to decision service, previously when we had entityId = '' it use to pass the first if condition and return empty string, however it is intended to return null as per test as well. So from now on it returns null if entityId = '' or  any value doesn't exists in variationIdmap

packages/optimizely-sdk/lib/optimizely/index.tests.js
fayyazarshad: good catch

packages/optimizely-sdk/lib/core/decision_service/index.js
fayyazarshad: **experimentKey** contains experiment or rollout key, whereas **loggingKey** is being used for that index logic for **getVariationForRollout**

packages/optimizely-sdk/lib/utils/enums/index.js
mjc1283: Yes I agree, let's combine these @fayyazarshad.

fayyazarshad: done

packages/optimizely-sdk/lib/utils/enums/index.js
```diff
@@ -97,7 +97,7 @@ export var LOG_MESSAGES = {
   SKIPPING_JSON_VALIDATION: '%s: Skipping JSON schema validation.',
   TRACK_EVENT: '%s: Tracking event %s for user %s.',
   USER_ASSIGNED_TO_VARIATION_BUCKET: '%s: Assigned bucket %s to user with bucketing ID %s.',
-  USER_ASSIGNED_TO_EXPERIMENT_BUCKET: '%s: Assigned experiment bucket %s to user %s.',
+  USER_ASSIGNED_TO_EXPERIMENT_BUCKET: '%s: Assigned bucket %s to user with bucketing ID %s.',
```
mjc1283: `USER_ASSIGNED_TO_EXPERIMENT_BUCKET` and `USER_ASSIGNED_TO_EXPERIMENT_BUCKET` are the same, so we don't need both. Please delete `USER_ASSIGNED_TO_VARIATION_BUCKET` and replace all uses of that with `USER_ASSIGNED_TO_EXPERIMENT_BUCKET`.

packages/optimizely-sdk/lib/utils/enums/index.js
```diff
@@ -96,8 +96,8 @@ export var LOG_MESSAGES = {
   SHOULD_NOT_DISPATCH_ACTIVATE: '%s: Experiment %s is not in "Running" state. Not activating user.',
   SKIPPING_JSON_VALIDATION: '%s: Skipping JSON schema validation.',
   TRACK_EVENT: '%s: Tracking event %s for user %s.',
-  USER_ASSIGNED_TO_VARIATION_BUCKET: '%s: Assigned variation bucket %s to user %s.',
-  USER_ASSIGNED_TO_EXPERIMENT_BUCKET: '%s: Assigned experiment bucket %s to user %s.',
+  USER_ASSIGNED_TO_VARIATION_BUCKET: '%s: Assigned bucket %s to user with bucketing ID %s.',
+  USER_ASSIGNED_TO_EXPERIMENT_BUCKET: '%s: Assigned bucket %s to user with bucketing ID %s.',
```
aliabbasrizvi, line 100: You can just use the same variable. Don't know why we are redeclaring.

aliabbasrizvi:
@oakbani dismissing your review to be able to merge.

PR #517, opened 2020-07-03
feat: Added Offline storage support to Event processor for React Native Apps

zashraf1985:
## Summary
Added Offline storage support to Event Processor for React Native Apps.
1. Stores events when device is offline and then redispatches when the device is online again, new event is ready to be dispatched or SDK is re-initialized.
2. Adds a new `eventMaxQueueSize` config option which represents number of offline events SDK will store at a time. Deault is 10k for now.
3. Adds a connectivity listener to redispatch events when device is back online.

Follow-up changes expected in `optimizely-sdk` package.
1. Add a new option to `EventProcessor` constructor to represent `maxQueueSize` in `React Native` entry point.
2. Change callback parameter to http status in `DefaultEventDispacher` for browser entry point which is also used for React Native entry point.
3. Add `AsyncStorage` module configuration for umd bundles in `rollup` config.

## Test plan
1. Wrote new unit tests
2. Tested offline with FSC suite to make sure it does not break existing event processor functionality.
3. Tested sequence of redispatching events using a local mock server and timestamps.


packages/event-processor/src/reactNativeEventProcessor.ts
```diff
@@ -0,0 +1,170 @@
...
+    if (this.isProcessingPendingEvents && this.pendingEventsPromise) {
+      return this.pendingEventsPromise
+    }
+    this.pendingEventsPromise = new Promise(async (resolvePendingEventPromise) => {
+      this.isProcessingPendingEvents = true
+      const formattedEvents: {[key: string]: any} = await this.pendingEventsStore.getEventsMap()
+      const eventKeys = Object.keys(formattedEvents)
+      for (let i = 0; i < eventKeys.length; i++) {
+        const eventKey = eventKeys[i]
+        if (this.eventsInProgress[eventKey]) {
```
mjc1283, line 124: Why do we need this check? How does the same event get dispatched twice?

packages/event-processor/src/eventDispatcher.ts
```diff
@@ -15,13 +15,7 @@
  */
 import { EventV1 } from "./v1/buildEventV1";
 
-export type EventDispatcherResponse = {
-  statusCode: number
-} | {
-  status: number
-}
```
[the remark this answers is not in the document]
zashraf1985: I think i changed it initially to try something quickly and then forgot to revert back. i will reverts it

packages/event-processor/src/index.ts
```diff
@@ -13,10 +13,11 @@
...
+
 export * from './events'
 export * from './eventProcessor'
 export * from './eventDispatcher'
 export * from './managed'
 export * from './pendingEventsDispatcher'
 export * from './v1/buildEventV1'
-export * from './v1/v1EventProcessor'
\ No newline at end of file
+export { LogTierV1EventProcessor } from './v1/v1EventProcessor'
```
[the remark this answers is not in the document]
zashraf1985: Because i exported both RN and normal event processor from the single file. i will probably separate them out.

packages/event-processor/src/v1/v1EventProcessor.react_native.ts
```diff
@@ -0,0 +1,199 @@
...
+const logger = getLogger('ReactNativeEventProcessor')
+
+const DEFAULT_MAX_QUEUE_SIZE = 10000
+const PENDING_EVENTS_STORE_KEY = 'fs_optly_pending_events'
+const EVENT_BUFFER_STORE_KEY = 'fs_optly_event_buffer'
+
+/**
+ * React Native Events Processor with Caching support for events when app is offline.
+ */
+export abstract class LogTierV1EventProcessor implements EventProcessor {
```
mjc1283, line 55: Can we name it `ReactNativeEventProcessor`? Or something with React Native in the name to differentiate from the standard one.

packages/event-processor/src/v1/v1EventProcessor.react_native.ts
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  async getPendingEventsPromise(): Promise<void>{
+    const formattedEvents: {[key: string]: any} = await this.pendingEventsStore.getEventsMap()
+    const eventEntries = objectEntries(formattedEvents)
+    // Using for loop to be able to wait for previous dispatch to finish before moving on to the new one
+    for (let i = 0; i < eventEntries.length; i++) {
+      const [eventKey, event] = eventEntries[i]
+      await this.dispatchEvent(eventKey, event)
+    }
```
mjc1283, line 164:
```suggestion
    for (const [eventKey, event] of eventEntries) {
      await this.dispatchEvent(eventKey, event)
    }
```

packages/event-processor/src/v1/v1EventProcessor.react_native.ts
```diff
@@ -0,0 +1,199 @@
...
+      sendEventNotification(this.notificationCenter, event)
+    })
+    // Tracking all the requests to dispatch to make sure request is completed before fulfilling the `stop` promise
+    this.requestTracker.trackRequest(requestPromise)
+    return requestPromise
+  }
+
+  process(event: ProcessableEvent): void {
+    // Adding events to buffer store. If app closes before dispatch, we can reprocess next time the app initializes
+    this.eventBufferStore.set(generateUUID(), event)
```
mjc1283, line 184: It is awkward that we have to generate the UUID to use as a key here, but we don't retrieve it using this key, we just get all of them by calling `getEventsList`. Consider changing the interface so you can set/append without passing a key.

packages/event-processor/src/v1/v1EventProcessor.react_native.ts
```diff
@@ -0,0 +1,206 @@
...
+    }
+  }
+
+  isSuccessResponse(status: number): boolean {
+    return status >= 200 && status < 400
+  }
+
+  async drainQueue(buffer: ProcessableEvent[]): Promise<void> {
+    // Retry pending failed events while draining queue
+    await this.processPendingEvents()
```
jaeopt, line 130: Not sure about the implication of this out-of-order dispatching, but if "processPendingEvents" fail to flush all pending events, the new dispatch below (line 147) can be fired before previous buffered events. Check it out.

packages/event-processor/src/v1/v1EventProcessor.react_native.ts
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  async getPendingEventsPromise(): Promise<void>{
+    const formattedEvents: {[key: string]: any} = await this.pendingEventsStore.getEventsMap()
+    const eventEntries = objectEntries(formattedEvents)
+    logger.debug('Processing %s pending events', eventEntries.length)
+    // Using for loop to be able to wait for previous dispatch to finish before moving on to the new one
+    for (const [eventKey, event] of eventEntries) {
+      await this.dispatchEvent(eventKey, event)
+    }
```
jaeopt, line 169: Another case of out-of-order delivery. If some events failed with errors, they stay in the buffer for retry later (out-of-order) since other events buffered later can be dispatched ok.

packages/event-processor/src/v1/v1EventProcessor.react_native.ts
```diff
@@ -0,0 +1,206 @@
...
+   * This stores individual events generated from the SDK till they are part of the pending buffer.
+   * The store is cleared right before the event is formatted to be dispatched.
+   * This is to make sure that individual events are not lost when app closes before the buffer was flushed.
+   */
+  private eventBufferStore: ReactNativeEventsStore<ProcessableEvent>
+  
+  constructor({
+    dispatcher,
+    flushInterval = 30000,
+    batchSize = 3000,
```
jaeopt, line 81: Check out this huge batch size default value. I understand we have max batched event size. 3000 events will be batched to be > 3MB. Not sure if this fits in the restrictions. 
I think default batch size can be much smaller like 10-100.

packages/event-processor/src/reactNativeEventsStore.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  public async getEventsMap(): Promise<{[key: string]: T}> {
+    return await this.cache.get(this.storeKey) || {}
+  }
+
+  public async getEventsList(): Promise<T[]> {
+    await this.synchronizer.getLock()
+    const eventsMap: {[key: string]: T} = await this.cache.get(this.storeKey) || {}
+    this.synchronizer.releaseLock()
+    return objectValues(eventsMap)
```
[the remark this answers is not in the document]
zashraf1985: Yes. Looping through map always iterates in the order of insertion.

packages/event-processor/src/v1/v1EventProcessor.react_native.ts
zashraf1985: Good point. I actually borrowed the value from the existing event processor. @mjc1283 ! Do you suggest to change it to something else.

zashraf1985: actually @jaeopt you are correct. this particular default never gets used by optimizely SDK, the sdk package already sets default of `10` if the user does not provide a value. This default will only take effect if someone uses event processor package directly. I will change it to `10` to make it consistent with SDK default.

packages/event-processor/src/v1/v1EventProcessor.react_native.ts
zashraf1985: fixed. skipping subsequent events on failure now.

PR #524, opened 2020-07-17
feat: Convert project_config_schema and json_schema_validator modules to TS

yavorona:
## Summary

Convert `project_config_schema` and `json_schema_validator` modules from JS to TS.

## Test plan
- Manual testing using local build and linked Node app
- Existing unit tests

### DIFF
--- a/packages/optimizely-sdk/lib/core/project_config/project_config_schema.ts
+++ b/packages/optimizely-sdk/lib/core/project_config/project_config_schema.ts
@@ -17,7 +17,9 @@
 /**
  * Project Config JSON Schema file used to validate the project json datafile
  */
-export var schema = {
+ import { JSONSchema4 } from 'json-schema';
+
+ var schemaDefinition = {
   $schema: 'http://json-schema.org/draft-04/schema#',
   type: 'object',
   properties: {
@@ -276,4 +278,6 @@ export var schema = {
   },
 };
 
-export default schema;
+const schema = schemaDefinition as JSONSchema4
+
+export { schema }
--- a/packages/optimizely-sdk/lib/core/project_config/project_config_schema.ts
+++ b/packages/optimizely-sdk/lib/core/project_config/project_config_schema.ts
@@ -280,4 +280,4 @@
 
 const schema = schemaDefinition as JSONSchema4
 
-export { schema }
+export default schema
--- a/packages/optimizely-sdk/lib/utils/json_schema_validator/index.ts
+++ b/packages/optimizely-sdk/lib/utils/json_schema_validator/index.ts
@@ -17,21 +17,21 @@ import { sprintf } from '@optimizely/js-sdk-utils';
 import { validate as jsonSchemaValidator } from 'json-schema';
 
 import { ERROR_MESSAGES } from '../enums';
-import projectConfigSchema from '../../core/project_config/project_config_schema';
+import { schema } from '../../core/project_config/project_config_schema';
 
-var MODULE_NAME = 'JSON_SCHEMA_VALIDATOR';
+const MODULE_NAME = 'JSON_SCHEMA_VALIDATOR';
 
 /**
  * Validate the given json object against the specified schema
- * @param  {Object} jsonObject The object to validate against the schema
- * @return {Boolean}           True if the given object is valid
+ * @param  {object} jsonObject The object to validate against the schema
+ * @return {boolean}           true if the given object is valid
  */
-export var validate = function(jsonObject) {
+export function validate(jsonObject: Record<string, unknown>): boolean {
   if (!jsonObject) {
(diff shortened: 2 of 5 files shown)


packages/optimizely-sdk/lib/utils/json_schema_validator/index.ts
```diff
@@ -17,21 +17,21 @@ import { sprintf } from '@optimizely/js-sdk-utils';
...
 
 /**
  * Validate the given json object against the specified schema
- * @param  {Object} jsonObject The object to validate against the schema
- * @return {Boolean}           True if the given object is valid
+ * @param  {object} jsonObject The object to validate against the schema
+ * @return {boolean}           true if the given object is valid
  */
-export var validate = function(jsonObject) {
+export function validate(jsonObject: Record<string, unknown>): boolean {
```
mikeng13, line 29: I think instead of using `Record...` you wanna just use `any` so we can validate anything against this. An alternative type to use maybe would be `JSON`

packages/optimizely-sdk/lib/core/project_config/project_config_schema.ts
```diff
@@ -276,4 +278,6 @@ export var schema = {
   },
 };
 
-export default schema;
+const schema = schemaDefinition as JSONSchema4
+
+export { schema }
```
mikeng13, line 283: I think it makes sense to keep the `export default schema` here since we will only have main export from this module. 

Then down below instead of importing `import { schema } ....` you would simply `import schema from ...`

packages/optimizely-sdk/lib/utils/json_schema_validator/index.ts
yavorona: I have tried `any` before , which gives `Argument 'jsonObject' should be typed with a non-any type` error. Did not think of `JSON`, thanks @mikeng13!

packages/optimizely-sdk/lib/utils/json_schema_validator/index.ts
```diff
@@ -17,21 +17,21 @@ import { sprintf } from '@optimizely/js-sdk-utils';
...
 
 /**
  * Validate the given json object against the specified schema
- * @param  {Object} jsonObject The object to validate against the schema
- * @return {Boolean}           True if the given object is valid
+ * @param  {object} jsonObject The object to validate against the schema
+ * @return {boolean}           true if the given object is valid
  */
-export var validate = function(jsonObject) {
+export function validate(jsonObject: JSON): boolean {
```
mjc1283, line 29: `JSON` isn't the right type for this. This `JSON` is the global `JSON` object with methods `parse` and `stringify`. See [here](https://github.com/microsoft/TypeScript/blob/b145eaf160e6b2729aaacb89856bc2346f85ac4c/lib/lib.es5.d.ts#L1052).

I think the right type for this is `{}`, an empty object with no properties. This is what's expected as the `jsonSchemaValidate` function's first argument.

packages/optimizely-sdk/lib/core/project_config/project_config_schema.ts
```diff
@@ -276,4 +278,6 @@ export var schema = {
   },
 };
 
-export default schema;
+const schema = schemaDefinition as JSONSchema4
```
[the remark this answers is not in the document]
yavorona: I agree, but I have tried it this way, and it complains `Type '{ projectId: { type: "string"; required: true; }; accountId: ... ... }] is not assignable to type '{ [k: string]: JSONSchema4; }'`.

packages/optimizely-sdk/lib/utils/json_schema_validator/index.ts
yavorona: @mjc1283 Here is the eslint error from using  `{}`:
```
error  Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead  @typescript-eslint/ban-types
```
That is why I had `Record<string, unknown>` before. I assume any object also include an empty object?

mjc1283: Hmm yea, this is forbidden by our lint rules. Actually I'm not sure why - I'm going to research it. Let's change it to `unknown`.

packages/optimizely-sdk/lib/core/project_config/project_config_schema.ts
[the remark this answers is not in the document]
yavorona: As discussed, I will create additional Jira tasks to fix/improve json schema validation and to document all the changes that we should do. I left `const schema = schemaDefinition as JSONSchema4` so that all our unit tests can still pass without having to modify any of json schema files. FYI @mjc1283 @mikeng13

PR #525, opened 2020-07-17
Revert "refactor: Update datafile-manager to use Async Storage implementation from utils (#508)"

oakbani:
## Summary

This reverts commit 9af51d840ce936726f88bc6d40a70933a8497639.
ReactNative Async Storage implementation was previosuly moved to Utils package in order to use it both in Datafile Manager and Event Processor. While testing the new offline storage in event processor, it was observed that event processor asked for installing react-native-community/async storage package even when run for node or browser. This was because we didn't have separate entrypoints in the utils package. It was decided to move react native async storage implementation to both datafile manager and event processor. 

## Test plan
All existing tests should pass

### DIFF
--- a/packages/datafile-manager/package.json
+++ b/packages/datafile-manager/package.json
@@ -49,7 +49,7 @@
   },
   "dependencies": {
     "@optimizely/js-sdk-logging": "^0.1.0",
-    "@optimizely/js-sdk-utils": "^0.2.0",
+    "@optimizely/js-sdk-utils": "^0.4.0",
     "decompress-response": "^4.2.1"
   },
   "peerDependencies": {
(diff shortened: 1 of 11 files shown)


packages/datafile-manager/package.json
```diff
@@ -48,7 +49,7 @@
   },
   "dependencies": {
     "@optimizely/js-sdk-logging": "^0.1.0",
-    "@optimizely/js-sdk-utils": "^0.3.2",
+    "@optimizely/js-sdk-utils": "^0.2.0",
```
mjc1283, line 52: I think this dependency should stay at `"^0.3.2"`

zashraf1985: Updated it to the newly released `0.4.0`

PR #526, opened 2020-07-18
feat: Convert user_profile_service_validator module to TS

yavorona:
## Summary

Convert `user_profile_service_validator` module from JS to TS.

## Test plan
- Manual testing using local build and linked Node app
- Existing unit tests

### DIFF
--- a/packages/optimizely-sdk/lib/utils/user_profile_service_validator/index.ts
+++ b/packages/optimizely-sdk/lib/utils/user_profile_service_validator/index.ts
@@ -21,23 +21,20 @@
 import { sprintf } from '@optimizely/js-sdk-utils';
 
 import { ERROR_MESSAGES } from '../enums';
-var MODULE_NAME = 'USER_PROFILE_SERVICE_VALIDATOR';
+
+const MODULE_NAME = 'USER_PROFILE_SERVICE_VALIDATOR';
 
 /**
  * Validates user's provided user profile service instance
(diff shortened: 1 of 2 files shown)


packages/optimizely-sdk/lib/utils/user_profile_service_validator/index.ts
```diff
@@ -21,23 +21,20 @@
...
 /**
  * Validates user's provided user profile service instance
- * @param  {Object}  userProfileServiceInstance
- * @return {boolean} True if the instance is valid
+ * @param  {object}  userProfileServiceInstance
+ * @return {boolean} true if the instance is valid
  * @throws If the instance is not valid
  */
-export var validate = function(userProfileServiceInstance) {
+export function validate(userProfileServiceInstance: Record<string, unknown>): boolean {
```
mikeng13, line 33: This shouldn't be a Record! We should define an interface for `UserProfileService` instead

yavorona: `UserProfileService` interface has been declared in `index.d.ts`, so I tried importing it with 
`{ UserProfileService } from '@optimizely/optimizely-sdk'` and `import { UserProfileService } from '../../index'`, which did not work. Any suggestion on how to import it would be much appreciated! @mjc1283 @mikeng13

PR #527, opened 2020-07-20
build: Use new Event Processor and Datafile Manager

oakbani:
## Summary
This PR updates package.json to use new Event Processor and Datafile Manager inside optimizely-sdk package. 
Also updates the key for batch size. 

### todo
- Update ep and dfm package versions in package.json after releasing.

## Issues
- "THING-1234" or "Fixes #123"


zashraf1985:
Made another one with the same work

PR #528, opened 2020-07-21
feat: Convert event_tags_validator module to TS

yavorona:
## Summary

- Convert `event_tags_validator` module from JS to TS
- Revise `user_profile_service_validator` module

## Test plan
- Existing unit tests

### DIFF
--- a/packages/optimizely-sdk/lib/utils/event_tags_validator/index.ts
+++ b/packages/optimizely-sdk/lib/utils/event_tags_validator/index.ts
@@ -21,22 +21,20 @@ import { sprintf } from '@optimizely/js-sdk-utils';
 
 import { ERROR_MESSAGES } from '../enums';
 
-var MODULE_NAME = 'EVENT_TAGS_VALIDATOR';
+import { EventTags } from '../../../../event-processor/src/events';
+
+const MODULE_NAME = 'EVENT_TAGS_VALIDATOR';
 
 /**
  * Validates user's provided event tags
- * @param  {Object}  event tags
- * @return {boolean} True if event tags are valid
+ * @param  {object}  eventTags
+ * @return {boolean} true if event tags are valid
  * @throws If event tags are not valid
  */
-export var validate = function(eventTags) {
+export function validate(eventTags: EventTags): boolean {
   if (typeof eventTags === 'object' && !Array.isArray(eventTags) && eventTags !== null) {
     return true;
   } else {
     throw new Error(sprintf(ERROR_MESSAGES.INVALID_EVENT_TAGS, MODULE_NAME));
   }
-}
-
-export default {
-  validate: validate,
 }
--- a/packages/optimizely-sdk/lib/utils/event_tags_validator/index.ts
+++ b/packages/optimizely-sdk/lib/utils/event_tags_validator/index.ts
@@ -25,7 +25,7 @@ const MODULE_NAME = 'EVENT_TAGS_VALIDATOR';
 
 /**
  * Validates user's provided event tags
- * @param  {object}  eventTags
+ * @param  {unknown}  eventTags
  * @return {boolean} true if event tags are valid
  * @throws If event tags are not valid
  */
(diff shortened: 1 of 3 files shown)


packages/optimizely-sdk/lib/utils/user_profile_service_validator/index.ts
```diff
@@ -30,6 +30,8 @@ const MODULE_NAME = 'USER_PROFILE_SERVICE_VALIDATOR';
  * @return {boolean} true if the instance is valid
  * @throws If the instance is not valid
  */
+
+//TODO: Use defined UserProfileService interface instead of 'Record<string, unknown>'
```
yavorona, line 34: `UserProfileService` interface is declared in '@optimizely/optimizely-sdk' molude in `index.d.ts`, so I tried importing it with 
`{ UserProfileService } from '@optimizely/optimizely-sdk'` and also `import { UserProfileService } from '../../index'`, which did not work. Any suggestions on how to import it would be much appreciated! @mjc1283 @mikeng13

mikeng13: Hmm yeah I think we'll have to do something a little more drastic. Like the UserProfileService interface needs to be defined in one of the `.ts` files, maybe it can be its own file under `plugins` and then the top level `index.d.ts` would import that interface and re-export. @mjc1283 wdyt?

packages/optimizely-sdk/lib/utils/event_tags_validator/index.ts
```diff
@@ -21,22 +21,20 @@ import { sprintf } from '@optimizely/js-sdk-utils';
...
 /**
  * Validates user's provided event tags
- * @param  {Object}  event tags
- * @return {boolean} True if event tags are valid
+ * @param  {object}  eventTags
+ * @return {boolean} true if event tags are valid
  * @throws If event tags are not valid
  */
-export var validate = function(eventTags) {
+export function validate(eventTags: EventTags): boolean {
```
mjc1283, line 34: I think the right type for `eventTags` is `unknown`. This is a value passed in by the user of unknown type.

packages/optimizely-sdk/lib/utils/user_profile_service_validator/index.ts
mjc1283: For me, `import { UserProfileService } from '@optimizely/optimizely-sdk';` works, @yavorona we can dig to this to figure out what's wrong. Maybe it's related to IDE configuration? I was able to get it to compile with `npm run build`.

But in any case, I don't think we want to use that `UserProfileService` type here. Again I think `unknown` is what we need, treating this as an unvalidated value (the user could have passed in anything).

@mikeng13 
> Like the UserProfileService interface needs to be defined in one of the .ts files, maybe it can be its own file under plugins and then the top level index.d.ts would import that interface and re-export. 

I think this will work. We could only export types at the top level that consumers would need, so not everything.

yavorona: > For me, `import { UserProfileService } from '@optimizely/optimizely-sdk';` works, @yavorona we can dig to this to figure out what's wrong. Maybe it's related to IDE configuration? I was able to get it to compile with `npm run build`.

Strange, even after compiling, it still ` Cannot find module '@optimizely/optimizely-sdk'` . 🤔

yavorona: > But in any case, I don't think we want to use that `UserProfileService` type here. Again I think `unknown` is what we need, treating this as an unvalidated value (the user could have passed in anything).

Changing it to `unknown` produces `TSError: ⨯ Unable to compile TypeScript:` for both `if (typeof userProfileServiceInstance.lookup !== 'function') {...}` and `else if (typeof userProfileServiceInstance.save !== 'function') {...}` . `any` in this case works though, so I think it would make sense to use it here. @mjc1283  what do you think?

PR #538, opened 2020-07-28
chore(event-processor): Updated utils version

zashraf1985:
## Summary
Updated utils version to `0.4.0`

## Test plan
All Existing tests should pass

### DIFF
--- a/packages/event-processor/package.json
+++ b/packages/event-processor/package.json
@@ -40,7 +40,7 @@
   },
   "dependencies": {
     "@optimizely/js-sdk-logging": "^0.1.0",
-    "@optimizely/js-sdk-utils": "0.2.0"
+    "@optimizely/js-sdk-utils": "^0.4.0"
   },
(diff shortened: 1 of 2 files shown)


packages/event-processor/package.json
```diff
@@ -40,7 +40,7 @@
   },
   "dependencies": {
     "@optimizely/js-sdk-logging": "^0.1.0",
-    "@optimizely/js-sdk-utils": "0.2.0"
+    "@optimizely/js-sdk-utils": "^0.4.0"
```
mikeng13, line 43: Oh we didn't ever use 3.2 in this package?

zashraf1985: We did earlier but that had a dependency related problem so we had to resort back to the version of the code we had earlier. We moved to 0.2 but then Matt rightly pointed that instead of going to an older version, we should fix the utils package and release a new version and here we are with this PR.

PR #542, opened 2020-07-28
feat: modify datafile managers to use datafile strings instead of objects

pthompson127:
Summary
----------
- Modified datafileManager and httpPollingDatafileManager to use datafile strings
- Removed unneeded JSON parsing of response body

Test Plan
----------
- Modified existing datafile manager tests to account for new changes to datafile variable type

### DIFF
--- a/packages/datafile-manager/__test__/httpPollingDatafileManager.spec.ts
+++ b/packages/datafile-manager/__test__/httpPollingDatafileManager.spec.ts
@@ -69,11 +69,11 @@ class TestDatafileManager extends HttpPollingDatafileManager {
 }
 
 const testCache: PersistentKeyValueCache = {
-  get(key: string): Promise<any | null> {
-    let val = null;
+  get(key: string): Promise<string> {
+    let val = '';
     switch (key) {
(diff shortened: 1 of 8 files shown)


packages/datafile-manager/__test__/httpPollingDatafileManager.spec.ts
```diff
@@ -134,28 +134,26 @@ describe('httpPollingDatafileManager', () => {
...
       await advanceTimersByTime(300000);
 
       expect(manager.responsePromises.length).toBe(2);
       await manager.responsePromises[1];
       expect(updateFn).toBeCalledTimes(1);
-      expect(updateFn).toBeCalledWith({
-        datafile: { fooz: 'barz' },
-      });
-      expect(manager.get()).toEqual({ fooz: 'barz' });
+      expect(updateFn.mock.calls[0][0]).toEqual({ datafile: '{"fooz": "barz"}' });
```
mjc1283, line 145: Suggest this instead:
```suggestion
      expect(updateFn).toBeCalledWith({
        datafile: '{"fooz": "barz"}',
      });
```

PR #548, opened 2020-08-01
feat: Convert fns module to TS

yavorona:
## Summary

- Convert `fns` module from JS to TS.

## Test plan
- Existing unit tests + running with local bundle

### DIFF
--- a/packages/optimizely-sdk/lib/utils/fns/index.ts
+++ b/packages/optimizely-sdk/lib/utils/fns/index.ts
@@ -15,12 +15,10 @@
  */
 import { generateUUID as uuid, keyBy as keyByUtil } from '@optimizely/js-sdk-utils';
 
-import { Config } from '@optimizely/optimizely-sdk';
-
 const MAX_SAFE_INTEGER_LIMIT = Math.pow(2, 53);
 
-
-export function assign(...args: [Config, ...Config[]]): Config {
+// eslint-disable-next-line
+export function assign(...args: [any]): any {
   if (!args[0]) {
     return {};
   }
--- a/packages/optimizely-sdk/lib/utils/fns/index.ts
+++ b/packages/optimizely-sdk/lib/utils/fns/index.ts
@@ -17,23 +17,20 @@ import { generateUUID as uuid, keyBy as keyByUtil } from '@optimizely/js-sdk-uti
 
 const MAX_SAFE_INTEGER_LIMIT = Math.pow(2, 53);
 
-//TODO: define Target type
-type Target = {
+type Config = {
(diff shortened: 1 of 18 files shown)


packages/optimizely-sdk/lib/utils/fns/index.ts
```diff
@@ -15,20 +15,23 @@
  */
 import { generateUUID as uuid, keyBy as keyByUtil } from '@optimizely/js-sdk-utils';
 
-var MAX_SAFE_INTEGER_LIMIT = Math.pow(2, 53);
+import { Config } from '@optimizely/optimizely-sdk';
```
yavorona, line 18: I believe this should be the right way of importing `Config` interface as it compiles successfully with `npm run build`. However, my TS compiler is failing with `Cannot find module '@optimizely/optimizely-sdk'` error, not sure why. Any ideas @mjc1283? 🥺

yavorona: This is now outdated. I changed this to type `any` after discussing with @mikeng13 . Please see the most recent commit for reference.

packages/optimizely-sdk/lib/utils/fns/index.ts
```diff
@@ -15,20 +15,21 @@
  */
 import { generateUUID as uuid, keyBy as keyByUtil } from '@optimizely/js-sdk-utils';
 
-var MAX_SAFE_INTEGER_LIMIT = Math.pow(2, 53);
+const MAX_SAFE_INTEGER_LIMIT = Math.pow(2, 53);
 
-export var assign = function (target) {
-  if (!target) {
+// eslint-disable-next-line
+export function assign(...args: [any]): any {
```
mjc1283, line 21: Is there a better type signature we can use here? The type of `Object.assign` is `assign<T, U>(target: T, source: U): T & U`.

packages/optimizely-sdk/lib/utils/fns/index.ts
```diff
@@ -15,20 +15,21 @@
  */
 import { generateUUID as uuid, keyBy as keyByUtil } from '@optimizely/js-sdk-utils';
 
-var MAX_SAFE_INTEGER_LIMIT = Math.pow(2, 53);
+const MAX_SAFE_INTEGER_LIMIT = Math.pow(2, 53);
 
-export var assign = function (target) {
+// eslint-disable-next-line
+export function assign(target: any, ...sources: any[]): any {
```
yavorona, line 21: @mjc1283 let me know if that works. I assumed we still want to go with `any` is this case.

PR #553, opened 2020-08-11
refactor: Consolidate Javascript SDK packages to depend on local versions of each other

zashraf1985:
## Summary
Javascript SDK uses released versions of sub packages. This makes it difficult to work on features that need to change multiple sub packages. This PR consolidates all the sub packages to use local versions of each other.

## Test plan
All existing unit and FSC tests should pass


zashraf1985:
Decided to consolidate in another way

PR #556, opened 2020-08-20
feat: Convert config_validator to TS

yavorona:
## Summary

- Convert `config_validator` module from JS to TS.

## Test plan
- Existing unit tests


packages/optimizely-sdk/lib/index.node.js
```diff
@@ -70,7 +70,7 @@ var createInstance = function(config) {
     }
 
     try {
-      configValidator.validate(config);
+      validate(config);
```
mjc1283, line 73: Same comment re: `validate` vs. `configValidator.validate`

mjc1283:
@yavorona Please take a look at the FSC failure, I'm not sure what happened there.

PR #561, opened 2020-09-01
feat: Convert custom_attribute_condition_evaluator to TS

yavorona:
## Summary

- Convert `custom_attribute_condition_evaluator` module from JS to TS.

## Test plan
- Existing unit tests and local bundle

### DIFF
--- a/packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.ts
+++ b/packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.ts
@@ -82,10 +82,7 @@ export function evaluate(condition: Condition, userAttributes: UserAttributes): 
   if (!conditionMatch) {
     evaluatorForMatch = exactEvaluator;
   } else {
-    evaluatorForMatch = EVALUATORS_BY_MATCH_TYPE[conditionMatch];
-  }
-  if (!evaluatorForMatch) {
-    return null;
+    evaluatorForMatch = EVALUATORS_BY_MATCH_TYPE[conditionMatch] || exactEvaluator;
   }
 
   return evaluatorForMatch(condition, userAttributes);
--- a/packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.ts
+++ b/packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.ts
@@ -13,20 +13,15 @@
  * See the License for the specific language governing permissions and      *
  * limitations under the License.                                           *
  ***************************************************************************/
-import { sprintf } from '@optimizely/js-sdk-utils';
 import { getLogger } from '@optimizely/js-sdk-logging';
-import { LoggerFacade } from '@optimizely/js-sdk-logging';
 import { UserAttributes } from '../../shared_types';
 
 import { isNumber, isSafeInteger } from '../../utils/fns';
-import {
-  LOG_LEVEL,
-  LOG_MESSAGES,
-} from '../../utils/enums';
+import { LOG_MESSAGES } from '../../utils/enums';
 
 const MODULE_NAME = 'CUSTOM_ATTRIBUTE_CONDITION_EVALUATOR';
 
-const logger = getLogger(MODULE_NAME)
+const logger = getLogger();
 
 const EXACT_MATCH_TYPE = 'exact';
 const EXISTS_MATCH_TYPE = 'exists';
@@ -49,7 +44,7 @@ type Condition = {
   value: string | number | boolean;
 }
 
-type ConditionEvaluator = (condition: Condition, userAttributes: UserAttributes, logger: LoggerFacade) => boolean | null;
+type ConditionEvaluator = (condition: Condition, userAttributes: UserAttributes) => boolean | null;
 
 const EVALUATORS_BY_MATCH_TYPE: { [conditionType: string]: ConditionEvaluator | undefined } = {};
 EVALUATORS_BY_MATCH_TYPE[EXACT_MATCH_TYPE] = exactEvaluator;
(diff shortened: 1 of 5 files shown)


packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.ts
```diff
@@ -15,29 +15,42 @@
...
+// TODO: Find place for UserAttributes and Condition types to live in
+type UserAttributes = {
+  // TODO[OASIS-6649]: Don't use object type
+  // eslint-disable-next-line  @typescript-eslint/no-explicit-any
+  [name: string]: any;
+};
+
+type Condition = {
+  [name: string]: string;
+};
```
mjc1283, line 35: Let's be more specific here. You can check out the structure of condition objects in the datafile by creating audiences with different condition types. 

Condition objects have `type`, `name`, `match`, and `value` properties. The functions in this module use `name`, `match`, and `value`.

packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.ts
```diff
@@ -70,37 +83,37 @@ export var evaluate = function(condition, userAttributes, logger) {
...
   return evaluatorForMatch(condition, userAttributes, logger);
-};
+}
 
 /**
  * Returns true if the value is valid for exact conditions. Valid values include
  * strings, booleans, and numbers that aren't NaN, -Infinity, or Infinity.
  * @param value
- * @returns {Boolean}
+ * @returns {?boolean}
```
mjc1283, line 94: Should this just be `boolean`, not `?boolean`?

packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.ts
```diff
@@ -15,29 +15,42 @@
...
+import { LoggerFacade } from '@optimizely/js-sdk-logging';
+
 import { isNumber, isSafeInteger } from '../../utils/fns';
 import {
   LOG_LEVEL,
   LOG_MESSAGES,
 } from '../../utils/enums';
 
-var MODULE_NAME = 'CUSTOM_ATTRIBUTE_CONDITION_EVALUATOR';
+// TODO: Find place for UserAttributes and Condition types to live in
```
mjc1283, line 26: We could create a shared types module at lib/shared_types.ts, and put this there.

It would be nice if somehow we could import `UserAttributes` from there and use it here, as well as also in lib/index.d.ts. I had some trouble figuring that out, it may not be possible. If so I think we can just duplicate the definition (have one in lib/index.d.ts, and one in lib/shared_types.ts which is imported by out TS modules).

yavorona: I also tried to  use `UserAttributes` in lib/index.d.ts from lib/shared_types.ts i, but could not figure it out how to make it work, so left lib/index.d.ts as is and followed your suggestion.

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -0,0 +1,14 @@
+export type UserAttributes = {
+   // TODO[OASIS-6649]: Don't use any type
+   // eslint-disable-next-line  @typescript-eslint/no-explicit-any
+   [name: string]: any;
+}
+
+export type Condition = {
```
mjc1283, line 7: `Condition` can live in the custom_attribute_condition_evaluator module

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -0,0 +1,14 @@
...
+   name: string;
+   type: string;
+   match: string;
```
mjc1283, line 10: The evaluator has the ability to handle when `match` is missing (in that case it is treated as an exact match), so let's capture that here:
```suggestion
   match?: string;
```

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -0,0 +1,14 @@
...
+   // TODO[OASIS-6649]: Don't use any type
+   // eslint-disable-next-line  @typescript-eslint/no-explicit-any
+   value: any;
```
mjc1283, line 13:
```suggestion
   value: string | number | boolean;
```

packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.ts
```diff
@@ -15,29 +15,32 @@
...
+const MATCH_TYPES = [
   EXACT_MATCH_TYPE,
   EXISTS_MATCH_TYPE,
   GREATER_THAN_MATCH_TYPE,
   LESS_THAN_MATCH_TYPE,
   SUBSTRING_MATCH_TYPE,
 ];
 
-var EVALUATORS_BY_MATCH_TYPE = {};
+const EVALUATORS_BY_MATCH_TYPE = {};
```
mjc1283, line 43: Let's add a type annotation on this object - otherwise it seems reading a property of it results in `any` (check the type of `evaluatorForMatch` on line 76). How about:
```suggestion
type ConditionEvaluator = (condition: Condition, userAttributes: UserAttributes) => boolean | null;
const EVALUATORS_BY_MATCH_TYPE: { [conditionType: string]: ConditionEvaluator | undefined } = {};
```

packages/optimizely-sdk/lib/core/custom_attribute_condition_evaluator/index.ts
```diff
@@ -47,21 +50,21 @@ EVALUATORS_BY_MATCH_TYPE[SUBSTRING_MATCH_TYPE] = substringEvaluator;
...
+ * @param  {Condition}        condition
+ * @param  {UserAttributes}   userAttributes
+ * @param  {LoggerFacade}     logger
+ * @return {?boolean}         true/false if the given user attributes match/don't match the given condition,
+ *                            null if the given user attributes and condition can't be evaluated
  * TODO: Change to accept and object with named properties
  */
-export var evaluate = function(condition, userAttributes, logger) {
-  var conditionMatch = condition.match;
+export function evaluate(condition: Condition, userAttributes: UserAttributes, logger: LoggerFacade): boolean | null {
```
mjc1283, line 60: This is a good opportunity to clean up logging in this module.

- Let's change all these functions so that they use a logger defined inside this module (like `const logger = getLogger(MODULE_NAME);`, and then remove the `logger` argument from the list of parameters.
- Call the logger method corresponding to the desired level, rather than passing in level (`logger.debug(...)` instead of `logger.log(LOG_LEVEL.DEBUG, ...)`

mjc1283 (changes requested):

packages/optimizely-sdk/lib/shared_types.ts
yavorona: @mjc1283 I noticed that when we set attribute to `has any value`, the `value` in that case becomes `null`, so I think we should capture it here as well: `value: string | number | boolean | null`. What do you think?

mjc1283 (changes requested):
Looks way better, thank you! Just one requested change then this should be ready to merge!

mjc1283 (approved):

PR #564, opened 2020-09-14
fix(datafile accessor): Add datafile accessor method to TS definitions

mjc1283:
## Summary
Add new datafile accessor method to TS type definitions

## Test plan

Manually tested with `npm link` and local app


mikeng13 (approved):
lgtm

PR #565, opened 2020-09-14
feat: Convert event_processor_config_validator to TS

yavorona:
## Summary
Convert the last utils module `event_processor_config_validator` from JS to TS.

## Test plan

Existing unit tests

### DIFF
--- a/packages/optimizely-sdk/lib/utils/event_processor_config_validator/index.ts
+++ b/packages/optimizely-sdk/lib/utils/event_processor_config_validator/index.ts
@@ -20,7 +20,7 @@ import { isSafeInteger } from '../fns';
  * @param {unknown}   eventBatchSize
  * @returns {boolean}
  */
-export function validateEventBatchSize(eventBatchSize: unknown): boolean {
+const validateEventBatchSize = function(eventBatchSize: unknown): boolean {
   if (typeof eventBatchSize === 'number' && isSafeInteger(eventBatchSize)) {
     return eventBatchSize >= 1;
   }
@@ -32,9 +32,14 @@ export function validateEventBatchSize(eventBatchSize: unknown): boolean {
  * @param {unknown}   eventFlushInterval
  * @returns {boolean}
  */
-export function validateEventFlushInterval(eventFlushInterval: unknown): boolean {
+const validateEventFlushInterval = function(eventFlushInterval: unknown): boolean {
   if (typeof eventFlushInterval === 'number' && isSafeInteger(eventFlushInterval)) {
     return eventFlushInterval > 0;
   }
   return false;
+}
+
+export default {
+  validateEventBatchSize: validateEventBatchSize,
+  validateEventFlushInterval: validateEventFlushInterval,
 }
--- a/packages/optimizely-sdk/lib/utils/event_processor_config_validator/index.ts
+++ b/packages/optimizely-sdk/lib/utils/event_processor_config_validator/index.ts
@@ -17,23 +17,24 @@ import { isSafeInteger } from '../fns';
 
 /**
  * Return true if the argument is a valid event batch size, false otherwise
- * @param {*} eventBatchSize
- * @returns boolean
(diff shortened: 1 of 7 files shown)


packages/optimizely-sdk/lib/utils/event_processor_config_validator/index.ts
```diff
@@ -17,23 +17,24 @@ import { isSafeInteger } from '../fns';
...
- * @param {*} eventBatchSize
- * @returns boolean
+ * @param {unknown}   eventBatchSize
+ * @returns {boolean}
  */
-export var validateEventBatchSize = function(eventBatchSize) {
-  return isSafeInteger(eventBatchSize) && eventBatchSize >= 1;
-};
+export function validateEventBatchSize(eventBatchSize: unknown): boolean {
+  if (typeof eventBatchSize === 'number' && isSafeInteger(eventBatchSize)) {
```
yavorona, line 24: Even though `isSaveInteger` would take care of checking the eventBatchSize type, we still need this additional type check here to avoid compiler error. Same condition applies on line 36 in `validateEventFlushInterval` method.

packages/optimizely-sdk/lib/utils/event_processor_config_validator/index.ts
```diff
@@ -32,9 +32,14 @@ export function validateEventBatchSize(eventBatchSize: unknown): boolean {
...
  */
-export function validateEventFlushInterval(eventFlushInterval: unknown): boolean {
+const validateEventFlushInterval = function(eventFlushInterval: unknown): boolean {
   if (typeof eventFlushInterval === 'number' && isSafeInteger(eventFlushInterval)) {
     return eventFlushInterval > 0;
   }
   return false;
 }
+
+export default {
```
yavorona, line 42: Went back to default export object to avoid issues with stubbing `validateEventFlushInterval` and `validateEventFlushInterval`.

packages/optimizely-sdk/lib/utils/event_processor_config_validator/index.ts
mikeng13: For these validations, we should rely on TS to validate the type. For example, wherever we take in the batchSize input, we should make the param type a number. That way we can guarantee that it is always a number. Then this method would only validate that the number is within a safe range.

mjc1283: @mikeng13 I agree in general, but batch size is passed in by the caller which could be plain JS. At some place (either here or at a higher level), I think such values must be represented as `unknown`, have the type checked at runtime, and then can convert to `number`.

mjc1283: @yavorona Would using [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) be helpful here, or in other validator functions? Maybe it would help with control flow and avoiding unhelpful compiler errors in the callers of these validators.

mjc1283 (approved):
LGTM. I added a suggestion about type guards. Perhaps that would be useful when entry points, or modules calling validators, become TS.

packages/optimizely-sdk/lib/utils/event_processor_config_validator/index.ts
mikeng13: Ah yeah that is a fair point.

yavorona: @mjc1283 isn't a `typeof eventBatchSize === 'number'` already a type guard? Another way I can think of is to use type predicate, which would look like this:
``` 
const validateEventBatchSize = function(eventBatchSize: unknown): boolean {
  if (isSafeInteger(eventBatchSize)) {
    return eventBatchSize as number >= 1;
  }
  return false;
}
```
Let me know what you think!

PR #568, opened 2020-09-15
feat: Create index.d.ts file for decision_service

yavorona:
## Summary
Create `index.d.ts` file for decision_service module

### DIFF
--- a/packages/optimizely-sdk/lib/core/decision_service/index.d.ts
+++ b/packages/optimizely-sdk/lib/core/decision_service/index.d.ts
@@ -14,8 +14,11 @@
  * limitations under the License.
  */
 
-declare module '@optimizely/optimizely-sdk/lib/core/decision_service' { 
+declare module '@optimizely/optimizely-sdk/lib/core/decision_service' {
   import { LogHandler } from '@optimizely/js-sdk-logging';
+  import { UserProfile, UserAttributes } from '@optimizely/optimizely-sdk';
+  import { ProjectConfig } from '@optimizely/optimizely-sdk/lib/core/project_config';
+  import { Experiment, Variation, FeatureFlag } from '@optimizely/optimizely-sdk/lib/core/project_config';
 
   export function createDecisionService(options: Options): DecisionService;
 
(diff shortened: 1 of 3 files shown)


packages/optimizely-sdk/lib/core/decision_service/index.d.ts
```diff
@@ -0,0 +1,151 @@
+/**
+ * Copyright 2018-2020, Optimizely
```
mikeng13, line 2: Should just be 2020

packages/optimizely-sdk/lib/core/decision_service/index.d.ts
```diff
@@ -0,0 +1,151 @@
...
+
+declare module '@optimizely/optimizely-sdk/lib/core/decision_service' { 
+
+  export interface DecisionService {
+    getVariation(experimentKey: string, userId: string, attributes?: UserAttributes): string | null;
+    // getVariationForFeature(configObj, feature, userId, attributes):
+  }
+
+  export interface ConfigObj {
+    __datafileStr: string;
```
mikeng13, line 25: Shouldn't document private properties

packages/optimizely-sdk/lib/core/decision_service/index.d.ts
```diff
@@ -0,0 +1,151 @@
...
+    rolloutIdMap: {[key:string]: Rollout};
+    experimentKeyMap: {[key: string]: Experiment} //look into this one closer
+    experimentIdMap: {[key: string]: Experiment};
+    variationIdMap: {[key: string]: Variation};
+    variationVariableUsageMap: {[key: string]: ExperimentVariableByID};
+    experimentFeatureMap: {[key: string]: string[]};
+    featureKeyMap: {[key: string]: FeatureFlag}; //look into this one closer
+  }
+
+  interface Event {
```
mikeng13, line 53: A lot of these entities are being defined in #567 already. We should reuse them.

packages/optimizely-sdk/lib/core/decision_service/index.d.ts
```diff
@@ -0,0 +1,183 @@
...
+    };
+  }
+
+  interface Options {
+    userProfileService: UserProfileService | null;
+    logger: LogHandler;
+    UNSTABLE_conditionEvaluators: unknown;
+  }
+
+  interface ConfigObj {
```
mjc1283, line 51: This should be imported from the project config module (see PR https://github.com/optimizely/javascript-sdk/pull/567)

packages/optimizely-sdk/lib/core/decision_service/index.d.ts
```diff
@@ -0,0 +1,183 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+declare module '@optimizely/optimizely-sdk/lib/core/decision_service' { 
+  import { LogHandler } from '@optimizely/js-sdk-logging';
+
+  export function createDecisionService(options: Options): DecisionService;
+
+  export interface UserProfileService {
```
mjc1283, line 22: Are `UserProfileService` and/or `UserProfile` needed in any other modules? There might be a more natural place for these to live.

packages/optimizely-sdk/lib/core/decision_service/index.d.ts
mjc1283: Agree, for all the entity types that are related to or located in project config.

packages/optimizely-sdk/lib/core/decision_service/index.d.ts
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  type Condition = {
+    name: string;
+    type: string;
+    match?: string;
+    value: string | number | boolean | null;
+  }
+  
+  type ConditionTree<Leaf> = Leaf | unknown[];
```
mjc1283, line 148: Should be imported from `condition_tree_evaluator`

packages/optimizely-sdk/lib/index.d.ts
```diff
@@ -135,11 +136,6 @@ declare module '@optimizely/optimizely-sdk' {
     dispatchEvent: (event: Event, callback: () => void) => void;
   }
 
-  export interface UserProfileService {
-    lookup: (userId: string) => UserProfile;
-    save: (profile: UserProfile) => void;
-  }
-
```
mjc1283, line 142: This, as well as `UserProfile`, still need to be exported out of this file. We can't break compatibility with what we previously were exporting.

mjc1283 (changes requested):
Requesting changes to locate type definitions in the appropriate module and import them into the consumers.

packages/optimizely-sdk/lib/core/decision_service/index.d.ts
yavorona: I moved them back based on the comment below.

mjc1283: > We are importing them from a root-level module to here, which is an anti-pattern.

Agreed. Let's keep them in decision service like you originally had, and consider moving them later on if it makes sense.

packages/optimizely-sdk/lib/core/project_config/index.d.ts
```diff
@@ -0,0 +1,168 @@
...
+    experiments: Experiment[]  
+    id: string
+    policy: string
+    trafficAllocation: Array<{
+      entityId: string;
+      endOfRange: number;
+    }>
+  } 
+
+type ConditionTree<Leaf> = Leaf | unknown[];
```
mjc1283, line 61: Import from `condition_tree_evaluator`

mjc1283 (changes requested):
LGTM but this depends on #567. If you want to merge it first, let's put more of a stub or skeleton in project_config/index.d.ts which can be filled in later by #567

packages/optimizely-sdk/lib/core/decision_service/index.d.ts
yavorona: Sounds good. `UserAttributes` would also need to be moved here for now to avoid anti-pattern.

mjc1283 (changes requested):
Still waiting for update after #567 is complete

packages/optimizely-sdk/lib/core/decision_service/index.d.ts
```diff
@@ -0,0 +1,107 @@
...
+  interface Options {
+    userProfileService: import('../../shared_types').UserProfileService | null;
+    logger: LogHandler;
+    UNSTABLE_conditionEvaluators: unknown;
+  }
+
+  interface Decision {
+    experiment: import('../../shared_types').Experiment | null;
+    variation: import('../../shared_types').Variation | null;
+    decisionSource: string;
```
mjc1283, line 105: We might want to use an `enum` for this. Can be updated later if so.

mjc1283 (approved):

PR #569, opened 2020-09-16
Event builder interface added

ozayr-zaviar:
## Summary
- index.d.ts added for event builder


packages/optimizely-sdk/lib/core/event_builder/index.d.ts
```diff
@@ -0,0 +1,83 @@
...
+            }>
+        }>
+        revision: string;
+        client_name: string;
+        client_version: string;
+        anonymize_ip:boolean;
+        enrich_decisions: boolean;
+    }
+
+    export interface EventBuilder {
```
msohailhussain, line 71: Only dispatchEvent should be here, rest not needed for now.

PR #576, opened 2020-09-25
feat: Convert optimizely module to TS

yavorona:
## Summary

- Convert `optimizely` module from JS to TS.

## Test plan
- Existing unit tests and local bundle


packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -89,19 +121,13 @@ export interface OptimizelyConfig {
...
-  // URL to which to send the HTTP request.
-  url: string;
-  // HTTP method with which to send the event.
-  httpVerb: 'POST';
-  // Value to send in the request body, JSON-serialized.
-  // TODO[OASIS-6649]: Don't use any type
-  // eslint-disable-next-line  @typescript-eslint/no-explicit-any
-  params: any;
+/**
+ * Temprorary placement of LogTierV1EventProcessorConfig
```
yavorona, line 125: I originally saved `LogTierV1EventProcessorConfig` in `event-processor/src/eventProcessor.ts`, which resulted in travis lint failure due to event-processor package not being set up https://travis-ci.org/github/optimizely/javascript-sdk/builds/731087456, so I moved it here for now

packages/optimizely-sdk/lib/core/project_config/index.js
```diff
@@ -247,14 +247,14 @@ export var getExperimentStatus = function(projectConfig, experimentKey) {
...
-  return this.getExperimentStatus(projectConfig, experimentKey) === EXPERIMENT_RUNNING_STATUS;
+  return getExperimentStatus(projectConfig, experimentKey) === EXPERIMENT_RUNNING_STATUS;
 };
 
 /**
  * Determine for given experiment if event is running, which determines whether should be dispatched or not
  */
 export var isRunning = function(projectConfig, experimentKey) {
-  return this.getExperimentStatus(projectConfig, experimentKey) === EXPERIMENT_RUNNING_STATUS;
+  return getExperimentStatus(projectConfig, experimentKey) === EXPERIMENT_RUNNING_STATUS;
```
yavorona, line 257: existed previously `this....`  resulted in browser test failures. I do not think we needed it here in the first place.
LMKWYT @mjc1283

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+  private __nextReadyTimeoutId: number;
+  private clientEngine: string;
+  private clientVersion: string;
+  private errorHandler: ErrorHandler;
+  private eventDispatcher: EventDispatcher;
+  private logger: LogHandler;
+  private projectConfigManager: ProjectConfigManager;
+  private notificationCenter: NotificationCenter;
+  private decisionService: DecisionService;
+  private eventProcessor: eventProcessor.EventProcessor;
```
mjc1283, line 80: It is weird that we don't consistently follow this double-underscore naming convention. Let's take this opportunity to clean it up. Since we have the `private` keyword, we don't need the naming convention anymore to hint that they're private.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+
+  constructor(config: projectConfig.ProjectConfig) {
```
mjc1283, line 82: I think `ProjectConfig` and the config object here are two separate things. Let's have a separate interface for the values accepted by the `Optimizely` constructor.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+
+    this.projectConfigManager = createProjectConfigManager({
+      datafile: config.datafile,
+      datafileOptions: config.datafileOptions,
+      jsonSchemaValidator: config.jsonSchemaValidator,
+      sdkKey: config.sdkKey,
+    });
+
+    this.__disposeOnUpdate = this.projectConfigManager.onUpdate(
+      function(this: Optimizely, configObj: projectConfig.ProjectConfig) {
```
mjc1283, line 107: I think we can use an arrow function instead of `bind`

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+  }
+
+  /**
+   * Returns a truthy value if this instance currently has a valid project config
+   * object, and the initial configuration object that was passed into the
+   * constructor was also valid.
+   * @return {boolean}
+   */
+  __isValidInstance(): boolean {
+  return this.__isOptimizelyConfigValid && !!this.projectConfigManager.getConfig();
```
mjc1283, line 167: Indentation

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+      experimentKey: experimentKey,
+      variationKey: variationKey,
+      userId: userId,
+      userAttributes: attributes,
+      clientEngine: this.clientEngine,
+      clientVersion: this.clientVersion,
+      configObj: configObj,
+    });
+    // TODO is it okay to not pass a projectConfig as second argument
+    this.eventProcessor.process(impressionEvent as eventProcessor.ProcessableEvent);
```
mjc1283, line 257: We should define in our `ImpressionEvent` interface what is returned from `buildImpressionEvent`. Then the type checker will let us know if that's incompatible with the `ProcessableEvent` expected by the event processor.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+      errorHandler: this.errorHandler,
+    });
+
+    this.eventProcessor = new eventProcessor.LogTierV1EventProcessor({
+      dispatcher: this.eventDispatcher,
+      flushInterval: config.eventFlushInterval,
+      batchSize: config.eventBatchSize,
+      maxQueueSize: config.eventMaxQueueSize, // TODO: update event-processor to include maxQueueSize
+      notificationCenter: this.notificationCenter,
+    } as LogTierV1EventProcessorConfig);
```
mjc1283, line 147: The `as LogTierV1EventProcessorConfig` could be covering up some errors. I recommend removing it.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+        userId: userId,
+        userAttributes: attributes,
+        clientEngine: this.clientEngine,
+        clientVersion: this.clientVersion,
+        configObj: configObj,
+      });
+      this.logger.log(LOG_LEVEL.INFO, sprintf(enums.LOG_MESSAGES.TRACK_EVENT, MODULE_NAME, eventKey, userId));
+      // TODO is it okay to not pass a projectConfig as second argument
+      this.eventProcessor.process(conversionEvent as eventProcessor.ProcessableEvent);
+      this.__emitNotificationCenterTrack(eventKey, userId, attributes as UserAttributes, eventTags);
```
mjc1283, line 347: Similar to above comments, let's not use type casting with `as`.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+        return enabledFeatures;
+      }
+
+      const configObj = this.projectConfigManager.getConfig();
+      if (!configObj) {
+        return enabledFeatures;
+      }
+      if (configObj.featureKeyMap) {
+        objectValues(configObj.featureKeyMap).forEach(
+          function(this: Optimizely, feature: FeatureFlag): void {
```
mjc1283, line 701: Recommend arrow function

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+          }
+
+          delete stringInputs['user_id'];
+        }
+        const inputKeys = Object.keys(stringInputs);
+        for (let index = 0; index < inputKeys.length; index++) {
+          const key = inputKeys[index];
+          if (!stringValidator.validate(stringInputs[key])) {
+            throw new Error(sprintf(ERROR_MESSAGES.INVALID_INPUT_FORMAT, MODULE_NAME, key));
+          }
```
mjc1283, line 536: Recommend changing to `forEach` instead of loop with index

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+      const featureFlag = projectConfig.getFeatureFromKey(configObj, featureKey, this.logger);
+      if (!featureFlag) {
+        return null;
+      }
+
+      const decision = this.decisionService.getVariationForFeature(configObj, featureFlag, userId, attributes);
+      const featureEnabled = decision.variation !== null ? decision.variation.featureEnabled : false;
+      const allVariables = {};
+
+      featureFlag.variables.forEach(function (this: Optimizely, variable: FeatureVariable) {
```
mjc1283, line 1124: Recommend arrow function.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+      const eventProcessorStoppedPromise = this.eventProcessor.stop();
+      if (this.__disposeOnUpdate) {
+        this.__disposeOnUpdate();
+        this.__disposeOnUpdate = null;
+      }
+      if (this.projectConfigManager) {
+        this.projectConfigManager.stop();
+      }
+      Object.keys(this.__readyTimeouts).forEach(
+        function(this: Optimizely, readyTimeoutId: string ) {
```
mjc1283, line 1251: Recommend arrow function

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+
+    let resolveTimeoutPromise: (value?: unknown) => void;
+    const timeoutPromise = new Promise(function(resolve: (value?: unknown) => void) {
+      resolveTimeoutPromise = resolve;
+    });
+
+    const timeoutId = this.__nextReadyTimeoutId;
+    this.__nextReadyTimeoutId++;
+
+    const onReadyTimeout = function(this: Optimizely) {
```
mjc1283, line 1327: Recommend arrow function

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+      function(this: Optimizely) {
+        clearTimeout(readyTimeout);
+        delete this.__readyTimeouts[timeoutId];
+        resolveTimeoutPromise({
+          success: true,
+        });
+      }.bind(this)
+    );
+
+    return Promise.race([this.__readyPromise, timeoutPromise]) as Promise<{ success: boolean; reason?: string | undefined; }>;
```
mjc1283, line 1357: Recommend removing `as` type cast

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+        this.logger.log(
+          LOG_LEVEL.WARNING,
+          sprintf(enums.LOG_MESSAGES.EVENT_KEY_NOT_FOUND, MODULE_NAME, eventKey)
+        );
+        this.logger.log(LOG_LEVEL.WARNING, sprintf(LOG_MESSAGES.NOT_TRACKING_USER, MODULE_NAME, userId));
+        return;
+      }
+
+      // remove null values from eventTags
+      eventTags = this.__filterEmptyValues(eventTags as EventTags);
```
mjc1283, line 334: Recommend removing `as` type cast

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1359 @@
...
+   * evaluated the value of the variable.
+   *
+   * @param  {string}          featureKey           Key of the feature whose variable's value is
+   *                                                being accessed
+   * @param  {boolean}         featureEnabled       Boolean indicating if feature is enabled or not
+   * @param  {Variation}       variation            variation returned by decision service
+   * @param  {FeatureVariable} variable             varible whose value is being evaluated
+   * @param  {string}          userId               ID for the user
+   * @return {string|null}                          String value of the variable or null if the
+   *                                                config Obj is null
```
mjc1283, line 852: Interesting - looks like at some point, this function was changed, but they didn't update the comment. It is not returning `string | null`, it is actually doing the type cast (calling `projectConfig.getTypeCastValue`). Probably the right return type is `unknown`.

packages/optimizely-sdk/lib/core/project_config/index.d.ts
```diff
@@ -13,97 +13,143 @@
...
+ *                                            FEATURE_VARIABLE_TYPES in
+ *                                            lib/utils/enums/index.js. The return value's
+ *                                            type is determined by this argument (boolean
+ *                                            for BOOLEAN, number for INTEGER or DOUBLE,
+ *                                            and string for STRING).
+ * @param   {LogHandler}    logger            Logger instance
+ * @returns {T}                               Variable value of the appropriate type, or
+ *                                            null if the type cast failed
+ */
+export function getTypeCastValue<T>(variableValue: string, type: string, logger: LogHandler): T;
```
mjc1283, line 136: Looking a little closer at `getTypeCastValue`, I think this could be improved (not necessary to do in this PR):

- The second argument`type` can be defined as an `enum`.
- Not sure the generic `T` is helpful here. I could see it helping if we wanted to restrict it to the feature variable types like `T extends string | number | boolean`, but with JSON variables, we're calling `JSON.parse`, the return type of which is broader. I would go with `unknown` instead of the generic `T`.

mjc1283 (changes requested):
- I recommended some small refactors to modernize the code
- `shared_types` is getting pretty big, which could be a sign of poor organization. Are there any types in there that should live in the modules that they're used in?
- The `as` type casts on the return value in the feature variable getter methods - there might be a way to make this more elegant using user-defined type guard functions. But no need to spend a lot of time on that rabbit hole right now.

packages/optimizely-sdk/lib/shared_types.ts
yavorona: `LogTierV1EventProcessorConfig` is the interface that includes `maxQueueSize`, which is not included in `LogTierV1EventProcessor` that is being used in this case. I used `{ ... } as LogTierV1EventProcessorConfig` as a temporary solution until we implement it in the event-processor.

packages/optimizely-sdk/lib/core/project_config/entities.ts
```diff
@@ -16,9 +16,20 @@
...
 }
 
 export interface FeatureFlag {
-  variables: FeatureVariable[];
+  rolloutId: string;
+  key: string;
+  id: string;
+  experimentIds: string[],
+  variables: FeatureVariable[],
+  variableKeyMap?: {[key: string]: FeatureVariable}
```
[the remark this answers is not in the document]
yavorona: you are right, this should not be optional.

packages/optimizely-sdk/lib/core/project_config/index.d.ts
```diff
@@ -13,97 +13,143 @@
...
+  jsonSchemaValidator?: object;
+  sdkKey?: string;
+  userProfileService?: UserProfileService | null;
+  UNSTABLE_conditionEvaluators?: unknown;
+  eventFlushInterval?: number;
+  eventBatchSize?: number;
+  datafileOptions?: DatafileOptions;
+  eventMaxQueueSize?: number;
+  logger: LogHandler;
+  experimentKeyMap:{[key: string]: Experiment};
```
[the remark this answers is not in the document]
yavorona: in the Optimizely class constructor, the logger is being used from the config object as `this.logger = config.logger`, so I think we would need to have it included here. Same applies to attributes such as clientEngine, clientVersion, errorHandler, eventDispatcher, isValidInstance, datafile, jsonSchemaValidator, sdkKey, and other ones. Excluding any of these properties will result in the compiler error.  @mjc1283 Please LMK if you think there is a better way to go about it.

packages/optimizely-sdk/lib/core/project_config/project_config_manager.js
```diff
@@ -282,6 +282,10 @@ ProjectConfigManager.prototype.stop = function() {
   this.__updateListeners = [];
 };
 
+export var createProjectConfigManager = function(config) {
+  return new ProjectConfigManager(config);
+}
+
 export default {
-  ProjectConfigManager: ProjectConfigManager,
+  createProjectConfigManager: createProjectConfigManager,
```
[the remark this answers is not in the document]
yavorona: Yes, and also to distinguish between `ProjectConfigManager` interface and the method that returns `ProjectConfigManager` similarly to how we defined notification_center.

packages/optimizely-sdk/lib/optimizely/index.ts
yavorona: `LogTierV1EventProcessorConfig` is the interface that includes `maxQueueSize`, which is not included in `LogTierV1EventProcessor` that is being used in this case. I used `{ ... } as LogTierV1EventProcessorConfig` as a temporary solution until we implement it in the event-processor.

packages/optimizely-sdk/lib/optimizely/index.ts
yavorona: I removed it by updating the signature of the  `__filterEmptyValues` to 
```
__filterEmptyValues(map: EventTags | undefined): EventTags | undefined { ... }
```

packages/optimizely-sdk/lib/optimizely/index.ts
yavorona: `this.__emitNotificationCenterTrack` in this case should have been accepting attributes and eventTags as optional parameters. I updated its signature and removed `as` type casting here.

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -35,11 +14,18 @@ export interface UserProfile {
   };
 }
 
+export type EventTags = {
+  [key: string]: string | number | null;
```
yavorona, line 18: I changed `EventTags` type to match the event-processor, which did not cause any issues in optimizely-sdk. However, I still think we should include `boolean` as an acceptable value. LMKWYT @mjc1283

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -23,6 +30,7 @@ export interface Experiment {
   // TODO[OASIS-6649]: Don't use object type
   // eslint-disable-next-line  @typescript-eslint/ban-types
   forcedVariations: object;
+  variationKeyMap?: {[key: string]: Variation}
```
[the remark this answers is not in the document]
yavorona: No, it should not be.

yavorona:
> * I recommended some small refactors to modernize the code
> * `shared_types` is getting pretty big, which could be a sign of poor organization. Are there any types in there that should live in the modules that they're used in?
> * The `as` type casts on the return value in the feature variable getter methods - there might be a way to make this more elegant using user-defined type guard functions. But no need to spend a lot of time on that rabbit hole right now.

The biggest part of the `shared_files` contains Optimizely Config Entities, which originally lived in our published index.d.ts file. 
If I keep it there, we would not be able to access it in our optimizely module where is it being used. Same happens if I move Optimizely Config Entities to project_config_manager, where it is also being used. @mjc1283  LMK if you think there is a better place for them to live.
 I haven’t found a good way to avoid using `as` type casts in our feature variable getters, but I will spend more time to think about it!

packages/optimizely-sdk/lib/optimizely/index.ts
yavorona: It is now removed. The compiler error from missing `maxQueueSize` in `LogTierV1EventProcessor` constructor was avoided by assigning the object to another variable before passing it to `LogTierV1EventProcessor`.

packages/optimizely-sdk/lib/index.d.ts
```diff
@@ -160,6 +161,10 @@ declare module '@optimizely/optimizely-sdk' {
     dispatchEvent: (event: Event, callback: () => void) => void;
   }
 
+  export type EventTags = {
```
yavorona, line 164: Per our discussion, `EventTags` should not accept boolean as a value

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+  /**
+   * Validate string inputs, user attributes and event tags.
+   * @param  {unknown}  stringInputs   Map of string keys and associated values
+   * @param  {unknown}  userAttributes Optional parameter for user's attributes
+   * @param  {unknown}  eventTags      Optional parameter for event tags
+   * @return {boolean}                 True if inputs are valid
+   *
+   */
+  private validateInputs(
+    stringInputs?: unknown,
```
mjc1283, line 543: Looking at the callers, it appears we always pass an object. This will simplify the implementation:

```suggestion
    // Later we could make these camelCase, snake_case is weird
    stringInputs: Record<'feature_key' | 'user_id' | 'variable_key' | 'experiment_key', unknown>
```

mjc1283 (changes requested):
Looks really good!

packages/optimizely-sdk/lib/optimizely/index.ts
yavorona: Made some modification to avoid missing properties compiler error and added `event_key`:
`Partial<Record<'feature_key' | 'user_id' | 'variable_key' | 'experiment_key' | 'event_key', unknown>>`

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+ * Unless required by applicable law or agreed to in writing, software      *
+ * distributed under the License is distributed on an "AS IS" BASIS,        *
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+ * See the License for the specific language governing permissions and      *
+ * limitations under the License.                                           *
+ ***************************************************************************/
+import { sprintf, objectValues } from '@optimizely/js-sdk-utils';
+import { LogHandler, ErrorHandler } from '@optimizely/js-sdk-logging';
+import { FeatureFlag, FeatureVariable } from '../core/project_config/entities';
+import { EventDispatcher } from '@optimizely/js-sdk-event-processor';
```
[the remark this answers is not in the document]
yavorona: Found incompatibility in our EventDispatcher interface with the one defined in event processor. 
The callback in `dispatchEvent` method is updated from `callback: () => void` to `callback: (response: { statusCode: number; }) => void`. I noted this change in the log.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+   */
+  private validateInputs(
+    // TODO: Make feature_key, user_id, variable_key, experiment_key camelCase
+    stringInputs: Partial<Record<'feature_key' | 'user_id' | 'variable_key' | 'experiment_key' | 'event_key', unknown>>,
+    userAttributes?: unknown,
+    eventTags?: unknown
+  ): boolean {
+    try {
+      // Null, undefined or non-string user Id is invalid.
+      if (typeof stringInputs === 'object' && stringInputs !== null) {
```
mjc1283, line 549: I think this check is not needed anymore - `stringInputs` has to be a non-null object.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -0,0 +1,1383 @@
...
+      }
+    }
+    if (!isSafeInteger(timeoutValue)) {
+      timeoutValue = DEFAULT_ONREADY_TIMEOUT;
+    }
+
+    let resolveTimeoutPromise: (value: OnReadyResult) => void;
+    const timeoutPromise = new Promise<OnReadyResult>(
+      (resolve) => {
+      resolveTimeoutPromise = resolve;
```
mjc1283, line 1346: Missing indentation

mjc1283 (approved):
Great work!

packages/optimizely-sdk/lib/core/notification_center/index.d.ts
```diff
@@ -14,79 +14,78 @@
...
-    FEATURE_VARIABLE = 'feature-variable',
-    ALL_FEATURE_VARIABLES = 'all-feature-variables',
-  }
+export enum NOTIFICATION_TYPES {
+  ACTIVATE = 'ACTIVATE:experiment, user_id,attributes, variation, event',
+  DECISION = 'DECISION:type, userId, attributes, decisionInfo',
+  LOG_EVENT = 'LOG_EVENT:logEvent',
+  OPTIMIZELY_CONFIG_UPDATE = 'OPTIMIZELY_CONFIG_UPDATE',
+  TRACK = 'TRACK:event_key, user_id, attributes, event_tags, event',
+}
```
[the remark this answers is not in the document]
yavorona: Works great!

PR #589, opened 2020-10-05
fix(project config manager): Don't log an error when not initialized with datafile

mjc1283:
## Summary
Fixes a regression introduced during refactoring for datafile accessor - any time the `datafile` property of the object passed to `createInstance` is missing or falsy, the logger will output (for example):

```
[OPTIMIZELY] - ERROR 2020-10-05T23:38:15.224Z CONFIG_VALIDATOR: No datafile specified. Cannot start optimizely
```

Passing in an `sdkKey`, but no `datafile`, is valid and supported when calling `createInstance`, but this message leads the user to believe otherwise. In this situation, nothing is actually wrong, but the message says "Cannot start optimizely".

To fix, in `ProjectConfigManager` `__initialize`, when `config.datafile` is falsy, we do not call `__handleNewDatafile`, avoiding the code path that eventually would log the error message.

## Test plan

- Added new unit test
- Manually tested


aliabbasrizvi (approved):
LGTM

PR #590, opened 2020-10-06
converted entry level modules to ts

ozayr-zaviar:
## Summary
- Converted `packages\optimizely-sdk\lib\index.node.js` to TS
- Converted `packages\optimizely-sdk\lib\index.browser.js` to TS
- Converted `packages\optimizely-sdk\lib\index.react_native.js` to TS

## Test plan
- All existing unit tests should pass
- All existing test cases on FSC should pass

### DIFF
--- a/packages/optimizely-sdk/lib/index.browser.ts
+++ b/packages/optimizely-sdk/lib/index.browser.ts
@@ -19,11 +19,8 @@ import {
   setLogLevel,
   setErrorHandler,
   getErrorHandler,
-  LogLevel,
-  ErrorHandler,
-  LogHandler,
+  LogLevel
 } from '@optimizely/js-sdk-logging';
-
 import { LocalStoragePendingEventsDispatcher } from '@optimizely/js-sdk-event-processor';
 import { assign } from './utils/fns';
 import * as configValidator from './utils/config_validator';
@@ -33,7 +30,7 @@ import * as enums from './utils/enums';
 import loggerPlugin from './plugins/logger';
 import Optimizely from './optimizely';
 import eventProcessorConfigValidator from './utils/event_processor_config_validator';
-import { EventDispatcher, UserProfileService, OptimizelyConfig } from './shared_types';
+import { OptimizelyConfig, Config } from './shared_types';
 
 const logger = getLogger();
 setLogHandler(loggerPlugin.createLogger());
@@ -45,25 +42,13 @@ const DEFAULT_EVENT_FLUSH_INTERVAL = 1000; // Unit is ms, default is 1s
 
 let hasRetriedEvents = false;
 
-interface Config {
-  datafile?: string;
-  errorHandler?: ErrorHandler;
-  eventDispatcher?: EventDispatcher;
-  logger?: LogHandler;
-  logLevel?: LogLevel;
-  userProfileService?: UserProfileService;
-  eventBatchSize?: number;
-  eventFlushInterval?: number;
-  sdkKey?: string;
-  isValidInstance?: boolean;
-}
-
 /**
  * Creates an instance of the Optimizely class
  * @param  {Config} config
  * @return {Optimizely} the Optimizely object
  */
 const createInstance = function (config: Config): Optimizely | null {
+const createInstance = function(config: Config): Optimizely | null {
   try {
     config = config || {};
 
--- a/packages/optimizely-sdk/lib/index.browser.ts
+++ b/packages/optimizely-sdk/lib/index.browser.ts
@@ -30,7 +30,7 @@ import * as enums from './utils/enums';
 import loggerPlugin from './plugins/logger';
 import Optimizely from './optimizely';
 import eventProcessorConfigValidator from './utils/event_processor_config_validator';
-import { OptimizelyConfig, Config } from './shared_types';
+import { configObj, Config } from './shared_types';
 
 const logger = getLogger();
 setLogHandler(loggerPlugin.createLogger());
@@ -88,7 +88,7 @@ const createInstance = function(config: Config): Optimizely | null {
       eventDispatcher = config.eventDispatcher;
     }
 
-    const optimizelyConfig = assign(
+    const internalConfig = assign(
       {
         clientEngine: enums.JAVASCRIPT_CLIENT_ENGINE,
         eventBatchSize: DEFAULT_EVENT_BATCH_SIZE,
@@ -101,11 +101,11 @@ const createInstance = function(config: Config): Optimizely | null {
         logger: logger,
         errorHandler: getErrorHandler(),
       }
-    ) as OptimizelyConfig;
+    ) as configObj;
 
     if (!eventProcessorConfigValidator.validateEventBatchSize(config.eventBatchSize)) {
       logger.warn('Invalid eventBatchSize %s, defaulting to %s', config.eventBatchSize, DEFAULT_EVENT_BATCH_SIZE);
-      optimizelyConfig.eventBatchSize = DEFAULT_EVENT_BATCH_SIZE;
+      internalConfig.eventBatchSize = DEFAULT_EVENT_BATCH_SIZE;
     }
     if (!eventProcessorConfigValidator.validateEventFlushInterval(config.eventFlushInterval)) {
       logger.warn(
@@ -113,10 +113,10 @@ const createInstance = function(config: Config): Optimizely | null {
         config.eventFlushInterval,
         DEFAULT_EVENT_FLUSH_INTERVAL
       );
-      optimizelyConfig.eventFlushInterval = DEFAULT_EVENT_FLUSH_INTERVAL;
+      internalConfig.eventFlushInterval = DEFAULT_EVENT_FLUSH_INTERVAL;
     }
 
-    const optimizely = new Optimizely(optimizelyConfig);
+    const optimizely = new Optimizely(internalConfig);
 
     try {
       if (typeof window.addEventListener === 'function') {
--- a/packages/optimizely-sdk/lib/index.browser.ts
+++ b/packages/optimizely-sdk/lib/index.browser.ts
@@ -87,23 +87,12 @@ const createInstance = function(config: SDKOptions): Optimizely | null {
       eventDispatcher = config.eventDispatcher;
     }
 
-    const additionalEntities = {
-      clientEngine: enums.JAVASCRIPT_CLIENT_ENGINE,
-      eventBatchSize: DEFAULT_EVENT_BATCH_SIZE,
-      eventFlushInterval: DEFAULT_EVENT_FLUSH_INTERVAL,
-      eventDispatcher: eventDispatcher
-    };
-
-    const optimizelyLoggers = {
-      logger: logger,
-      errorHandler: getErrorHandler(),
-    };
-
-    const modifiedConfig = {...additionalEntities, ...config, ...optimizelyLoggers};
+    let eventBatchSize = config.eventBatchSize;
+    let eventFlushInterval = config.eventFlushInterval;
 
     if (!eventProcessorConfigValidator.validateEventBatchSize(config.eventBatchSize)) {
       logger.warn('Invalid eventBatchSize %s, defaulting to %s', config.eventBatchSize, DEFAULT_EVENT_BATCH_SIZE);
-      modifiedConfig.eventBatchSize = DEFAULT_EVENT_BATCH_SIZE;
+      eventBatchSize = DEFAULT_EVENT_BATCH_SIZE;
     }
     if (!eventProcessorConfigValidator.validateEventFlushInterval(config.eventFlushInterval)) {
       logger.warn(
@@ -111,19 +100,27 @@ const createInstance = function(config: SDKOptions): Optimizely | null {
         config.eventFlushInterval,
         DEFAULT_EVENT_FLUSH_INTERVAL
       );
-      modifiedConfig.eventFlushInterval = DEFAULT_EVENT_FLUSH_INTERVAL;
+      eventFlushInterval = DEFAULT_EVENT_FLUSH_INTERVAL;
     }
 
-    const optimizely = new Optimizely(modifiedConfig);
+    const optimizelyOptions = {
+      clientEngine: enums.JAVASCRIPT_CLIENT_ENGINE,
+      eventDispatcher: eventDispatcher,
+      ...config,
+      eventBatchSize: eventBatchSize,
+      eventFlushInterval: eventFlushInterval,
+      logger: logger,
+      errorHandler: getErrorHandler()
+    };
+
+    const optimizely = new Optimizely(optimizelyOptions);
 
     try {
       if (typeof window.addEventListener === 'function') {
         const unloadEvent = 'onpagehide' in window ? 'pagehide' : 'unload';
         window.addEventListener(
           unloadEvent,
-          () => {
-            optimizely.close();
-          },
+          optimizely.close,
           false
         );
       }
@@ -160,9 +157,9 @@ export default {
   logging: loggerPlugin,
   errorHandler: defaultErrorHandler,
   eventDispatcher: defaultEventDispatcher,
-  enums: enums,
+  enums,
   setLogger: setLogHandler,
-  setLogLevel: setLogLevel,
-  createInstance: createInstance,
-  __internalResetRetryState: __internalResetRetryState,
+  setLogLevel,
+  createInstance,
+  __internalResetRetryState,
 };
--- a/packages/optimizely-sdk/lib/index.browser.ts
+++ b/packages/optimizely-sdk/lib/index.browser.ts
@@ -22,7 +22,6 @@ import {
   LogLevel
 } from '@optimizely/js-sdk-logging';
 import { LocalStoragePendingEventsDispatcher } from '@optimizely/js-sdk-event-processor';
-import { assign } from './utils/fns';
 import configValidator from './utils/config_validator';
 import defaultErrorHandler from './plugins/error_handler';
 import defaultEventDispatcher from './plugins/event_dispatcher/index.browser';
@@ -30,7 +29,7 @@ import * as enums from './utils/enums';
 import loggerPlugin from './plugins/logger';
 import Optimizely from './optimizely';
 import eventProcessorConfigValidator from './utils/event_processor_config_validator';
-import { ConfigObj, Config } from './shared_types';
+import { OptimizelyOptions, SDKOptions } from './shared_types';
 
 const logger = getLogger();
 setLogHandler(loggerPlugin.createLogger());
@@ -44,10 +43,11 @@ let hasRetriedEvents = false;
 
 /**
  * Creates an instance of the Optimizely class
- * @param  {Config} config
- * @return {Optimizely} the Optimizely object
+ * @param  {SDKOptions} config
+ * @return {Optimizely|null} the Optimizely object
+ *                           null on error 
  */
-const createInstance = function(config: Config): Optimizely | null {
+const createInstance = function(config: SDKOptions): Optimizely | null {
   try {
     config = config || {};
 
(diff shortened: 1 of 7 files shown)


packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -13,16 +13,18 @@
...
   setLogLevel,
   setErrorHandler,
   getErrorHandler,
   LogLevel,
+  ErrorHandler,
+  LogHandler,
 } from '@optimizely/js-sdk-logging';
-import { LocalStoragePendingEventsDispatcher } from '@optimizely/js-sdk-event-processor';
 
+import { LocalStoragePendingEventsDispatcher } from '@optimizely/js-sdk-event-processor';
```
msohailhussain, line 27: why left blank line?

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -31,32 +33,37 @@ import * as enums from './utils/enums';
...
-var DEFAULT_EVENT_FLUSH_INTERVAL = 1000; // Unit is ms, default is 1s
-
-var hasRetriedEvents = false;
+const MODULE_NAME = 'INDEX_BROWSER';
+const DEFAULT_EVENT_BATCH_SIZE = 10;
+const DEFAULT_EVENT_FLUSH_INTERVAL = 1000; // Unit is ms, default is 1s
+
+let hasRetriedEvents = false;
+
+interface Config {
```
msohailhussain, line 48: document it. we should export as well? not sure.

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -97,7 +104,7 @@ var createInstance = function(config) {
       eventDispatcher = config.eventDispatcher;
     }
 
-    config = assign(
+    const optimizelyConfig = assign(
```
msohailhussain, line 107: why changed to `optimizelyConfig`. `OptimizelyConfig` is different than what is expected here.

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -110,29 +117,29 @@ var createInstance = function(config) {
         logger: logger,
         errorHandler: getErrorHandler(),
       }
-    );
+    ) as OptimizelyConfig;
```
msohailhussain, line 120: this is not correct.

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -31,32 +33,37 @@ import * as enums from './utils/enums';
...
- * @param  {Object}         config.userProfileService
- * @param  {Object}         config.eventBatchSize
- * @param  {Object}         config.eventFlushInterval
- * @param  {string}         config.sdkKey
- * @return {Object}         the Optimizely object
+ * @param  {Config} config
+ * @return {Optimizely} the Optimizely object
  */
-var createInstance = function(config) {
+const createInstance = function (config: Config): Optimizely | null {
```
msohailhussain, line 66: `function(config: Config)` should be without space.

packages/optimizely-sdk/lib/index.react_native.ts
```diff
@@ -13,47 +13,54 @@
...
 setLogLevel(LogLevel.INFO);
 
-var DEFAULT_EVENT_BATCH_SIZE = 10;
-var DEFAULT_EVENT_FLUSH_INTERVAL = 1000; // Unit is ms, default is 1s
-var DEFAULT_EVENT_MAX_QUEUE_SIZE = 10000;
+const DEFAULT_EVENT_BATCH_SIZE = 10;
+const DEFAULT_EVENT_FLUSH_INTERVAL = 1000; // Unit is ms, default is 1s
+const DEFAULT_EVENT_MAX_QUEUE_SIZE = 10000;
+
+interface Config {
```
msohailhussain, line 45: this is being used in all index level files, we should move it to shared_types.

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -77,6 +79,26 @@ export interface OptimizelyVariable {
   value: string;
 }
 
+export interface OptimizelyConfig {
```
msohailhussain, line 82: Isn't it defined [here](https://github.com/optimizely/javascript-sdk/blob/master/packages/optimizely-sdk/lib/optimizely/index.ts#L56). 
Please use the right term. `[OptimizelyConfig](https://github.com/optimizely/javascript-sdk/blob/master/packages/optimizely-sdk/lib/core/optimizely_config/index.js)` is different.

msohailhussain (changes requested):
Please address and fix all issues.

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -149,7 +140,7 @@ var createInstance = function(config) {
   }
 };
 
-var __internalResetRetryState = function() {
+const __internalResetRetryState = function (): void {
```
msohailhussain, line 143: we have default indentation without `space` please fix at all places.

packages/optimizely-sdk/lib/index.react_native.ts
```diff
@@ -13,47 +13,38 @@
...
   getErrorHandler,
-  LogLevel,
+  LogLevel
 } from '@optimizely/js-sdk-logging';
-
 import * as enums from './utils/enums';
 import { assign } from './utils/fns';
 import Optimizely from './optimizely';
-import configValidator from './utils/config_validator';
+import * as configValidator from './utils/config_validator';
```
msohailhussain, line 27: why not modified in other files?

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -77,6 +79,39 @@ export interface OptimizelyVariable {
   value: string;
 }
 
+export interface Config {
```
msohailhussain, line 82: add documentation for both `Config` and `ConfigObj`

msohailhussain (changes requested):
please address.

packages/optimizely-sdk/lib/index.browser.ts
zashraf1985: Its probably a line break between module imports and internal imports

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -97,7 +88,7 @@ var createInstance = function(config) {
       eventDispatcher = config.eventDispatcher;
     }
 
-    config = assign(
+    const internalConfig = assign(
```
zashraf1985, line 91: Since you  are now in TS, you dont need to use assign. you can probably use spread syntax to override values.

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -110,29 +101,29 @@ var createInstance = function(config) {
         logger: logger,
         errorHandler: getErrorHandler(),
       }
-    );
+    ) as ConfigObj;
```
zashraf1985, line 104: This conversions `as ConfigObj` might not be needed. Are you sure its throwing error without this.?

packages/optimizely-sdk/lib/index.node.ts
```diff
@@ -83,7 +73,7 @@ var createInstance = function(config) {
       config.isValidInstance = false;
     }
 
-    config = assign(
+    const internalConfig = assign(
```
zashraf1985, line 76: You can use spread operator here too instead of assign. Also rename internalConfig to something better

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -77,6 +79,45 @@ export interface OptimizelyVariable {
...
+  eventBatchSize?: number;
+  eventFlushInterval?: number;
+  sdkKey?: string;
+  isValidInstance?: boolean;
+}
+
+/**
+ * Config required to create optimizely object
+ */
+export interface ConfigObj {
```
zashraf1985, line 101: I see why you need two different configs/options but `ConfigObj` is not a good name. Two sets of feedback on this one.
1. One can be inherited from the other because outer options are a subset of internal options.
2. Names should be relevant. `Config` makes sense if its the only interface we use. `ConfigObj` does not make sense at all. And when we have two different objects with many similar properties, both `Config` and `ConfigObj` look like inappropriate choices. One choice could be `SDKOptions` for the outer interface and `OptimizelyOptions` for the internal . You can give it a thought and try to come up with better and more appropriate names.

packages/optimizely-sdk/lib/index.react_native.ts
```diff
@@ -30,30 +29,22 @@ import defaultErrorHandler from './plugins/error_handler';
...
- * @param  {Object}         config.eventDispatcher
- * @param  {Object}         config.logger
- * @param  {Object}         config.logLevel
- * @param  {Object}         config.userProfileService
- * @param  {Object}         config.eventBatchSize
- * @param  {Object}         config.eventFlushInterval
- * @param  {string}         config.sdkKey
- * @return {Object}         the Optimizely object
+ * @param  {Config} config
+ * @return {Optimizely} the Optimizely object
```
zashraf1985, line 45: this doc comment should be updated to reflect that it can also return null.

zashraf1985 (changes requested):
Looks good overall. I have some concerns on the naming of certain variables and Interfaces. Secondly, as a general rule of typescript conversion. Please feel free to simplify some code to use latest javascript constructs. A lot of code can be modernized and simplified  as now we are using typescript. For example: using spread operators instead of assign, using arrow functions where they make code look simpler etc.

packages/optimizely-sdk/lib/index.browser.ts
ozayr-zaviar: Yes it is throwing error without `as ConfigObj`

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -97,42 +88,41 @@ var createInstance = function(config) {
...
+      eventFlushInterval: DEFAULT_EVENT_FLUSH_INTERVAL,
+      eventDispatcher: eventDispatcher
+    };
+
+    const optimizelyLoggers = {
+      logger: logger,
+      errorHandler: getErrorHandler(),
+    };
+
+    const modifiedConfig = {...additionalEntities, ...config, ...optimizelyLoggers}  as OptimizelyOptions;
```
yavorona, line 103: I think can avoid using `as` type cast here by updating `isValidInstance` in `SDKOptions` object to not be optional.

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -97,42 +88,41 @@ var createInstance = function(config) {
...
+    const optimizely = new Optimizely(modifiedConfig);
 
     try {
       if (typeof window.addEventListener === 'function') {
-        var unloadEvent = 'onpagehide' in window ? 'pagehide' : 'unload';
+        const unloadEvent = 'onpagehide' in window ? 'pagehide' : 'unload';
         window.addEventListener(
           unloadEvent,
-          function() {
+          function () {
```
yavorona, line 125: Recommend using arrow function here

packages/optimizely-sdk/lib/index.node.ts
```diff
@@ -83,35 +73,34 @@ var createInstance = function(config) {
...
+      eventDispatcher: defaultEventDispatcher,
+      eventFlushInterval: DEFAULT_EVENT_FLUSH_INTERVAL
+    };
+
+    const optimizelyLoggers = {
+      logger: logger,
+      errorHandler: getErrorHandler(),
+    };
+
+    const modifiedConfig = {...additionalEntities, ...config, ...optimizelyLoggers}  as OptimizelyOptions;
```
yavorona, line 88: Same as in browser entry module: we can avoid using `as` type cast here by updating `isValidInstance` in `SDKOptions` object to not be optional.

packages/optimizely-sdk/lib/index.react_native.ts
```diff
@@ -78,41 +69,40 @@ var createInstance = function(config) {
...
+      eventMaxQueueSize: DEFAULT_EVENT_MAX_QUEUE_SIZE,
+      eventFlushInterval: DEFAULT_EVENT_FLUSH_INTERVAL
+    };
+
+    const optimizelyLoggers = {
+      logger: logger,
+      errorHandler: getErrorHandler(),
+    };
+
+    const modifiedConfig = {...additionalEntities, ...config, ...optimizelyLoggers}  as OptimizelyOptions;
```
yavorona, line 85: Same as comment above: we can avoid using `as` type cast here by updating `isValidInstance` in `SDKOptions` object to not be optional.

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -97,42 +87,41 @@ var createInstance = function(config) {
...
+      eventFlushInterval: DEFAULT_EVENT_FLUSH_INTERVAL,
+      eventDispatcher: eventDispatcher
+    };
+
+    const optimizelyLoggers = {
+      logger: logger,
+      errorHandler: getErrorHandler(),
+    };
+
+    const modifiedConfig = {...additionalEntities, ...config, ...optimizelyLoggers};
```
msohailhussain, line 102: `updatedConfig`

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -97,42 +87,41 @@ var createInstance = function(config) {
...
-      },
-      config,
-      {
-        eventDispatcher: eventDispatcher,
-        // always get the OptimizelyLogger facade from logging
-        logger: logger,
-        errorHandler: getErrorHandler(),
-      }
-    );
+    const additionalEntities = {
```
msohailhussain, line 90: need to change variable name.

packages/optimizely-sdk/lib/index.browser.ts
zashraf1985: the names `additionalEntities` and `optimizelyLoggers` do not make much sense. Instead of building interim objects, you can add those directly when building the `modifiedConfig` or `updatedConfig` object. something like
```
const modifiedConfig = {
  clientEngine: enums.JAVASCRIPT_CLIENT_ENGINE,
  eventBatchSize: DEFAULT_EVENT_BATCH_SIZE,
  eventFlushInterval: DEFAULT_EVENT_FLUSH_INTERVAL,
  eventDispatcher: eventDispatcher,
  ...config,
  logger: logger,
  errorHandler: getErrorHandler(),
}
```

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -97,42 +87,41 @@ var createInstance = function(config) {
...
+      logger: logger,
+      errorHandler: getErrorHandler(),
+    };
+
+    const modifiedConfig = {...additionalEntities, ...config, ...optimizelyLoggers};
 
     if (!eventProcessorConfigValidator.validateEventBatchSize(config.eventBatchSize)) {
       logger.warn('Invalid eventBatchSize %s, defaulting to %s', config.eventBatchSize, DEFAULT_EVENT_BATCH_SIZE);
-      config.eventBatchSize = DEFAULT_EVENT_BATCH_SIZE;
+      modifiedConfig.eventBatchSize = DEFAULT_EVENT_BATCH_SIZE;
```
zashraf1985, line 106: declaring this object above with such detail and then mutating some of the elements makes it confusing. more cleaner code could be to assign `eventBatchSize` to a local variable and then declare the `optimizelyOptions` once at the end.

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -97,42 +87,41 @@ var createInstance = function(config) {
...
     }
     if (!eventProcessorConfigValidator.validateEventFlushInterval(config.eventFlushInterval)) {
       logger.warn(
         'Invalid eventFlushInterval %s, defaulting to %s',
         config.eventFlushInterval,
         DEFAULT_EVENT_FLUSH_INTERVAL
       );
-      config.eventFlushInterval = DEFAULT_EVENT_FLUSH_INTERVAL;
+      modifiedConfig.eventFlushInterval = DEFAULT_EVENT_FLUSH_INTERVAL;
```
zashraf1985, line 114: The same comment about `eventBatchSize` applies here too

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -97,42 +87,41 @@ var createInstance = function(config) {
...
     }
 
-    var optimizely = new Optimizely(config);
+    const optimizely = new Optimizely(modifiedConfig);
 
     try {
       if (typeof window.addEventListener === 'function') {
-        var unloadEvent = 'onpagehide' in window ? 'pagehide' : 'unload';
+        const unloadEvent = 'onpagehide' in window ? 'pagehide' : 'unload';
```
zashraf1985, line 121: Nonblocking: Can we move these string values somewhere inside constants

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -97,42 +87,41 @@ var createInstance = function(config) {
...
         window.addEventListener(
           unloadEvent,
-          function() {
+          () => {
             optimizely.close();
           },
```
zashraf1985, line 126:
```suggestion
          optimizely.close,
```

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -164,8 +153,8 @@ export {
   setLogHandler as setLogger,
   setLogLevel,
   createInstance,
-  __internalResetRetryState,  
-}
+  __internalResetRetryState,
+};
 
 export default {
```
zashraf1985, line 159: since this is now converted to TS. The ones that are not renamed while exporting do not need name and value pair. It can now be.
```
export default {
  logging: loggerPlugin,
  errorHandler: defaultErrorHandler,
  eventDispatcher: defaultEventDispatcher,
  enums,
  setLogger: setLogHandler,
  setLogLevel,
  createInstance,
  __internalResetRetryState,
};

File: packages/optimizely-sdk/lib/index.d.ts [hunk 27]
```diff
@@ -51,6 +51,8 @@ declare module '@optimizely/optimizely-sdk' {
 
   export type DatafileOptions = import ('./shared_types').DatafileOptions;
 
+  export type SDKOptions = import ('./shared_types').SDKOptions;
```
zashraf1985, line 54: I am not sure about this but you can give it a try. Since `SDKOptions` is not used here, you can perhaps export it directly by doing something like
`export { SDKOptions } from './sharedTypes`
I am not totally sure about it. I might be totally wrong.

packages/optimizely-sdk/lib/index.browser.ts
zashraf1985: Also you can probably name it `optimizelyOptions` instead of `modifiedConfig`

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -77,6 +79,45 @@ export interface OptimizelyVariable {
...
+  logLevel?: LogLevel;
+  userProfileService?: UserProfileService;
+  eventBatchSize?: number;
+  eventFlushInterval?: number;
+  sdkKey?: string;
+  isValidInstance: boolean;
+}
+
+/**
+ * Config required to create optimizely object
```
zashraf1985, line 99: Probably change it to `Options required`

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -77,6 +79,45 @@ export interface OptimizelyVariable {
...
+  errorHandler?: ErrorHandler;
+  eventDispatcher?: EventDispatcher;
+  logger?: LogHandler;
+  logLevel?: LogLevel;
+  userProfileService?: UserProfileService;
+  eventBatchSize?: number;
+  eventFlushInterval?: number;
+  sdkKey?: string;
+  isValidInstance: boolean;
+}
```
zashraf1985, line 96: I believe we should probably add one line documentation comment to each option here. This might help while autogenerating docs

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -77,6 +79,45 @@ export interface OptimizelyVariable {
...
+  // eslint-disable-next-line  @typescript-eslint/ban-types
+  jsonSchemaValidator?: object;
+  sdkKey?: string;
+  userProfileService?: UserProfileService | null;
+  UNSTABLE_conditionEvaluators?: unknown;
+  eventFlushInterval?: number;
+  eventBatchSize?: number;
+  datafileOptions?: DatafileOptions;
+  eventMaxQueueSize?: number;
+  logger: LogHandler;
```
zashraf1985, line 118: NIT: Sort them alphabetically maybe?

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -77,6 +79,45 @@ export interface OptimizelyVariable {
...
+  datafile?: string;
+  errorHandler?: ErrorHandler;
+  eventDispatcher?: EventDispatcher;
+  logger?: LogHandler;
+  logLevel?: LogLevel;
+  userProfileService?: UserProfileService;
+  eventBatchSize?: number;
+  eventFlushInterval?: number;
+  sdkKey?: string;
+  isValidInstance: boolean;
```
zashraf1985, line 95: NIT: Sort them alphabetically maybe?

zashraf1985 (changes requested):
Detailed comments provided in browser entry file also apply to the other two entry points. I have NOT duplicated comments to all three entry points.

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -97,44 +87,40 @@ var createInstance = function(config) {
...
     try {
       if (typeof window.addEventListener === 'function') {
-        var unloadEvent = 'onpagehide' in window ? 'pagehide' : 'unload';
+        const unloadEvent = 'onpagehide' in window ? 'pagehide' : 'unload';
         window.addEventListener(
           unloadEvent,
-          function() {
-            optimizely.close();
-          },
+          optimizely.close,
```
mjc1283, line 123: Have you tested this change? I think it won't work. `close` is not bound to `optimizely`, so when it is invoked it may not have the right `this` value.

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -77,6 +79,55 @@ export interface OptimizelyVariable {
...
+  eventFlushInterval?: number;
+  eventMaxQueueSize?: number;
+  isValidInstance: boolean;
+  // TODO[OASIS-6649]: Don't use object type
+  // eslint-disable-next-line  @typescript-eslint/ban-types
+  jsonSchemaValidator?: object;
+  logger: LogHandler;
+  sdkKey?: string;
+  userProfileService?: UserProfileService | null;
+}
```
mjc1283, line 129: I think `OptimizelyOptions` should be defined in `lib/optimizely/index.ts`.

mjc1283 (changes requested):
Mostly LGTM, two small requests

packages/optimizely-sdk/lib/index.browser.ts
zashraf1985: Good catch! I asked uzair to change it this way to make it look simple and clean but it didn't occur to me that it will mess up with `this`. @ozayr-zaviar ! please change it back to what it was or an arrow function whatever makes sense.

mjc1283 (approved):

msohailhussain (approved):
lgtm

zashraf1985 (approved):
LGTM! Great Work @ozayr-zaviar

PR #597, opened 2020-10-09
feat: Upgrade TypeScript to v4.0.3 in optimizely-sdk package

yavorona:
## Summary
- Upgrade TypeScript in `optimizely-sdk` package to v4.0.3
- Use default export object for `fns` module to avoid stubbing issue that came out after the upgrade

## Test plan
 - Existing unit tests

### DIFF
--- a/packages/optimizely-sdk/lib/utils/event_processor_config_validator/index.ts
+++ b/packages/optimizely-sdk/lib/utils/event_processor_config_validator/index.ts
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { isSafeInteger } from '../fns';
+import fns from '../fns';
 
 /**
  * Return true if the argument is a valid event batch size, false otherwise
@@ -21,7 +21,7 @@ import { isSafeInteger } from '../fns';
  * @returns {boolean}
  */
 const validateEventBatchSize = function(eventBatchSize: unknown): boolean {
-  if (typeof eventBatchSize === 'number' && isSafeInteger(eventBatchSize)) {
+  if (typeof eventBatchSize === 'number' && fns.isSafeInteger(eventBatchSize)) {
     return eventBatchSize >= 1;
   }
   return false;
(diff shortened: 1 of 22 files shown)


packages/optimizely-sdk/lib/utils/fns/index.ts
```diff
@@ -41,23 +41,30 @@ export function assign(target: any, ...sources: any[]): any {
...
 }
+
+export default {
+  assign: assign,
+  currentTimestamp: currentTimestamp,
+  isSafeInteger: isSafeInteger,
+  keyBy: keyBy,
+  uuid: uuid,
+  isNumber: isNumber,
+}
```
spham1412, line 70: You can just do:

```
export default {
  assign,
  currentTimestamp,
  isSafeInteger,
  keyBy,
  uuid,
  isNumber,
}
```

packages/optimizely-sdk/lib/utils/event_processor_config_validator/index.ts
```diff
@@ -13,15 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { isSafeInteger } from '../fns';
+import fns from '../fns';
```
spham1412, line 16: Since you added the default export, the old implementation should work as it is, no need to change to import the whole `fns`?

spham1412 (changes requested):
One question: since you added default export, the `import { function } from 'fns';` should work as it as, isn't it?

packages/optimizely-sdk/lib/utils/event_processor_config_validator/index.ts
yavorona: Default object export does not allow destructured imports in this case.

packages/optimizely-sdk/lib/utils/fns/index.ts
```diff
@@ -18,7 +18,7 @@ import { generateUUID as uuid, keyBy as keyByUtil } from '@optimizely/js-sdk-uti
 const MAX_SAFE_INTEGER_LIMIT = Math.pow(2, 53);
 
 // eslint-disable-next-line
-export function assign(target: any, ...sources: any[]): any {
+const assign = function(target: any, ...sources: any[]): any {
```
mjc1283, line 21: Why do these need to be declared with `const`, instead of `function`?

yavorona: Good point. No need for that.

mjc1283 (approved):

spham1412 (approved):

PR #599, opened 2020-10-13
feat(flag-decisions): Add support for sending flag decisions

yavorona:
## Summary
- Add experiment key, feature flag key, and decision type to impression events.
- Send events for ALL decision types if `sendFlagDecisions` is enabled.
- Add metadata field to capture metadata.flag_key, metadata.rule_key, metadata.rule_type, metadata.variation_key

## Test plan
* All FSC and unit tests


yavorona:
This pr depends on https://github.com/optimizely/javascript-sdk/pull/601.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -486,7 +522,7 @@ export default class Optimizely {
    *                                      then clear the existing experiment-to-variation mapping.
    * @return {boolean}                    A boolean value that indicates if the set completed successfully.
    */
-  setForcedVariation(experimentKey: string, userId: string, variationKey: string | null): boolean {
+  setForcedVariation(experimentKey: string, userId: string, variationKey: string): boolean {
```
mjc1283, line 525: Is this an unrelated fix, or does it matter to flag decisions?

mjc1283 (changes requested):
My main suggestion is to try passing around the decision object that includes as much as possible (experiment, variation, rule key, rule type, etc.). The potential benefit is to reduce the number of additional parameters and repeated checking for null/empty string.

packages/optimizely-sdk/lib/optimizely/index.ts
yavorona: Let's keep it as it was, it should not matter to flag decisions. I updated `DecisionService.prototype.setForcedVariation` to avoid confusion.

mjc1283 (changes requested):
This looks great! The changes in the `Optimizely` class are so much cleaner 💯 .

Requesting changes for adding a few unit tests in the `decision` module.

packages/optimizely-sdk/lib/core/project_config/index.js
```diff
@@ -297,7 +297,7 @@ export var getVariationKeyFromId = function(projectConfig, variationId) {
  */
 export var getVariationIdFromExperimentAndVariationKey = function(projectConfig, experimentKey, variationKey) {
   var experiment = projectConfig.experimentKeyMap[experimentKey];
-  if (experiment.variationKeyMap.hasOwnProperty(variationKey)) {
+  if (variationKey !== '' && experiment.variationKeyMap.hasOwnProperty(variationKey)) {
```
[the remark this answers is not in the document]
yavorona: Good point. I will remove it.

mjc1283 (approved):
Please add an entry to CHANGELOG.md. Otherwise, LGTM.

pawels-optimizely (approved):
lgtm

PR #601, opened 2020-10-16
feat(flag_decisions): update decisionSnapshot in event-processor to include metadata object 

yavorona:
## Summary
 - update decisionSnapshot to include metadata object to support sending flag decisions

## Test plan
Unit tests
## Issues
- [OASIS-7113](https://optimizely.atlassian.net/browse/OASIS-7113)

### DIFF
--- a/packages/event-processor/src/events.ts
+++ b/packages/event-processor/src/events.ts
@@ -45,18 +45,22 @@ export interface ImpressionEvent extends BaseEvent {
   }
 
   layer: {
-    id: string
+    id: string | null
   } | null
 
   experiment: {
-    id: string
+    id: string | null
     key: string
   } | null
 
   variation: {
-    id: string
+    id: string | null
     key: string
   } | null
+
+  ruleKey: string
+  flagKey: string
+  ruleType: string
 }
 
 export interface ConversionEvent extends BaseEvent {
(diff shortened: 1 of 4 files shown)


packages/event-processor/src/events.ts
```diff
@@ -45,18 +45,22 @@ export interface ImpressionEvent extends BaseEvent {
...
 
   experiment: {
-    id: string
+    id: string | null
     key: string
   } | null
 
   variation: {
-    id: string
+    id: string | null
```
mjc1283, line 57: Why are these now allowed to be `null`?

mjc1283 (approved):
LGTM - one question about layer/experiment/variation id being null inside `ImpressionEvent`.

packages/event-processor/src/events.ts
yavorona: When decision.experiment is not defined, we assign a ruleKey/experimentKey to an empty string, in which case I assign an experimentId and layerId to null. Similarly, the variationId is now null when variation is not defined.  Please see the implementation [here](https://github.com/optimizely/javascript-sdk/pull/599/files#diff-78019ffee5096346957fb1ff408d5bdc8b27c9c40318f57e651953a455d73b3eR48) for details. Another way would be to assign these params to an empty string instead of null. LMKWYT @mjc1283

mjc1283: Sounds good, I think using `null` is ok.

PR #602, opened 2020-10-16
Update package.json

### LINKED ISSUE #485: Installation triggers not-applicable peerDep warning
## What I wanted to do

Install optimizely-sdk package

## What I expected to happen

I expected to have no warnings or errors.

## What actually happened

```
npm WARN @optimizely/js-sdk-datafile-manager@0.5.0 requires a peer of @react-native-community/async-storage@^1.2.0 but none is installed. You must install peer dependencies yourself.
```

This dependency is not applicable for projects that are not using react-native, so it should not be marked as a required peerDep. It should be optional.

## Steps to reproduce

`npm i @optimizely/optimizely-sdk@4.0.0`

**`@optimizely/optimizely-sdk` version:**  4.0.0

babeya:
Remove dependencies that are both in dev and peer dependencies

## Summary
- Remove devDependencies and peerDependencies which are the same 

Both dependencies triggers warning on our application about missing peer dependencies. It may not be the right way to do it properly (maybe including them in the package.json would fix the issue), it's more a way to put some light on this issue

Have a good day

PS: sorry for my broken english.

## Test plan

None 
## Issues
Fixes #485

### DIFF
--- a/packages/event-processor/package.json
+++ b/packages/event-processor/package.json
@@ -40,19 +40,15 @@
   },
   "dependencies": {
     "@optimizely/js-sdk-logging": "^0.1.0",
-    "@optimizely/js-sdk-utils": "^0.4.0"
+    "@optimizely/js-sdk-utils": "^0.4.0",
+    "@react-native-community/netinfo": "5.9.4",
+    "@react-native-community/async-storage": "^1.2.0"
   },
   "devDependencies": {
-    "@react-native-community/async-storage": "^1.2.0",
-    "@react-native-community/netinfo": "^5.9.4",
     "@types/jest": "^24.0.9",
     "jest": "^23.6.0",
     "jest-localstorage-mock": "^2.4.0",
     "ts-jest": "^23.10.5",
     "typescript": "^4.0.3"
-  },
-  "peerDependencies": {
-    "@react-native-community/netinfo": "5.9.4",
-    "@react-native-community/async-storage": "^1.2.0"
   }
 }


aliabbasrizvi:
@babeya thank you for your submission. As a first time contributor, can you please accept our Contributor License Agreement in [CONTRIBUTING.md](https://github.com/optimizely/javascript-sdk/blob/master/CONTRIBUTING.md)

ZebraFlesh:
I'm confused: why are these packages being added to `dependencies` instead of `optionalDependencies`? I'm not using react-native so why would I want these packages?

babeya:
@ZebraFlesh I agree, i also do not use react-native and don't want to include module deticated to it. 

But it triggers warning about missing peers on my applications. 
I don't want to include missing dependencies in my apps as it would feel even worst as adding them here. I don't think it's the right solution, i just wanted to raise awarness on this particular issue, collect your inputs and fix it properly.

I didn't know about optionnalDependencies, but from what i read, it will be installed anyway ? It's only useful if the install fail  ?
https://docs.npmjs.com/files/package.json#optionaldependencies 
https://classic.yarnpkg.com/en/docs/dependency-types/

Have a good day !

ZebraFlesh:
> I didn't know about optionnalDependencies, but from what i read, it will be installed anyway ? It's only useful if the install fail ?

`optionalDependencies` may either be installed or missing at run time. It's a mechanism for package authors to declare, "I might use these versions of these packages if they are present". My understanding is it would allow the non-react-native projects to skip these dependencies. The react-native projects could them include them. The caveat is that code changes may be required, such as failing gracefully when one of the optional packages cannot be found.

babeya:
@ZebraFlesh Yep it sounds better in optionnalDependencies than in dependencies. 
But npm will still try to install it, in the end, if the module installation does not fail, we still have a react-native module in a react application.

mhatvan:
I am also experiencing these unmet peer dependency warnings for `@react-native-community/*` packages, is there any update on this?

keppel2 (dismissed):

mhatvan:
This might be helpful: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta

zashraf1985:
Made the react native peer dependencies optional which should resolve this issue.

PR #606, opened 2020-10-30
feat: setup TypeDoc for generating API documentation 

yavorona:
## Summary
- setup Typedoc for generating optimizely-sdk API documentation. As a start, Typedoc is setup in `optimizely-sdk` package. Once all packages are upgraded to TS v.4.0.3 including the root directory, the typedoc will be setup in the root of js-sdk.

Files included in the docs in this pr:
- optimizely/index.ts
- project_config_manager.d.ts (will be updated to project_config_manager.ts once converted to TS)
- optimizely-sdk README.md

## Test plan
Existing unit tests

## Issues
- [OASIS-7279](https://optimizely.atlassian.net/browse/OASIS-7279)

[OASIS-7279]: https://optimizely.atlassian.net/browse/OASIS-7279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/optimizely-sdk/docs/config/typedoc.json
+++ b/packages/optimizely-sdk/docs/config/typedoc.json
@@ -1,18 +1,19 @@
 {
   "inputFiles": [
-    "./lib/optimizely/index.ts",
-    "./lib/core/optimizely_config/index.ts",
-    "./lib/core/notification_center/index.ts"
+    "./lib/optimizely",
+    "./lib/core/optimizely_config",
+    "./lib/core/notification_center",
   ],
   "out": "./docs/out",
-  "exclude": "**/*+(.spec|.d|.tests).(ts/js)",
+  "mode": "modules",
   "excludePrivate": true,
   "excludeProtected": true,
+  "excludeNotDocumented": true,
   "excludeExternals": true,
+  "hideGenerator": true,
   "readme": "README.md",
   "name": "Optimizely JavaScript SDK",
   "includeVersion": true,
-  "includeDeclarations": true,
   "theme": "./node_modules/typedoc-neo-theme/bin/default",
   "links": [{
     "label": "GitHub",
@@ -20,10 +21,10 @@
   }],
   "outline": [
     {
-      "API Reference": { "Optimizely's APIs": "classes/optimizely" },
+      "API Reference": { "Optimizely's APIs": "modules/_optimizely_index_" },
       "Configuration Data": {
-        "Optimizely Config": "classes/optimizelyconfig",
-        "Notification Center": "classes/notificationcenter"
+        "Optimizely Config": "modules/_core_optimizely_config_index_",
+        "Notification Center": "modules/_core_notification_center_index_"
       }
     }
   ]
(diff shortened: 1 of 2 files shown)


packages/optimizely-sdk/docs/README.md
```diff
@@ -0,0 +1,13 @@
...
+Getting Started
+---------------
+
+### Building documentation locally
+
+To generate JavaScript SDK documentation locally, execute the following command from `optimizely-sdk/` directory:
+
+    npm run build-docs
+
+This will build HTML docs in `docs/out/index.html`. Open this file in your web browser to see the docs.
```
mjc1283, line 13: I think we should have a link from the main README, to this file. cc @fscelliott, let us know what you think.

packages/optimizely-sdk/docs/config/typedoc.json
```diff
@@ -0,0 +1,26 @@
+{
+  "inputFiles": [
+    "./lib/optimizely/index.ts",
+    "./lib/core/project_config/project_config_manager.d.ts"
+  ],
```
mjc1283, line 5: Let's include the entry points in here. `createInstance` is the first method most users will need to know about.

mjc1283 (changes requested):
The output looks really good! A few comments:

- There is a `globals` section that shows `Optimizely`, `OptimizelyOptions`, `ProjectConfigManager`, `createProjectConfigManager`. This seems wrong - we don't create any global variables.
- `ProjectConfigManager` also has its own section. User's won't use it directly so I don't think it needs to be documented so prominently. Especially when other APIs that are used directly are missing (for example `OptimizelyConfig`, `NotificationCenter`). It would be ideal if everything in lib/index.d.ts and the entry point files was documented.
- The `close()` method in the Optimizely APIs section has something wrong with the formatting in its return value section.

packages/optimizely-sdk/docs/README.md
fscelliott: @mjc1283 I haven't' been following the typescript efforts. So it looks as if we're adding capability for users to generate the docs themselves locally, but I assume we aren't publishing the autogenerated docs to any permanent URL? 
Also I want to understand the context for the link. I thought Agent was based on Go SDK.  But it looks like the typedoc is actually being generated from https://github.com/optimizely/javascript-sdk in typedocs.json? Why would they want the JS docs in this Go context?

fscelliott: Also @mjc1283 I can review the typedocs once they're ready to be generated w/ npm run build-docs (when I see this PR closed, I'll assume that's the case)

mjc1283: > So it looks as if we're adding capability for users to generate the docs themselves locally, but I assume we aren't publishing the autogenerated docs to any permanent URL?

Not right now we aren't, but I think that can happen later.

> Also I want to understand the context for the link. I thought Agent was based on Go SDK. But it looks like the typedoc is actually being generated from https://github.com/optimizely/javascript-sdk in typedocs.json? Why would they want the JS docs in this Go context?

This is auto-generated API documentation for JavaScript SDK - unrelated to Go SDK or Agent.

> Also @mjc1283 I can review the typedocs once they're ready to be generated w/ npm run build-docs (when I see this PR closed, I'll assume that's the case)

👍  Thanks!

PR #617, opened 2020-11-10
feat: add decide api

yavorona:
## Summary
Implemented:
- `OptimizelyUserContext`
- `createUserContext`, `decide` in optimizely client
-  partial `Decide Reasons`

## Test plan
Testing with local bundle and unit tests (some unit tests need fixing stubs)

## Issues
[OASIS-6934](https://optimizely.atlassian.net/browse/OASIS-6934)
[OASIS-6940](https://optimizely.atlassian.net/browse/OASIS-6940)
[OASIS-6942](https://optimizely.atlassian.net/browse/OASIS-6942)
[OASIS-6943](https://optimizely.atlassian.net/browse/OASIS-6943)


packages/optimizely-sdk/lib/optimizely/index.tests.js
```diff
@@ -210,7 +210,6 @@ describe('lib/optimizely', function() {
         it('should not log an error when sdkKey is provided and datafile is not provided', function() {
           new Optimizely({
             clientEngine: 'node-sdk',
-            eventBuilder: eventBuilder,
```
yavorona, line 213: Seems  unnecessary to provide `eventBuilder` to create Optimizely instance.

packages/optimizely-sdk/lib/optimizely_user_context/index.tests.js
```diff
@@ -0,0 +1,316 @@
...
+            variationKey: 'variation_with_traffic',
+            enabled: true,
+            variables: variablesExpected,
+            ruleKey: 'exp_no_audience',
+            flagKey: flagKey,
+            userContext: user,
+            reasons: [],
+          }
+          assert.deepEqual(decision, expectedDecisionObj);
+          //TODO: figure out stubbing of eventDispatcher
```
yavorona, line 219: Works when tested locally. Just need to figure out how to properly stub it.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -110,6 +114,12 @@ export default class Optimizely {
     this.isOptimizelyConfigValid = config.isValidInstance;
     this.logger = config.logger;
 
+    let defaultDecideOptions = config.defaultDecideOptions ?? [];
+    if (!Array.isArray(defaultDecideOptions)) {
+      this.logger.log(LOG_LEVEL.DEBUG, sprintf(LOG_MESSAGES.INVALID_DECIDE_OPTIONS, MODULE_NAME));
```
mjc1283, line 119: I believe the logger has format string support built in, there's no need to call `sprintf` here. Try this:
```suggestion
      this.logger.debug(LOG_MESSAGES.INVALID_DECIDE_OPTIONS, MODULE_NAME);
```
A lot of the log statements in this model do not use this because they predate this support being added in the logger.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -1415,4 +1425,159 @@ export default class Optimizely {
...
+      }
+    }
+
+    return allDecideOptions;
+  }
+
+  decide(
+    user: OptimizelyUserContext,
+    key: string,
+    options?: OptimizelyDecideOptions[]
```
mjc1283, line 1480: Would it simplify the code to provide a default, then we know `options` is always an array?
```suggestion
    options: OptimizelyDecideOptions[] = [],
```

packages/optimizely-sdk/lib/optimizely_decision/index.ts
```diff
@@ -0,0 +1,64 @@
...
+import OptimizelyUserContext from '../optimizely_user_context';
+
+export default class OptimizelyDecision {
+  private variationKey?: string | null;
+  private enabled: boolean;
+  private variables: { [variableKey: string]: unknown } | null;
+  private ruleKey?: string | null;
+  private flagKey: string;
+  private userContext: OptimizelyUserContext;
+  private reasons: string[];
```
mjc1283, line 25: I don't think these should be private - these decisions go back to the end user, they need access to these things, right?

Making them `readonly` might be useful, also.

Since there are no non-static methods, I don't see a strong need to make this a class, rather than an interface. But I don't mind it much either.

packages/optimizely-sdk/lib/optimizely_user_context/index.tests.js
mjc1283: IMO, this does not need to be tested here. It's the responsibility of other modules, not `OptimizelyUserContext`. Using a fake Optimizely client will make it easier to orchestrate the tests you do need. `OptimizelyUserContext` should return whatever decision came from the inner client.

packages/optimizely-sdk/lib/optimizely_user_context/index.tests.js
```diff
@@ -0,0 +1,316 @@
...
+    var createdLogger = logger.createLogger({
+      logLevel: LOG_LEVEL.DEBUG,
+      logToConsole: false,
+    });
+    describe('#setAttribute', function() {
+      it('should set attributes when provided at instantiation of OptimizelyUserContext', function() {
+        var userId = 'user1';
+        var attributes = { test_attribute: 'test_value' };
+        var user = new OptimizelyUserContext({
+          optimizely,
```
mjc1283, line 49: I'm confused - where is a value assigned to `optimizely`? I see it declared on 39 but not assigned. How do tests work if `optimizely` is `undefined`?

packages/optimizely-sdk/lib/optimizely_user_context/index.tests.js
```diff
@@ -0,0 +1,316 @@
...
+        optimizely = new Optimizely({
+          clientEngine: 'node-sdk',
+          datafile: getTestDecideProjectConfig(),
+          errorHandler: errorHandler,
+          eventDispatcher: eventDispatcher,
+          jsonSchemaValidator: jsonSchemaValidator,
+          logger: createdLogger,
+          isValidInstance: true,
+          defaultDecideOptions: [],
+        });
```
mjc1283, line 143: It's better to create a fake optimizely object that fulfills the same interface rather than creating a real one. OptimizelyUserContext and its unit tests should be isolated from the real Optimizely implementation.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -92,6 +95,7 @@ export default class Optimizely {
   private notificationCenter: NotificationCenter;
   private decisionService: DecisionService;
   private eventProcessor: EventProcessor;
+  private defaultDecideOptions?: OptimizelyDecideOptions[];
```
mjc1283, line 98: From how this is used in `getAllDecideOptions`, it seems better to save them as an object with boolean properties, like
```ts
private defaultDecideOptions: { [key: OptimizelyDecideOptions]: boolean | undefined }
```

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -1415,4 +1425,159 @@ export default class Optimizely {
...
+
+  /**
+   *
+   * Get all decide options.
+   *
+   * @param  {OptimizelyDecideOptions[]}          options     Optional decide options
+   * @return {OptimizelyDecideOptions[]}          Array of all provided decide options
+   */
+  private getAllDecideOptions(options?: OptimizelyDecideOptions[]): OptimizelyDecideOptions[] {
+    const allDecideOptions = this.defaultDecideOptions ?? [];
```
mjc1283, line 1461: We don't want to mutate `this.defaultDecideOptions`:
```suggestion
    const allDecideOptions = [...this.defaultDecideOptions] ?? [];
```
Also, could remove the last part if we ensure `this.defaultDecideOptions` is an array.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -1415,4 +1425,159 @@ export default class Optimizely {
...
+    if (options !== undefined) {
+      if (!Array.isArray(options)) {
+        this.logger.log(LOG_LEVEL.DEBUG, sprintf(LOG_MESSAGES.INVALID_DECIDE_OPTIONS, MODULE_NAME));
+      } else {
+        options.forEach((option) => {
+          if (!allDecideOptions.includes(option) && OptimizelyDecideOptions[option]) {
```
mjc1283, line 1467: Even though the array length is always small, it irks me that we are searching with `includes` every time. Since we're repeatedly looking up options, we could create an object for fast lookups - perhaps another private property.

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -117,6 +117,14 @@ export interface Rollout {
   experiments: Experiment[];
 }
 
+export enum OptimizelyDecideOptions {
+  DISABLE_DECISION_EVENT = 'DISABLE_DECISION_EVENT',
+  ENABLED_FLAGS_ONLY =  'ENABLED_FLAGS_ONLY',
+  IGNORE_USER_PROFILE_SERVICE = 'IGNORE_USER_PROFILE_SERVICE',
+  INCLUDE_REASONS = 'INCLUDE_REASONS',
+  EXCLUDE_VARIABLES = 'EXCLUDE_VARIABLES'
+}
```
mjc1283, line 126: Let's make sure this is usable when running the SDK from regular JS, not TS. `enum`s are not only types, they have a runtime representation too. Perhaps it's fine, but something to be aware of. I believe everything else in this module is a pure type definition that gets erased.

mjc1283 (changes requested):
Feel free to break this into several PRs for separate parts of the implementation.

packages/optimizely-sdk/lib/optimizely/index.ts
yavorona: I think this would require a bit more work than just changing it as shown above. Right now we have logger as `ConsoleLogHandler`, where it would require changing it to `OptimizelyLogger` to support built-in format string. I will create a separate ticket to have it updated.

packages/optimizely-sdk/lib/shared_types.ts
yavorona: I think moving it to `@optimizely/optimizely-sdk/lib/utils/enums` module might be a better idea!

PR #621, opened 2020-11-11
refactor: SDKOptions refactor

ozayr-zaviar:
## Summary
- `SDKOptions` updated to `Config`

## Test plan
- All existing test cases should paas.
- All FSC test cases should pass.


packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -80,29 +81,30 @@ export interface OptimizelyVariable {
...
 /**
+ * The options object given to Optimizely.createInstance.
  * Entry level Config Entities
  */
-export interface SDKOptions {
-  // Datafile string
+export interface Config {
+  // TODO[OASIS-6649]: Don't use object type
+  // eslint-disable-next-line  @typescript-eslint/ban-types
   datafile?: string;
```
mjc1283, line 90: We must maintain the same exact `Config` interface that was previously available.
```suggestion
  datafile?: object | string;
```

mjc1283 (changes requested):

packages/optimizely-sdk/lib/index.d.ts
```diff
@@ -51,7 +51,7 @@ declare module '@optimizely/optimizely-sdk' {
 
   export type DatafileOptions = import ('./shared_types').DatafileOptions;
 
-  export type SDKOptions = import ('./shared_types').SDKOptions;
+  export type Config = import ('./shared_types').Config;
```
mjc1283, line 54: Please add this for backwards compatibility, in case a user was importing & using `SDKOptions`:

```ts
  // Keep this alias for backwards compatibility
  export type SDKOptions = Config;
```

mjc1283 (changes requested):

zashraf1985:
This is not needed any more as it will be covered by a better fix in #745

zashraf1985:
This is not needed any more as it will be covered by a better fix in #745

PR #624, opened 2020-11-12
feat: Convert notification center to TS

yavorona:
## Summary
- Convert `notification_center` module to TS

## Test plan

Existing unit tests

## Issues
[OASIS-7067](https://optimizely.atlassian.net/browse/OASIS-7067)


packages/optimizely-sdk/lib/core/notification_center/index.ts
```diff
@@ -0,0 +1,237 @@
...
+   * If there was an error and the listener was not added, addNotificationListener returns -1. This
+   * can happen if the first argument is not a valid notification type, or if the same callback
+   * function was already added as a listener by a prior call to this function.
+   */
+  addNotificationListener<T extends ListenerPayload>(
+    notificationType: string,
+    callback: NotificationListener<T>
+  ): number {
+    try {
+      const isNotificationTypeValid = objectValues(NOTIFICATION_TYPES ).indexOf(notificationType as notificationTypesEnum) > -1;
```
mjc1283, line 89: Not requesting change, but this is coming up a lot. I wonder if there's a more idiomatic solution for TS that doesn't require using `as`  when iterating over `Object.keys()/Object.values()` return value.

mjc1283 (approved):
LGTM. Brief comments added.

packages/optimizely-sdk/lib/core/notification_center/index.ts
```diff
@@ -0,0 +1,237 @@
...
+} from '../../utils/enums';
+
+const MODULE_NAME = 'NOTIFICATION_CENTER';
+
+export interface NotificationCenterOptions {
+  logger: LogHandler;
+  errorHandler: ErrorHandler;
+}
+
+export interface ListenerEntry {
```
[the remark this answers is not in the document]
yavorona: No, will be updated.

packages/optimizely-sdk/lib/core/notification_center/index.ts
yavorona: Good point. Not sure how to avoid using it here, will look into it.

PR #625, opened 2020-11-13
fix: Refactor index.d.ts imports

yavorona:
## Summary
 - Refactor `lib/index.d.ts` imports such that it only imports from `shared_files`.

## Test plan
- Unit tests


mjc1283 (approved):
Thanks @yavorona ! LGTM

PR #626, opened 2020-11-18
feat: Convert optimizely_config module to TS

yavorona:
## Summary
 - Convert `optimizely_config` module to TS

## Test plan
Existing unit and fsc tests and running local bundle
## Issues
- [OASIS-7068](https://optimizely.atlassian.net/browse/OASIS-7068)


packages/optimizely-sdk/lib/core/optimizely_config/index.ts
```diff
@@ -0,0 +1,250 @@
...
+  getDatafile(): string {
+    return this.datafile;
+  }
+
+  /**
+   * Get Experiment Ids which are part of rollout
+   * @param       {Rollout[]}                  rollouts
+   * @returns     {ExperimentIds}              Experiment Ids which are part of rollout
+   */
+  private getRolloutExperimentIds(rollouts: Rollout[]): ExperimentIds {
```
mjc1283, line 92: For `getRolloutExperimentIds`, `getExperimentsMap`, `getMergedVariablesMap`, and `getFeaturesMap`, I don't think they need to be class methods. They don't access any data from `this`, they just use arguments. I suggest leaving them as separate (unexported) functions in this module, or making them static methods.

mjc1283 (changes requested):

mjc1283 (approved):

PR #643, opened 2021-01-13
feat: Clone OptimizelyUserContext before calling decide api

yavorona:
## Summary
The OptimizelyUserContext instance is returned as a part of decision at the end of the decide api call. If the instance is not cloned at the beginning, it may have different states than one used for the decision.
In this pr, cloning of user context is implemented before calling decide api.

## Test plan
Unit tests + Manual testing

## Issues
- [OASIS-2700](https://optimizely.atlassian.net/browse/OASIS-7400)


jaeopt (approved):
LGTM

mjc1283 (approved):

PR #644, opened 2021-01-15
fix: Change experimentKey and variationKey to be null in rollout in decisionInfo payload

yavorona:
## Summary
 - This pr changes `variationKey` and `ruleKey` from empty string to `null`  in a rollout `decisionInfo` payload in order to pass FSC tests.
 - Changed `OptimizelyDecideOptions` enums to `OptimizelyDecideOption` to be consistent with other SDKs.
## Test plan
- Unit tests
- [FSC tests](https://travis-ci.com/github/optimizely/fullstack-sdk-compatibility-suite/builds/213362981) with `uzair/decide-api-endpoints` testapp branch - all pass except for DECIDE WORLD tests, which will be addressed separately

### DIFF
--- a/packages/optimizely-sdk/lib/optimizely/index.ts
+++ b/packages/optimizely-sdk/lib/optimizely/index.ts
@@ -1496,8 +1496,8 @@ export default class Optimizely {
     reasons.push(...decisionVariation.reasons);
     const decisionObj = decisionVariation.result;
     const decisionSource = decisionObj.decisionSource;
-    const experimentKey = decision.getExperimentKey(decisionObj);
-    const variationKey = decision.getVariationKey(decisionObj);
+    const experimentKey = decisionObj.experiment?.key ?? null;
+    const variationKey = decisionObj.variation?.key ?? null;
     const flagEnabled: boolean = decision.getFeatureEnabledFromVariation(decisionObj);
     if (flagEnabled === true) {
       this.logger.log(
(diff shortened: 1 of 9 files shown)


packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -1496,8 +1496,8 @@ export default class Optimizely {
     reasons.push(...decisionVariation.reasons);
     const decisionObj = decisionVariation.result;
     const decisionSource = decisionObj.decisionSource;
-    const experimentKey = decision.getExperimentKey(decisionObj);
-    const variationKey = decision.getVariationKey(decisionObj);
```
mjc1283, line 1500: Are `getExperimentKey` & `getVariationKey` methods no longer used, so should delete them?

mjc1283 (approved):

jaeopt (approved):
LGTM

packages/optimizely-sdk/lib/optimizely/index.ts
yavorona: These methods are being used in `buildImpressionEvent` and `emitNotificationCenterActivate` methods (we are still sending `rule_key` and `variation_key`   as empty strings inside `metadata` object in this case). I am not sure if they should be changed to `null`. @jaeopt do you have any thought on that?

jaeopt: @yavorona I think we should keep those. Event and notification will have different requirements (empty string instead of null) than optimizelyDecision (null).

PR #647, opened 2021-02-02
feat: defaultDecideOptions in SDKOptions

ozayr-zaviar:
## Summary
- `defaultDecideOptions` in SDKOption type added.

## Test plan
- All testcases including `@DEFAULT_DECIDE_OPTIONS` should pass.

### DIFF
--- a/packages/optimizely-sdk/lib/shared_types.ts
+++ b/packages/optimizely-sdk/lib/shared_types.ts
@@ -207,6 +207,8 @@ export interface SDKOptions {
   sdkKey?: string;
   // user profile that contains user information
   userProfileService?: UserProfileService;
+  // dafailt options for decide API
+  defaultDecideOptions?: OptimizelyDecideOption[] | undefined
 }
 
 export type OptimizelyExperimentsMap = {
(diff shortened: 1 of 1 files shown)


packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -207,6 +207,8 @@ export interface SDKOptions {
   sdkKey?: string;
   // user profile that contains user information
   userProfileService?: UserProfileService;
+  // dafailt options for decide API
+  defaultDecideOptions?: OptimizelyDecideOption[] | undefined
```
yavorona, line 211: Not sure if specifying `undefined` as a type is needed in this case as it is an optional parameter. WDYT @mjc1283 ?

mjc1283: Agree, @yavorona . @ozayr-zaviar please fix.

mjc1283 (changes requested):

yavorona (approved):

mjc1283 (approved):

PR #655, opened 2021-03-01
fix: Fix type definitions to avoid compiling source files

yavorona:
## Summary
To prevent users from compiling our TS source files, the type definitions file `index.d.ts` has been updated to exclude imports from our source files (`OptimizelyUserContext` and `OptimizelyDecision`).

## Test plan
Existing unit tests


zashraf1985 (approved):
LGTM!

PR #658, opened 2021-03-23
feat: Convert logger module to TS

yavorona:
## Summary
- Convert `index.js` and `index.react_native.js`to TS
- Remove `enums.js` since it is not consumed by any module

## Test plan
Existing unit tests


packages/optimizely-sdk/lib/core/bucketer/index.tests.js
```diff
@@ -24,7 +24,7 @@ import {
   LOG_MESSAGES,
   LOG_LEVEL,
 } from '../../utils/enums';
-import logger from '../../plugins/logger';
+import * as logger from '../../plugins/logger';
```
zashraf1985, line 27: Does importing by name work?

packages/optimizely-sdk/lib/plugins/logger/index.ts
```diff
@@ -13,21 +13,24 @@
...
+type ConsoleLogHandlerConfig = {
+  logLevel?: LogLevel | string;
+  logToConsole?: boolean;
+  prefix?: string;
+}
 
-NoOpLogger.prototype.log = function() {};
+export class NoOpLogger {
+  log(): void {
+    () => {};
```
zashraf1985, line 26: can this be removed?

zashraf1985 (changes requested):

packages/optimizely-sdk/lib/core/bucketer/index.tests.js
yavorona: good point, imported as individual export

packages/optimizely-sdk/lib/plugins/logger/index.ts
yavorona: `NoOpLogger`is consumed by react-native logger, not sure if it is a good idea to remove it

zashraf1985: Oh no. I m just talking about the empty arrow function wrapped inside the log function. Think that extra empty arrow function might be redundant

yavorona: Oh that is true, it is not needed here and is now removed

packages/optimizely-sdk/lib/plugins/logger/index.react_native.ts
```diff
@@ -31,38 +32,30 @@ function getLogLevelName(level) {
...
-export var createNoOpLogger = function() {
+export function createNoOpLogger(): NoOpLogger {
   return new NoOpLogger();
 }
-
-export default {
-  createLogger,
-  createNoOpLogger,
-};
+
```
zashraf1985, line 61: I dont know why github is showing `no newline at end of file` error even though it appears that there is a new line there. Can you check.

packages/optimizely-sdk/lib/plugins/logger/index.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016-2017, 2020, Optimizely
+ * Copyright 2021, Optimizely
```
zashraf1985, line 2: Shouldn't this include previous years too. Technically, this is a new file but logically this is a rename. Just a suggestion, feel free to do whatever makes sense to you here

zashraf1985 (approved):
LGTM! Approved with a few optional suggestions

PR #680, opened 2021-06-08
feat: Convert event builder module to TS

yavorona:
## Summary
- Convert `event_builder` module to TS

## Test plan
Existing unit tests


packages/optimizely-sdk/lib/core/event_builder/event_helpers.ts
```diff
@@ -0,0 +1,256 @@
...
+  type: 'conversion';
+  timestamp: number;
+  uuid: string;
+  user: {
+    id: string;
+    attributes: VisitorAttribute[];
+  };
+  context: EventContext;
+  event: {
+    id: string | null;
```
keppel2, line 110: Could this be `id:? string;` ?

packages/optimizely-sdk/lib/core/event_builder/index.ts
```diff
@@ -0,0 +1,315 @@
...
+  ruleKey: string,
+  ruleType: string,
+  flagKey: string,
+  enabled: boolean
+): Snapshot {
+
+  const campaignId = experimentId ? getLayerId(configObj, experimentId) : null;
+
+  let variationKey = variationId ? getVariationKeyFromId(configObj, variationId) : null;
+  variationKey = variationKey ? variationKey: '';
```
zashraf1985, line 197:
```suggestion
  variationKey = variationKey || '';
```

zashraf1985 (changes requested):

packages/optimizely-sdk/lib/core/event_builder/event_helpers.ts
zashraf1985: @keppel2 you are right by intuition but JS and and TS are weird languages. `id?` is equivalent to `string | undefined` but not equivalent to `string | null`.

packages/optimizely-sdk/lib/core/event_builder/event_helpers.ts
```diff
@@ -0,0 +1,256 @@
...
+    key: string;
+  };
+  revenue: number | null;
+  value: number | null;
+  tags: EventTags | undefined;
```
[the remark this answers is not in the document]
yavorona: Not sure about this change. I think declaring this property as optional may not be the right way since we always pass tags to `buildConversionEvent` method. Its value can be of type `EventTags` or `undefined` because we filter out its null values before passing it to buildConversionEvent or it may simply be undefined because user did not specify it in `track` call. @zashraf1985 Please LMKWYT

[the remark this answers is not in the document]
zashraf1985: I think it makes sense. Ignore my comment on this

packages/optimizely-sdk/lib/core/event_builder/event_helpers.ts
```diff
@@ -0,0 +1,259 @@
...
+  enabled,
+  userAttributes,
+  clientEngine,
+  clientVersion,
+}: ImpressionConfig): ImpressionEvent {
+
+  const ruleType = decisionObj.decisionSource;
+  const experimentKey = decision.getExperimentKey(decisionObj);
+  const variationKey = decision.getVariationKey(decisionObj);
+
```
zashraf1985, line 138: Super NIT: Extra line break

zashraf1985 (approved):
LGTM!

PR #681, opened 2021-06-09
fix:(event processor): Update ConversionEvent interface 

yavorona:
## Summary
- Update `ConversionEvent` interface to allow event `id` type `null` and `tags` type `undefined`

The "why", or other context.

## Test plan
- Existing unit tests + added new tests


zashraf1985 (approved):
Looks good but just curios why is it needed.

yavorona:
> Looks good but just curios why is it needed.

This is needed to avoid compiler errors in `optimizely-sdk` package, where we also allow event `id` to be `null` and `tags` to be `undefined`  when building a conversionEvent obj that is being processed by event processor in our `optimizely` module.

PR #690, opened 2021-07-14
fix: Send correct impression event data for projects with duplicate rule keys

yavorona:
## Summary
Provide correct variation id and experiment id when building and sending an impression event.

## Test plan
local bundle + unit tests + fsc


jaeopt (approved):
LGTM

packages/optimizely-sdk/lib/core/event_builder/event_helpers.ts
```diff
@@ -134,10 +133,10 @@ export const buildImpressionEvent = function({
 
   const ruleType = decisionObj.decisionSource;
   const experimentKey = decision.getExperimentKey(decisionObj);
```
msohailhussain, line 135: where this is needed in the code?

PR #694, opened 2021-07-22
Add interface DecisionListenerPayload

piotrski:
## Summary
- Added missing type for `DecisionListenerPayload`

I'm not sure what should be the correct type for `type` property.
[Beta version of docs](https://docs.developers.optimizely.com/full-stack/v4.0/docs/decision-notification-listener#section-parameters) says it's `flag` but [current version](https://docs.developers.optimizely.com/full-stack/docs/decision-notification-listener#section-type-parameter) states it can be many different strings. I couldn't find any union of enum for that, but maybe I'm missing something. 😉


raju-opti:
closing this, the modified file does not exist anymore.

PR #695, opened 2021-07-26
feat: Use enum to define type in getTypeCastValue

yavorona:
## Summary
Use enum to define type in getTypeCastValue in project_config

### DIFF
--- a/packages/optimizely-sdk/lib/core/project_config/index.ts
+++ b/packages/optimizely-sdk/lib/core/project_config/index.ts
@@ -35,7 +35,8 @@ import {
   FeatureVariable,
   Variation,
   OptimizelyVariation,
-  VariationVariable,
+  VariableType,
+  VariationVariable
 } from '../../shared_types';
 
 interface TryCreatingProjectConfigConfig {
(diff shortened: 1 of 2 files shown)


packages/optimizely-sdk/lib/core/project_config/index.ts
```diff
@@ -582,9 +583,10 @@ export const getVariableValueForVariation = function(
  * @returns {*}                       Variable value of the appropriate type, or
  *                                    null if the type cast failed
  */
+
```
keppel2, line 586: Is this necessary?

packages/optimizely-sdk/lib/core/project_config/index.ts
```diff
@@ -203,7 +204,7 @@ export const createProjectConfig = function(
       // Converting it to a first-class json type while creating Project Config
       feature.variables.forEach((variable) => {
         if (variable.type === FEATURE_VARIABLE_TYPES.STRING && variable.subType === FEATURE_VARIABLE_TYPES.JSON) {
-          variable.type = FEATURE_VARIABLE_TYPES.JSON;
```
zashraf1985, line 206: why is this explicit cast needed for the `json` value but it works otherwise for other values without casting?

zashraf1985 (changes requested):

packages/optimizely-sdk/lib/core/project_config/index.ts
yavorona: I believe there is no otherwise cases of setting a value of `variable.type` besides this one. The FEATURE_VARIABLE_TYPES.JSON is type `string` and `variable.type` is expected to be `VariableType` so we needed to case it here

zashraf1985 (approved):
Looks Good!

PR #696, opened 2021-07-30
feat: Added React Native Client Engines to be sent in events

zashraf1985:
## Summary
Added Two new client engine values to be sent in events.

**react-native-sdk**: Client name will be set to `react-native-sdk` when a react native application will use react sdk i.e `@optimizely/react-sdk`.

**react-native-js-sdk**: Client name will be set to `react-native-js-sdk` when a react native application will use javascript sdk directly i.e `@optimizely/optimizely-sdk`.

### Client Versions
Client versions behave the way they did without any change. Current behaviour is.
1. When a react native app uses javascript SDK directly, clientVersion will be the version of javascript SDK. Which means `react-native-js-sdk` goes with `clientVersion` of javascript SDK.
2. When a react native app uses React SDK, clientVersion will be the version passed in by react SDK. This means `react-native-sdk` goes with `clientVersion` of react SDK. 

## Test plan
Made appropriate changes to already existing unit tests.

### DIFF
--- a/packages/optimizely-sdk/lib/index.react_native.tests.js
+++ b/packages/optimizely-sdk/lib/index.react_native.tests.js
@@ -92,7 +92,7 @@ describe('javascript-sdk/react-native', function() {
         assert.equal(optlyInstance.clientVersion, '4.6.2');
       });
 
-      it('should set the Javascript client engine and version', function() {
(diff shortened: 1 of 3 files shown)


yavorona (approved):
LGTM! We should just wait for travis to run prior to merging.

PR #704, opened 2021-09-18
Removed OptimizelyConfig creation from SDK initialization to reduce intialization time

zashraf1985:
## Summary
`OptimizelyConfig` object is created during initialization of `Optimizely` Instance. This almost doubles the initialization time for the SDK. This PR moves the creation of `OptimizelyConfig` object to `getOptimizelyConfig`. The `OptimizelyConfig` object will be created and cached when the `getOptimizelyConfig` API is called for the first time after every data file update.

## Test plan
- Manually tested thoroughly
- Updated unit tests
- All FSC test pass


yavorona (approved):
Great observation! It looks like it could be a great improvement in other sdks as well.

PR #715, opened 2021-11-01
Removed client engine validation

zashraf1985:
## Summary
Removed Client engine validation in the SDK to allow tracking events from new clients without friction.

## Test plan
- Manually Tested
- All tests pass


The-inside-man (approved):
LGTM!

PR #718, opened 2021-11-03
fix: impression event attributes exp, var should be empty instead of null

msohailhussain:
## Summary
- experimentId and variationId should be empty if no variation or experiment is found.

## Test plan
- Manually tested
- All tests pass

### DIFF
--- a/packages/event-processor/src/v1/buildEventV1.ts
+++ b/packages/event-processor/src/v1/buildEventV1.ts
@@ -146,8 +146,8 @@ function makeConversionSnapshot(conversion: ConversionEvent): Visitor.Snapshot {
 function makeDecisionSnapshot(event: ImpressionEvent): Visitor.Snapshot {
   const { layer, experiment, variation, ruleKey, flagKey, ruleType, enabled } = event
   let layerId = layer ? layer.id : null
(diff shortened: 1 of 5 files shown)


packages/event-processor/src/v1/buildEventV1.ts
```diff
@@ -146,8 +146,8 @@ function makeConversionSnapshot(conversion: ConversionEvent): Visitor.Snapshot {
 function makeDecisionSnapshot(event: ImpressionEvent): Visitor.Snapshot {
   const { layer, experiment, variation, ruleKey, flagKey, ruleType, enabled } = event
   let layerId = layer ? layer.id : null
-  let experimentId = experiment ? experiment.id : null
-  let variationId = variation ? variation.id : null
+  let experimentId = experiment ? (experiment.id ? experiment.id : '') : ''
+  let variationId = variation ? variation.id : ''
```
zashraf1985, line 150:
```suggestion
  let experimentId = experiment?.id ?? ''
  let variationId = variation?.id ?? ''
```

zashraf1985 (approved):
Approved with a NIT!

PR #729, opened 2021-12-14
fix: Checking variation in flagvariation map instead of checking it only in experiment

mnoman09:
## Summary
- Checking variation in flagvariation map for forced decision

### DIFF
--- a/packages/optimizely-sdk/lib/core/decision_service/index.ts
+++ b/packages/optimizely-sdk/lib/core/decision_service/index.ts
@@ -615,7 +615,11 @@ export class DecisionService {
           decideReasons.push(...decisionVariation.reasons);
           variationKey = decisionVariation.result;
           if (variationKey) {
-            const variation = getFlagVariationByKey(configObj, feature.key, variationKey);
+            let variation = null;
+            variation = experiment.variationKeyMap[variationKey];
+            if (!variation) {
+              variation = getFlagVariationByKey(configObj, feature.key, variationKey);
+            }
             variationForFeatureExperiment = {
               experiment: experiment,
               variation: variation,
--- a/packages/optimizely-sdk/lib/core/decision_service/index.ts
+++ b/packages/optimizely-sdk/lib/core/decision_service/index.ts
@@ -31,6 +31,7 @@ import {
   getExperimentAudienceConditions,
   getExperimentFromId,
   getExperimentFromKey,
+  getFlagVariationByKey,
   getTrafficAllocation,
   getVariationIdFromExperimentAndVariationKey,
   getVariationFromId,
@@ -614,7 +615,7 @@ export class DecisionService {
           decideReasons.push(...decisionVariation.reasons);
           variationKey = decisionVariation.result;
           if (variationKey) {
-            const variation = experiment.variationKeyMap[variationKey];
+            const variation = getFlagVariationByKey(configObj, feature.key, variationKey);
             variationForFeatureExperiment = {
               experiment: experiment,
               variation: variation,
@@ -1016,7 +1017,7 @@ export class DecisionService {
     const decideReasons: (string | number)[][] = [];
 
     // check forced decision first
-    const forcedDecisionResponse = user.findValidatedForcedDecision(flagKey, rule.key);
+    const forcedDecisionResponse = user.findValidatedForcedDecision(configObj, flagKey, rule.key);
     decideReasons.push(...forcedDecisionResponse.reasons);
 
     const forcedVariaton = forcedDecisionResponse.result;
@@ -1048,7 +1049,7 @@ export class DecisionService {
 
     // check forced decision first
     const rule = rules[ruleIndex];
-    const forcedDecisionResponse = user.findValidatedForcedDecision(flagKey, rule.key);
+    const forcedDecisionResponse = user.findValidatedForcedDecision(configObj, flagKey, rule.key);
     decideReasons.push(...forcedDecisionResponse.reasons);
 
     const forcedVariaton = forcedDecisionResponse.result;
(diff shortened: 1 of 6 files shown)


jaeopt (changes requested):
Change looks good. We need a unit test covering the case.

msohailhussain (approved):
lgtm. please address jae's comments.

packages/optimizely-sdk/lib/optimizely_user_context/index.ts
```diff
@@ -211,22 +215,24 @@ export default class OptimizelyUserContext {
...
 
     const decideReasons: (string | number)[][] = [];
     const forcedDecision = this.findForcedDecision({ flagKey, ruleKey });
     let variation = null;
     let variationKey;
-    if (forcedDecision) {
+    if (config && forcedDecision) {
       variationKey = forcedDecision.variationKey;
-      variation = this.optimizely.getFlagVariationByKey(flagKey, variationKey);
+      variation = getFlagVariationByKey(config, flagKey, variationKey);
```
zashraf1985, line 235: Why did this had to change? this looks like a weird pattern to pass in config from outside.

packages/optimizely-sdk/lib/optimizely_user_context/index.ts
```diff
@@ -211,22 +215,24 @@ export default class OptimizelyUserContext {
 
   /**
    * Finds a validated forced decision for specific flagKey and optional ruleKey.
+   * @param     {ProjectConfig}config               A projectConfig.
    * @param     {string}       flagKey              A flagKey.
    * @param     {ruleKey}      ruleKey              A ruleKey (optional).
    * @return    {DecisionResponse<Variation|null>}  DecisionResponse object containing valid variation object and decide reasons.
    */
   findValidatedForcedDecision(
+    config: ProjectConfig,
```
zashraf1985, line 224: This is a very weird pattern. why do we need to pass in config from outside when we have access to optimizely instance.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -1672,24 +1672,4 @@ export default class Optimizely {
...
-    }
-
-    const variations = configObj.flagVariationsMap[flagKey];
-    const result = find(variations, item => item.key === variationKey)
-    if (result) {
-      return result;
-    }
-
-    return null;
-  }
```
zashraf1985, line 1694: why remove this and move to `project_config` and pass that in separately?

packages/optimizely-sdk/lib/core/decision_service/index.ts
```diff
@@ -614,7 +615,11 @@ export class DecisionService {
           decideReasons.push(...decisionVariation.reasons);
           variationKey = decisionVariation.result;
           if (variationKey) {
-            const variation = experiment.variationKeyMap[variationKey];
+            let variation = null;
+            variation = experiment.variationKeyMap[variationKey];
+            if (!variation) {
+              variation = getFlagVariationByKey(configObj, feature.key, variationKey);
+            }
```
zashraf1985, line 622: As far as i understand, this was the only change needed here. i am unable to understand why were other changes made at all.

zashraf1985 (changes requested):
There is a weird pattern of passing in config from outside was introduced here which will make this `findValidatedForcedDecision` API unusable from outside. Please see if you can find a better way to do it or let me know, we can try to figure something out together

packages/optimizely-sdk/lib/optimizely/index.ts
msohailhussain: As per my understanding, `project_config` is used for mapping all datafile properties and their k/v mappings. Optimizely class is not for this purpose. I have not seen any single mapping in this class. That's why removed it and add it at proper location where you can see all other getter and setters.

packages/optimizely-sdk/lib/optimizely_user_context/index.ts
msohailhussain: optimizelyInstance.config may have different version if we try to access it from the optimizely instance. That's why the config we capture in the start of any API, we use it across all methods.

packages/optimizely-sdk/lib/core/decision_service/index.ts
msohailhussain: Secondly, `optimizely_instance`'s access is not available in decide_service, that's why moved to `project_config`

zashraf1985 (approved):
Based on @msohailhussain 's explanation and based on the fact that we have done the same in other SDKs, i am approving this. But i have serious concerns in general on the cyclic dependencies we are creating in the SDKs. This is definitely going to come back and bite us later. We need to think about redesigning without these cyclic dependencies later.
@jaeopt @dustin-sier @The-inside-man

packages/optimizely-sdk/lib/optimizely_user_context/index.tests.js
```diff
@@ -176,7 +176,39 @@ describe('lib/optimizely_user_context', function() {
...
+        });
+        user.setForcedDecision({ flagKey: flagKey, ruleKey }, { variationKey });
+        var decision = user.decide(flagKey, options);
+        sinon.assert.calledWithExactly(
+          fakeOptimizely.decide,
+          user,
+          flagKey,
+          options
+        );
+        assert.deepEqual(decision, fakeDecision);
```
jaeopt, line 209: It looks like checking fakeDecision vs fakeDecision, which is always true. Can you confirm?

jaeopt (changes requested):
Question about a test case. Can you check?

packages/optimizely-sdk/lib/optimizely_user_context/index.tests.js
mnoman09: you are right. It's my mistake I copied the test from above decide tests and in all those tests same thing is happening. I don't know why.

jaeopt (approved):
LGTM

PR #730, opened 2021-12-22
refac: findValidatedForcedDecision method moved to decision service

mnoman09:
## Summary
-  Moves findValidatedForcedDecision method to decision service as validatedForcedDecision

## Test plan
All unit test and FSC tests should pass.


packages/optimizely-sdk/lib/optimizely_user_context/index.ts
```diff
@@ -213,91 +213,6 @@ export default class OptimizelyUserContext {
...
-        }
-      }
-    }
-
-    return {
-      result: variation,
-      reasons: decideReasons,
-    }
-  }
-
```
ozayr-zaviar, line 300: import logger on line 16 and `get_logger` on line 35 can be removed.

msohailhussain (commented):
lgtm

jaeopt (approved):
LGTM

jaeopt (approved):
LGTM

packages/optimizely-sdk/lib/core/decision_service/index.ts
```diff
@@ -776,6 +776,98 @@ export class DecisionService {
...
+    flagKey: string,
+    ruleKey?: string
+  ): DecisionResponse<Variation | null> {
+
+    const decideReasons: (string | number)[][] = [];
+    const forcedDecision = user.getForcedDecision({ flagKey, ruleKey });
+    let variation = null;
+    let variationKey;
+    const userId = user.getUserId()
+    if (config && forcedDecision) {
```
zashraf1985, line 799: Suggestion (non binding): If we move the check for `config` at the top, we will save a redundant call to `getForcedDecision` when config is not available.

zashraf1985 (approved):
LGTM! with a non binding suggestion.

PR #742, opened 2022-02-09
chore: removed unused npm package from root

zashraf1985:
## Summary
Javascript SDK is a monorepo which contains the javascript SDK and supporting packages in `packages` subfolder. We initially had a parent npm package at root to manage subpackages together but its no longer being used. This PR removed that unused npm package and updates readme.

## Test plan
All existing tests should pass


opti-jnguyen (approved):
LGTM!

PR #745, opened 2022-02-25
refactor: Replaced explicit typings with auto generated ones

ozayr-zaviar:
## Summary
This SDK was originally javascript only. We had to add explicit typings in `lib/index.d.ts` to support typescript users. Now that it has been converted to typescript, we can safely remove `lib/index.d.ts` and autogenerate typings. This PR does exactly that. We also encountered some discrepancies with existing typings so we fixed those too. 

## Test plan
- Manually tested thoroughly with typescript projects
- All existing tests pass.


packages/optimizely-sdk/lib/export_types.ts
```diff
@@ -0,0 +1,27 @@
...
+  OptimizelyExperiment,
+  OptimizelyFeature,
+  OptimizelyDecisionContext,
+  OptimizelyForcedDecision,
+  EventTags,
+  Event,
+  EventDispatcher,
+  DatafileOptions,
+  SDKOptions,
+  OptimizelyOptions,
```
zashraf1985, line 15: SDKOptions and OptimizelyOptions are not needed to be exposed. Config is enough for external usage

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -263,12 +277,105 @@ export interface OptimizelyVariable {
...
+export interface TrackListenerPayload extends ListenerPayload {
+  eventKey: string;
+  eventTags: EventTags;
+  logEvent: Event;
+}
+
 /**
  * Entry level Config Entities
  */
 export interface SDKOptions {
```
zashraf1985, line 374: Can you compare it with `Config` and remove it if not needed. this looks redundant and might not be needed anymore.

zashraf1985 (changes requested):
This looks good so far. Requested a few more changes

zashraf1985 (approved):
LGMT! Great work @ozayr-zaviar

zashraf1985:
@opti-jnguyen It will be great if we can get your review on this whenever you find some time.

packages/optimizely-sdk/lib/plugins/datafile_manager/http_polling_datafile_manager.ts
```diff
@@ -21,8 +21,10 @@ import fns from '../../utils/fns';
 
 export function createHttpPollingDatafileManager(
   sdkKey: string,
-  logger: LoggerFacade,  
-  datafile?: string,
+  logger: LoggerFacade,
+  // TODO[OASIS-6649]: Don't use object type
+  // eslint-disable-next-line  @typescript-eslint/ban-types
```
opti-jnguyen, line 26: Were these comments left here intentionally?

zashraf1985: https://optimizely.atlassian.net/browse/OASIS-6649
This should be moved to the tech debt epic. This is something thats tricky to resolve

packages/optimizely-sdk/lib/shared_types.ts
opti-jnguyen: Config and SDKOptions shares most of the same variables except for:

Config:

1. Extra `jsonSchemaValidator` variable in interface.
2. Instead of using `LogLevel | string`, uses union type on all LOG_LEVEL enums.

SDKOptions contains:

1. Extra `eventMaxQueueSize` variable in interface.
2. Extra `isValidInstance` variable in interface. 

---

We should consider having a base interface both `Config` and `SDKOptions` extend.

zashraf1985 (changes requested):
@ozayr-zaviar I have a few more change request.
1. `Config` should also export a jsonSchemaValidator just like `OptimizelyOptions`
2. `isValidInstance` is not logically needed on `Config`. We were just using `config` to hold this value and pass on to `Optimizely` constructor instead of storing it in to a local variable in the entry point. Please fix this.
3. Update copyright headers wherever applicable.

packages/optimizely-sdk/lib/shared_types.ts
zashraf1985: thanks for pointing these out. WE are consolidating these two types in to one because `Config` was exposed from the explicit typings while `SDKOptions` was being used in the actual code. We need to use one of those now because we are removing the explicit types. We will keep `Config` and remove `SDKOptions` because thats what was being exposed to typescript users. I have requested @ozayr-zaviar to add the missing variables mentioned in your comments.

zashraf1985 (approved):
LGTM! Thanks for cleaning up the typings mess. Great work @ozayr-zaviar

packages/optimizely-sdk/lib/index.d.ts
```diff
@@ -1,242 +0,0 @@
...
-  import * as enums from '@optimizely/optimizely-sdk/lib/utils/enums';
-  import * as logging from '@optimizely/optimizely-sdk/lib/plugins/logger';
-
-  export { enums, logging };
-
-  export function setLogger(logger: LogHandler | null): void;
-
-  export function setLogLevel(level: enums.LOG_LEVEL | string): void;
-
-  export function createInstance(config: Config): Client;
```
Develliot, line 28: This is a breaking change

PR #747, opened 2022-03-21
feat: addNotificationListener exported

ozayr-zaviar:
## Summary
- addNotification listener type exported

## Test plan
- All testcases and FSC should pass

### DIFF
--- a/packages/utils/src/index.ts
+++ b/packages/utils/src/index.ts
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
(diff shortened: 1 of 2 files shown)


packages/utils/src/index.ts
```diff
@@ -154,4 +155,8 @@ export enum NOTIFICATION_TYPES {
 
 export interface NotificationCenter {
   sendNotifications(notificationType: NOTIFICATION_TYPES, notificationData?: any): void
+  addNotificationListener<T extends ListenerPayload>(
```
zashraf1985, line 158: What if we simply remove this NotificationCenter from here and export this whole thing from `optimizely-sdk` package?

zashraf1985 (changes requested):

PR #763, opened 2022-05-27
refactor: removed redundant type definitions for datafile-manager package

shaharyar123:
## Summary
Removed redundant type definitions for datafile-manager package.

## Test Plan
- All existing tests pass


packages/optimizely-sdk/lib/plugins/datafile_manager/http_polling_datafile_manager.ts
```diff
@@ -25,8 +25,8 @@ export function createHttpPollingDatafileManager(
   // TODO[OASIS-6649]: Don't use object type
   // eslint-disable-next-line  @typescript-eslint/ban-types
   datafile?: string | object,
-  datafileOptions?: DatafileOptions,
-): DatafileManager {  
+  datafileOptions?: DatafileOptions
```
zashraf1985, line 28: This whole file change is just the whitespace change, can you revert it so that its removed from diff.

zashraf1985 (changes requested):

PR #775, opened 2022-08-05
ODP Datafile Parsing

opti-jnguyen:
## Summary
- Parses new datafile odp values (“publicKey” and “host”) from the “integrations” section. These values are used to send requests to the ODP server.
- Parses datafile with and without the “integrations” section or empty section as well.

---

## Technical Changes
- Added new `integrations` field to the `project_config_schema` composed of array items which are objects with the keys of `key`, `host`, and `publicKey`.
- Added new `Integrations` 
- Added new fields to `ProjectConfig` interface to prepare catching this new incoming `integrations` field:
  - `integrations`
  - `integrationKeyMap` (Optional)
  - `publicKeyForOdp` (Optional)
  - `hostForOdp` (Optional)
  - `allSegments`
- Exports new `getAudienceSegments` function from project_config. This function is used for recursively evaluating through audience conditions qualified matches, returning a concatenated array of all condition values where the condition `match` is equal to the value `qualified` .

## Test plan
- Tests cover project config creation with new test datafiles and check for new allSegments, publicKeyForOdp, and hostForOdp variables.
- Tests check that evaluation of datafiles with and without ODP integrations work.
- Tests check that evaluation of datafiles with and without segments work.
- Tests check that `getAudienceSegments` works as a standalone function.

## Issues
- OASIS-8390

---

Note: Prettier caused many whitespace changes to be applied. To see only the related changes, click the gear icon and select "Hide whitespace changes" in the diff.

### DIFF
--- a/packages/optimizely-sdk/lib/core/project_config/index.tests.js
+++ b/packages/optimizely-sdk/lib/core/project_config/index.tests.js
@@ -786,7 +786,7 @@ describe('lib/core/project_config', function () {
                   "value": "odp-segment-1",
                   "type": "third_party_dimension",
                   "name": "odp.audiences",
-                  "match": "not-qualified"
+                  "match": "invalid"
                 }
               ]
             ]
@@ -824,7 +824,7 @@ describe('lib/core/project_config', function () {
 
       it('should convert integrations from the datafile into the project config', () => {
         assert.exists(config.integrations);
-        assert.equal(config.integrations.length, 1);
+        assert.equal(config.integrations.length, 3);
       });
 
       it('should populate the public key value from the odp integration', () => {
@@ -835,9 +835,9 @@ describe('lib/core/project_config', function () {
         assert.exists(config.hostForOdp)
       })
 
-      it('should contain all expected odp segments in allSegments', () => {
-        assert.equal(config.allSegments.length, 3)
-        assert.deepEqual(config.allSegments, ['odp-segment-1', 'odp-segment-2', 'odp-segment-3'])
+      it('should contain all expected unique odp segments in allSegments', () => {
+        assert.equal(config.allSegments.size, 3)
+        assert.deepEqual(config.allSegments, new Set(['odp-segment-1', 'odp-segment-2', 'odp-segment-3']))
       })
     });
 
@@ -849,7 +849,7 @@ describe('lib/core/project_config', function () {
 
       it('should convert integrations from the datafile into the project config', () => {
         assert.exists(config.integrations);
-        assert.equal(config.integrations.length, 1);
+        assert.equal(config.integrations.length, 3);
       });
 
       it('should populate the public key value from the odp integration', () => {
@@ -862,8 +862,8 @@ describe('lib/core/project_config', function () {
         assert.equal(config.hostForOdp, 'https://api.zaius.com')
       })
 
-      it('should contain all expected odp segments in all segments', () => {
-        assert.equal(config.allSegments.length, 0)
+      it('should contain all expected unique odp segments in all segments', () => {
+        assert.equal(config.allSegments.size, 0)
       })
     });
 
--- a/packages/optimizely-sdk/lib/core/project_config/index.tests.js
+++ b/packages/optimizely-sdk/lib/core/project_config/index.tests.js
@@ -33,13 +33,13 @@ import configValidator from '../../utils/config_validator';
 var buildLogMessageFromArgs = args => sprintf(args[1], ...args.splice(2));
 var logger = getLogger();
 
-describe('lib/core/project_config', function() {
-  describe('createProjectConfig method', function() {
-    it('should set properties correctly when createProjectConfig is called', function() {
+describe('lib/core/project_config', function () {
+  describe('createProjectConfig method', function () {
+    it('should set properties correctly when createProjectConfig is called', function () {
       var testData = testDatafile.getTestProjectConfig();
       var configObj = projectConfig.createProjectConfig(testData);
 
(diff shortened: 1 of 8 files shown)


packages/optimizely-sdk/lib/core/project_config/index.ts
```diff
@@ -161,6 +167,12 @@ export const createProjectConfig = function(
   projectConfig.audiencesById = keyBy(projectConfig.audiences, 'id');
   assign(projectConfig.audiencesById, keyBy(projectConfig.typedAudiences, 'id'));
 
+  if (!projectConfig.allSegments) projectConfig.allSegments = []
+
+  Object.keys(projectConfig.audiencesById).forEach((audience) => {
+    projectConfig.allSegments = projectConfig.allSegments.concat(getAudienceSegments(projectConfig.audiencesById[audience]))
```
jaeopt, line 173: Can we filter out redundant segments from multiple audiences?

packages/optimizely-sdk/lib/core/project_config/index.tests.js
```diff
@@ -727,20 +727,84 @@ describe('lib/core/project_config', function() {
...
+            "and",
+            [
+              "or",
+              [
+                "or",
+                {
+                  "value": "odp-segment-1",
+                  "type": "third_party_dimension",
+                  "name": "odp.audiences",
+                  "match": "not-qualified"
```
jaeopt, line 789: This test is a little strange. We do not have "not-qualified" match type. If intentional error testing, it'll be more clear with a different name to avoid confusion like "invalid".

packages/optimizely-sdk/lib/tests/test_data.js
```diff
@@ -3141,10 +3141,230 @@ var typedAudiencesConfig = {
...
+    }
+  ],
+  "groups": [],
+  "integrations": [
+    {
+      "key": "odp",
+      "host": "https://api.zaius.com",
+      "publicKey": "W4WzcEs-ABgXorzY7h1LCQ"
+    }
+  ],
```
jaeopt, line 3259: Let's have one more test case like `{key = odp, a = 1, b = true}, {key=x, a=y}` to validate it can tolerate future changes ("key" is the only field required).

packages/optimizely-sdk/lib/core/project_config/index.ts
jaeopt: Add a test case for this redundant segments.

jaeopt (changes requested):
Looks good! A couple of changes suggested.

packages/optimizely-sdk/lib/core/project_config/index.ts
```diff
@@ -98,6 +99,11 @@ export interface ProjectConfig {
   accountId: string;
   flagRulesMap: { [key: string]: Experiment[] };
   flagVariationsMap: { [key: string]: Variation[] };
+  integrations: Integration[];
+  integrationKeyMap?: { [key: string]: Integration };
+  publicKeyForOdp?: string;
+  hostForOdp?: string;
+  allSegments: string[];
```
zashraf1985, line 106: you could use a set here to avoid duplicates

packages/optimizely-sdk/lib/core/project_config/index.ts
zashraf1985, line 173: This might contain duplicates, if you use set it will take care of it automatically i believe.

zashraf1985 (changes requested):
Looks Great! Suggested one change.

packages/optimizely-sdk/lib/core/project_config/index.ts
opti-jnguyen: Resolved here: https://github.com/optimizely/javascript-sdk/pull/775/commits/baa63ea428de6ff92fb8d98ccd7cae09caac3864

packages/optimizely-sdk/lib/core/project_config/index.tests.js
opti-jnguyen: Resolved here: https://github.com/optimizely/javascript-sdk/pull/775/commits/baa63ea428de6ff92fb8d98ccd7cae09caac3864

packages/optimizely-sdk/lib/tests/test_data.js
opti-jnguyen: Resolved here: https://github.com/optimizely/javascript-sdk/pull/775/commits/baa63ea428de6ff92fb8d98ccd7cae09caac3864

packages/optimizely-sdk/lib/core/project_config/index.ts
opti-jnguyen: Resolved here: https://github.com/optimizely/javascript-sdk/pull/775/commits/baa63ea428de6ff92fb8d98ccd7cae09caac3864

packages/optimizely-sdk/lib/core/project_config/index.ts
opti-jnguyen: Resolved here: https://github.com/optimizely/javascript-sdk/pull/775/commits/baa63ea428de6ff92fb8d98ccd7cae09caac3864

opti-jnguyen:
@jaeopt , @zashraf1985 - updated to resolve your comments above. Let me know if the solution looks good to you or if I missed anything... thanks!

jaeopt (approved):
LGTM

zashraf1985 (approved):
Looks Great!

PR #776, opened 2022-08-15
feat: Add ODP/ATS VuidManager

mikechu-optimizely:
## Summary
Full Stack client should create and manage a unique device or visitor ID (VUID), persistent and shared between multiple SDK instances.

## Test plan
Added new unit test spec files to base package

## Issues
OASIS-8412


packages/optimizely-sdk/tests/vuidManager.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  it('should make a VUID', async () => {
+    // arrange
+    const manager = await VuidManager.instance(cache);
+
+    // act
+    const vuid = manager.makeVuid();
+
+    // assert
+    expect(vuid.startsWith('vuid_')).toBe(true);
+    expect(vuid.length).toBeGreaterThan(20);
```
jaeopt, line 39: Can check equalTo(32)

packages/optimizely-sdk/tests/vuidManager.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    expect(manager1.isVuid(vuid2)).toBeTruthy();
+
+    await cache.remove("optimizely-odp");
+
+    // should end up being a new instance since we just removed it above
+    await manager2.load(cache);
+    const vuid3 = manager2.vuid;
+
+    expect(vuid3).not.toStrictEqual(vuid1);
+  });
```
jaeopt, line 71: expect(manager2.isVuid(vuid3)).toBeTruthy();

jaeopt (changes requested):
Looks good. A couple of suggestions.
I see some JS codes I cannot cover.

mikechu-optimizely:
This PR remains ready for @zashraf1985 's review as well.

jaeopt (approved):
LGTM

packages/optimizely-sdk/lib/plugins/key_value_cache/browserAsyncStorageCache.ts
```diff
@@ -0,0 +1,42 @@
...
+import PersistentKeyValueCache from './persistentKeyValueCache';
+
+export default class BrowserAsyncStorageCache implements PersistentKeyValueCache {
+  contains(key: string): Promise<boolean> {
+    return Promise.resolve( localStorage.getItem(key) !== null);
+  }
+
+  get(key: string): Promise<string> {
+    const item = localStorage.getItem(key);
+    return item === null ? Promise.resolve().then(() => '') : Promise.resolve(item);
```
zashraf1985, line 26:
```suggestion
    return Promise.resolve(item || null);
```

packages/optimizely-sdk/lib/plugins/key_value_cache/inMemoryAsyncStorageCache.ts
```diff
@@ -0,0 +1,93 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import PersistentKeyValueCache from './persistentKeyValueCache';
+
+/**
+ * An in-memory singleton string cache, supporting strings as keys and value.
+ */
+export default class InMemoryAsyncStorageCache implements PersistentKeyValueCache {
```
zashraf1985, line 22: Why do we need this in memory? If we are looking for an in-memory cache for some other scenario, we should keep it synchronized. Forcing it to be async does not look like a good pattern. You are implementing a `PersistentKeyValueCache` as an `in-memory` which is a contradiction in itself. If you want both persistent and in-memory for different use cases, they should be different interfaces and implementations, one with promises and one synchronous

packages/optimizely-sdk/lib/plugins/key_value_cache/reactNativeAsyncStorageCache.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export default class ReactNativeAsyncStorageCache implements PersistentKeyValueCache {
+  async contains(key: string): Promise<boolean> {
+    const val = await AsyncStorage.getItem(key);
+    return val !== null;
+  }
+
+  async get(key: string): Promise<string> {
+    const val =  await AsyncStorage.getItem(key);
+    return val ? val : '';
```
zashraf1985, line 28: why are you returning empty string if async storage returns a falsy value? Shouldn't it return null in this case because it implies that key doesn't exist int he cache

packages/optimizely-sdk/lib/plugins/vuid_manager/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    const cachedValue = await cache.get(this._keyForVuidMap);
+    if (cachedValue) {
+      const dict = cachedValue as Map<string, string>;
+      if (typeof dict?.has === "function" && dict?.has(this._keyForVuid)) {
+        const oldVuid = dict.get(this._keyForVuid);
+        if (oldVuid) {
+          validVuid = oldVuid;
+        }
+      }
+    }
```
zashraf1985, line 71: Why are you getting `cachedValue` as dict. Shouldn't it be a `String` or `null` i.e `vuid` value directly?

packages/optimizely-sdk/lib/plugins/vuid_manager/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    // make sure UUIDv4 is used (not UUIDv1 or UUIDv6) since the trailing 5 chars will be truncated. See TDD for details.
+    const uuidV4 = uuid();
+    const formatted = uuidV4.replace(/-/g, '').toLowerCase();
+    const vuidFull = `${this._prefix}${formatted}`;
+
+    return (vuidFull.length <= maxLength) ? vuidFull : vuidFull.substring(0, maxLength);
+  }
+
+  public async save(vuid: string, cache: PersistentKeyValueCache): Promise<void> {
+    const dict = new Map<string, string>([[this._keyForVuid, vuid]]);
```
zashraf1985, line 95: Why do you need to create a map and save it. Why not simply use the cache as a key value store and set and get string values for vuid directly?

zashraf1985 (changes requested):

packages/optimizely-sdk/lib/plugins/vuid_manager/index.ts
mikechu-optimizely: @jaeopt Do we need to cache a dictionary of VUIDs or just the vuid key-val?

jaeopt: Can we keep the dictionary as is? Originally, we had other values in the dictionary (like a list of users bound with the vuid). It turns out that those values are not needed persistently, so removed. It looks like not clean to have the dict to keep the vuid only, but we may need to add props in future.

zashraf1985: Thanks @jaeopt. Instead of serializing/deserializing and managing the whole dictionary ourselves, do you think it makes sense to use the cace prefix to differentiate the items from others. For example we can prefix all the entries for this cache with `vuid-manager-`. In this way, we will still be using a primary key value storage instead of pulling a value manually from our own custom dictionary which can grow large and cause potential performance issues. What do you guys think?

jaeopt: Yeah, using prefix makes sense to simplify the access and also can be extended. I suggest to use "optimizely-" for the prefix for grouping all keys for us. Let's change it to a flat key-value with "optimizely-vuid" as a key in this case. I'll fix swift-sdk for consistency.

jaeopt: @zashraf1985 @mikechu-optimizely

mikechu-optimizely: Oops. I missed this directive.  I will adjust my code and tests.

packages/optimizely-sdk/tests/vuidManager.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// <reference types="jest" />
+
+import { VuidManager } from '../lib/plugins/vuid_manager';
+import PersistentKeyValueCache from '../lib/plugins/key_value_cache/persistentKeyValueCache';
+import { anyString, anything, instance, mock, resetCalls, verify, when } from 'ts-mockito';
```
zashraf1985, line 21: This is good!

zashraf1985 (approved):
Looks Great!

PR #777, opened 2022-08-19
feat: Add ODP LRU Cache

opti-jnguyen:
## Summary
A generic LRU cache to support efficient caching of the ODP segments results.

## Test plan
Added new unit tests files for Client/Server/React Native implementations.

## Issues
OASIS-8410


opti-jnguyen:
Originally created React Native specific variations for AsyncStorage consideration, but since we're only concerned with an in-memory LRU Cache, will remove LRU Cache code specific to React Native.

packages/optimizely-sdk/lib/core/odp/lru_cache/LRUCacheElement.ts
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * LRUCacheElement represents an individual generic item within the LRUCache
+ */
+export class LRUCacheElement<V> {
```
jaeopt, line 20:
```suggestion
export class CacheElement<V> {
```
This element is a generic cache element, not only for LRU.

packages/optimizely-sdk/lib/core/odp/lru_cache/LRUCache.ts
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Least-Recently Used Cache (LRU Cache) Implementation with Generic Key-Value Pairs
+ * Analogous to a Map that has a specified max size and a timeout per element.
+ * - Removes the least-recently used element from the cache if max size exceeded.
+ * - Removes stale elements (entries older than their timeout) from the cache.
+ */
+export class LRUCache<K, V> {
+    private _map: Map<K, LRUCacheElement<V>> = new Map()
+    private _maxSize = 100 // Defines maximum size of _map
+    private _timeout = 1000 * 600 // Milliseconds each entry has before it becomes stale
```
jaeopt, line 28: Default cache size depends on that it's for server (node) or client (browser). If we want to feed them per env, can we consider to remove this default value here?

packages/optimizely-sdk/lib/core/odp/lru_cache/LRUCache.ts
```diff
@@ -0,0 +1,125 @@
...
+     */
+    public lookup(key: K): V | null {
+        if (this._maxSize <= 0) { return null }
+
+        const element: LRUCacheElement<V> | undefined = this._map.get(key)
+
+        if (!element) return null
+
+        this._map.delete(key)
+        this._map.set(key, element)
```
jaeopt, line 51: Can we move this adding back to cache after "is_stale" check below?  We do not need to keep staled item.

packages/optimizely-sdk/lib/core/odp/lru_cache/LRUCache.ts
```diff
@@ -0,0 +1,125 @@
...
+     */
+    public reset(): void {
+        if (this._maxSize <= 0) return
+
+        this._map.clear()
+    }
+
+    /**
+     * Reads value from specified key without moving elements in the LRU Cache.
+     * Removes the element if it is stale.
```
jaeopt, line 90: I think peek will be more useful without side effect. Can we remove the stale clean up?

packages/optimizely-sdk/lib/core/odp/lru_cache/LRUCache.ts
```diff
@@ -0,0 +1,125 @@
...
+     */
+    public peek(key: K): V | null {
+        if (this._maxSize <= 0) return null
+
+        const element: LRUCacheElement<V> | undefined = this._map.get(key)
+
+        if (element?.is_stale(this._timeout)) {
+            this._map.delete(key)
+            return null
+        }
```
jaeopt, line 101: Remove?

jaeopt (changes requested):
Looks good! A few small changes suggested.

packages/optimizely-sdk/lib/core/odp/lru_cache/LRUCache.ts
opti-jnguyen: @jaeopt - Main reason this was added because I noticed when using `peek()` at elements that should be stale returned a value.

packages/optimizely-sdk/lib/core/odp/lru_cache/LRUCache.ts
opti-jnguyen: @jaeopt - Should `peek()` return null if stale?

opti-jnguyen: ^ Resolved - will remove completely.

jaeopt (approved):
LGTM

PR #778, opened 2022-08-22
feat: Add ODP GraphQL AP Interface

mikechu-optimizely:
## Summary
Add plugin to provide an internal service for ODP GraphQL server access. Should support 3 potential clients for Browser, Node, ReactNative

## Test plan
New .spec. files added unit tests

## Issues
OASIS-8409


packages/optimizely-sdk/lib/plugins/odp/graphql_manager.ts
```diff
@@ -0,0 +1,89 @@
...
+      this._logger.log(LogLevel.WARNING, `Audience segments fetch failed (${errors})`);
+
+      return [] as string[];
+    }
+
+    if (parsedSegments?.data?.customer?.audiences?.edges === null) {
+      this._logger.log(LogLevel.WARNING, 'Audience segments fetch failed (decode error)');
+
+      return [] as string[];
+    }
```
zashraf1985, line 71: If you invert this check, it will work better. The current check will only log error if the whole path leading up to the `edges` is present. If anything is missing in between, it will consider it a success response and will try to parse it after the check and crash. You could probably check for negation and invert it so it will cover the null cases too

packages/optimizely-sdk/lib/plugins/odp/graphql_manager.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    return parsedSegments.data.customer.audiences.edges.filter(edge => edge.node.state == QUALIFIED).map(edge => edge.node.name);
+  }
+
+  private parseSegmentsResponseJson(jsonResponse: string): Response | undefined {
+    let jsonObject = {};
+    try {
+      jsonObject = JSON.parse(jsonResponse);
+    } catch {
```
zashraf1985, line 80: Can you catch the actuall error and print its details to the logs

packages/optimizely-sdk/lib/plugins/odp/odp_client.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  public async querySegments(parameters: QuerySegmentsParameters): Promise<string | undefined> {
+    const emptyResponse = undefined;
+
+    if (!parameters?.ApiHost || !parameters?.ApiKey) {
+      this._logger.log(LogLevel.ERROR, 'No ApiHost or ApiKey set before querying segments');
+      return emptyResponse;
+    }
+
+    const fetchFailureMessage = 'Audience segments fetch failed';
```
zashraf1985, line 39: This can be moved near the top of the file outside the function.

packages/optimizely-sdk/lib/plugins/odp/odp_client.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export class OdpClient implements IOdpClient {
+  private readonly _logger: LogHandler;
+
+  constructor(logger: LogHandler) {
+    this._logger = logger;
+  }
+
+  public async querySegments(parameters: QuerySegmentsParameters): Promise<string | undefined> {
+    const emptyResponse = undefined;
```
zashraf1985, line 32: lets make it null. And you also dont need to define it. Simply return null explicitly where you want. that will be more readable

packages/optimizely-sdk/lib/plugins/odp/query_segments_parameters.ts
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+/**
+ * Handles parameters used in querying ODP segments
+ */
+export class QuerySegmentsParameters {
+  /**
+   * Optimizely Data Platform API key
+   */
+  public ApiKey: string | undefined;
```
zashraf1985, line 24: javascript convention is camelCase for the vars. It would be apiKey. Same applies for other members of this class.

packages/optimizely-sdk/lib/plugins/odp/query_segments_parameters.ts
```diff
@@ -0,0 +1,66 @@
...
+   * Value for the user key
+   */
+  public UserValue: string | undefined;
+
+  /**
+   * Audience segments to check for inclusion in the experiment
+   */
+  public SegmentToCheck: string[] | undefined;
+
+  constructor(parameters: { UserValue: string; ApiKey: string; UserKey: string; SegmentsToCheck: string[]; ApiHost: string }) {
```
zashraf1985, line 46: variable names should be camelCase

packages/optimizely-sdk/lib/plugins/odp/query_segments_parameters.ts
```diff
@@ -0,0 +1,66 @@
...
+    Object.assign(this, parameters);
+  }
+
+  /**
+   * Converts the QuerySegmentsParameters into JSON
+   * @returns GraphQL JSON payload
+   */
+  public ToJson(): string {
```
zashraf1985, line 54: this function name should also be cameCase

zashraf1985, line 54: This is not converting it in json. It's probably converting it into the format expected by the graphQL call. If that is the case, then i will suggest changing the name of method.

packages/optimizely-sdk/lib/utils/json_schema_validator/index.ts
```diff
@@ -23,23 +23,32 @@ const MODULE_NAME = 'JSON_SCHEMA_VALIDATOR';
...
  * Validate the given json object against the specified schema
- * @param  {unknown} jsonObject The object to validate against the schema
- * @return {boolean}            true if the given object is valid
+ * @param {unknown} jsonObject The object to validate against the schema
+ * @param {JSONSchema4} validationSchema Provided schema to use for validation
+ * @param {boolean} shouldThrow Should validation throw if invalid JSON object
+ * @return {boolean} true if the given object is valid; throws or false if invalid
  */
-export function validate(jsonObject: unknown): boolean {
+export function validate(jsonObject: unknown, validationSchema: JSONSchema4 = schema, shouldThrow = true): boolean {
```
zashraf1985, line 31: NIT: shouldThrowOnError

zashraf1985 (changes requested):
Looks great! Suggested a few changes. I assume `fetch` should work in all environments but it will be great if you can check it once for browser, node and react native.

packages/optimizely-sdk/lib/plugins/odp/graphql_manager.ts
zashraf1985: I am assuming `?` fails with undefined. If it fails consistently with `null` then it should work. A safe check would be
```if (!arsedSegments?.data?.customer?.audiences?.edges)```

Comment (mikechu-optimizely) [reply packages/optimizely-sdk/lib/plugins/odp/graphql_manager.ts:71]: Yup. [Optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html) will stop on nullish (null or undefined) expressions.

> At its core, optional chaining lets us write code where TypeScript can immediately stop running some expressions if we run into a null or undefined.

I'll invert the pattern as directed.

Comment (mikechu-optimizely) [reply packages/optimizely-sdk/lib/plugins/odp/graphql_manager.ts:71]: I also see my check for `=== null` would not eval to true since optional chaining returns `undefined`

File: packages/optimizely-sdk/lib/plugins/odp/odp_client.ts (hunk 4 above)
Comment (mikechu-optimizely) [reply packages/optimizely-sdk/lib/plugins/odp/odp_client.ts:32]: I tend to like self-documenting code which is why I declared and used a `const` here.   This gives the reader the understanding that I'm intentionally returning what amounts to a "no response".  

After further reading, I agree that the union type should have a `null` instead of `undefined`.  `undefined` gives the sense that nothing's been set yet, while `null` provides knowledge of an intentional "empty" return.  I'll alter this through the code in this PR.

**Review by zashraf1985 — APPROVED** [review]

File: packages/optimizely-sdk/lib/plugins/odp/graphql_manager.ts [hunk 9]
```diff
@@ -117,8 +117,7 @@ export class GraphqlManager implements IGraphQLManager {
 
     try {
       jsonObject = JSON.parse(jsonResponse);
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    } catch (error: any) {
```
opti-jnguyen: Looks like removing this is causing CI/CD to throw a linting error?

packages/optimizely-sdk/lib/utils/http_request_handler/request_handler_factory.ts
```diff
@@ -0,0 +1,35 @@
...
+import { RequestHandler } from './http';
+import { NodeRequestHandler } from './node_request_handler';
+import { BrowserRequestHandler } from './browser_request_handler';
+
+/**
+ * Factory to create the appropriate type of RequestHandler based on a provided context
+ */
+export class RequestHandlerFactory {
+  public static createHandler(type: string, logger?: LogHandler, timeout?: number): RequestHandler {
+    if (type === 'node') {
```
zashraf1985, line 27: These "node" and "browser" can go as enums

zashraf1985 (changes requested):
Looks Great! Just added a few suggestions.

zashraf1985 (approved):

PR #785, opened 2022-09-12
feat: Added ODP RestApi interface for sending Events to ODP

mikechu-optimizely:
## Summary
This module provides an internal service for ODP event REST API access.

## Test plan

- Added new unit tests to .spec
- All Existing Unit tests pass.
- All existing full-stack compatibility tests pass.

## Issues
- FSSDK-8386 (OASIS-8386 [to pass jira_ticket_reference_check])


packages/optimizely-sdk/tests/odpClient.spec.ts
```diff
@@ -77,167 +45,383 @@ describe('OdpClient', () => {
...
+                  'name': 'has_email',
+                  'state': 'qualified',
+                },
+              },
+              {
+                'node': {
+                  'name': 'has_email_opted_in',
+                  'state': 'qualified',
+                },
+              },
```
jaeopt, line 76: It'll be a better test if we have a node with 'state': 'not-ready' as well.

packages/optimizely-sdk/tests/odpClient.spec.ts
```diff
@@ -18,46 +18,14 @@
 
 import { anyString, anything, instance, mock, resetCalls, verify, when } from 'ts-mockito';
 import { ErrorHandler, LogHandler, LogLevel } from '../lib/modules/logging';
-import { OdpClient } from '../lib/plugins/odp/odp_client';
+import { OdpClient, RETRY_ADVISED_BUT_NO_HTTP_STATUS_AVAILABLE } from '../lib/plugins/odp/odp_client';
 import { QuerySegmentsParameters } from '../lib/plugins/odp/query_segments_parameters';
 import { BrowserRequestHandler } from '../lib/utils/http_request_handler/browser_request_handler';
 import { NodeRequestHandler } from '../lib/utils/http_request_handler/node_request_handler';
+import { SendEventsParameters } from '../lib/plugins/odp/send_events_parameters';
+import { OdpEvent } from '../lib/plugins/odp/odp_event';
```
jaeopt, line 26: Can we splint tests into 2 files segmentFetch and odpEvent? In many cases, they should handle errors and responses in different ways. It'll be easier to follow if we split them.

jaeopt (changes requested):
Looks good!
A few changes suggested. I'll review test cases again after splitting them into 2 groups.

packages/optimizely-sdk/lib/plugins/odp/odp_client.ts
```diff
@@ -84,11 +99,49 @@ export class OdpClient implements IOdpClient {
...
+   * @returns
+   * 1. null, When there was a non-recoverable error and no retry is needed.
+   * 2. 0 If an unexpected error occurred and retrying can be useful.
+   * 3. HTTPStatus code if httpclient was able to make the request and was able to receive response.
+   *    It is recommended to retry if status code was 5xx.
+   */
+  public async sendOdpEvents(parameters: SendEventsParameters): Promise<number | null> {
+    if (!parameters?.apiEndpoint || !parameters?.apiKey) {
+      this._logger.log(LogLevel.ERROR, 'No ApiEndpoint or ApiKey set before attempting to send ODP events');
+      return RETRY_NOT_ADVISED;
```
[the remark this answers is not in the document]
mikechu-optimizely: Ah. Ok. I will typically create a constant, so other developers know my intention without reading the method doc or internal comments. I will export the constant too, so my tests are looking for the same value. 

I will revert to using an explicit `null` and remove the indicators/constants throughout.

packages/optimizely-sdk/tests/odpClient.spec.ts
mikechu-optimizely: Sure.  I can separate. 

I was attempting to reuse common fixtures and organize via describe() blocks.

packages/optimizely-sdk/lib/plugins/odp/odp_client.ts
```diff
@@ -81,7 +69,7 @@ export class OdpClient implements IOdpClient {
   public async querySegments(parameters: QuerySegmentsParameters): Promise<string | null> {
     if (!parameters?.apiEndpoint || !parameters?.apiKey) {
       this._logger.log(LogLevel.ERROR, 'No ApiHost or ApiKey set before querying segments');
-      return NULL_JSON_RESPONSE;
+      return null;
     }
```
jaeopt, line 74: We also filter out the request when "segmentsToCheck" is empty. The ODP server will reject it as an error.
SDK need to return empty segments array for them locally. I do not see that in "graphql_manager". Please add the filter.

jaeopt (changes requested):
All the changes look good!
A couple of small changes suggested.

packages/optimizely-sdk/lib/plugins/odp/odp_event.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export class OdpEvent {
+  private readonly _type: string;
```
zashraf1985, line 18: My preference would be to not use these leading `underscores` because it's already typescript and we are explicitly using `private` modifier. What do you think?

packages/optimizely-sdk/lib/plugins/odp/odp_request_parameters.ts
```diff
@@ -0,0 +1,38 @@
...
+  apiKey: string | undefined;
+  apiEndpoint: string | undefined;
+  httpVerb: string;
+}
+
+export abstract class OdpRequestParameters implements IOdpRequestParameters {
+  /**
+   * Optimizely Data Platform API key
+   */
+  public apiKey: string | undefined;
```
zashraf1985, line 27:
```suggestion
  public apiKey?: string;
```

packages/optimizely-sdk/lib/plugins/odp/odp_request_parameters.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  /**
+   * Fully-qualified URL to ODP events endpoint
+   */
+  public apiEndpoint: string | undefined;
```
zashraf1985, line 32:
```suggestion
  public apiEndpoint?: string;
```

zashraf1985 (changes requested):
Looks great overall! 

Super NIT: There are some places where `{type} | undefined` is used, i believe those can be changed to `{type}?`.

packages/optimizely-sdk/lib/plugins/odp/odp_client.ts
mikechu-optimizely: @jaeopt Is this an ok solution to empty `segmentsToCheck`?

```javascript
if (parameters?.segmentsToCheck?.length === 0) {
  this._logger.log(LogLevel.WARNING, 'No segments to query');
  return null;
}
```

jaeopt: @mikechu-optimizely  let's return an empty array (valid) as I see null as a fetch error.

packages/optimizely-sdk/lib/plugins/odp/odp_client.ts
```diff
@@ -84,11 +87,49 @@ export class OdpClient implements IOdpClient {
...
+  /**
+   * Handler for sending ODP events
+   * @param parameters
+   * @returns
+   * 1. null, When there was a non-recoverable error and no retry is needed.
+   * 2. 0 If an unexpected error occurred and retrying can be useful.
+   * 3. HTTPStatus code if httpclient was able to make the request and was able to receive response.
+   *    It is recommended to retry if status code was 5xx.
+   */
+  public async sendEvents(parameters: SendEventsParameters): Promise<number | null> {
```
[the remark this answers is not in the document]
mikechu-optimizely: That's a cool trick. 

Later I have `parameters.toJson();` which I'm not certain how I'd handle if deref-ing in the signature.

I'm committing a possible solution.  Tell me what you think.

packages/optimizely-sdk/lib/plugins/odp/odp_client.ts
jaeopt: @mikechu-optimizely Also remove log since this a common case when no segments are used for any audiences.

packages/optimizely-sdk/lib/utils/http_request_handler/browser_request_handler.ts
```diff
@@ -18,8 +18,6 @@ import { AbortableRequest, Headers, RequestHandler, Response } from './http';
 import { REQUEST_TIMEOUT_MS } from './config';
 import { LogHandler, LogLevel } from '../../modules/logging';
 
-const READY_STATE_DONE = 4;
```
mikechu-optimizely, line 21: @zashraf1985 Are you ok if we remove this in exchange for `XMLHttpRequest.DONE`?

zashraf1985: I think that will actually be better.

zashraf1985 (approved):
LGTM!

mikechu-optimizely:
@jaeopt Shall I hold for your approval too?

jaeopt (changes requested):
All changes for RestApi look good. A couple of suggestions for GraphQL Api.

packages/optimizely-sdk/lib/plugins/odp/odp_client.ts
```diff
@@ -67,16 +67,27 @@ export class OdpClient implements IOdpClient {
...
+    const { apiEndpoint, apiKey, httpVerb, userKey, userValue, segmentsToCheck } = parameters;
+
+    if (segmentsToCheck?.length === 0) {
+      return '';
+    }
+
+    if (!apiEndpoint || !apiKey) {
       this._logger.log(LogLevel.ERROR, 'No ApiHost or ApiKey set before querying segments');
       return null;
     }
```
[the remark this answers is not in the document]
mikechu-optimizely: I've added the check at the graphql_manager level too.  If you're okay with the redundancy, I'd like the check to also be on odp_client.

I'm happy to remove it, though.

packages/optimizely-sdk/lib/plugins/odp/graphql_manager.ts
```diff
@@ -73,27 +73,34 @@ export class GraphqlManager implements IGraphQLManager {
...
+
     const segmentsResponse = await this._odpClient.querySegments(parameters);
     if (!segmentsResponse) {
       this._logger.log(LogLevel.ERROR, 'Audience segments fetch failed (network error)');
+      return null;
+    }
+
+    if (segmentsToCheck?.length === 0) {
       return EMPTY_SEGMENTS_COLLECTION;
     }
```
jaeopt, line 98: Can we move this check up to before "odpClient.querySegments()" call? We should not call API with empty segmentsToCheck.

packages/optimizely-sdk/lib/plugins/odp/odp_client.ts
```diff
@@ -83,6 +79,10 @@ export class OdpClient implements IOdpClient {
...
 
+    if (segmentsToCheck?.length === 0) {
+      return '';
+    }
+
```
jaeopt, line 85: Return '' string here will be translated into a JSON error in the upper level.
I think we better remove all these redundant check here and requires apiKey/apiEndpoints/segmentsToCheck to have not-emtpy values.

packages/optimizely-sdk/lib/plugins/odp/odp_client.ts
```diff
@@ -118,17 +118,17 @@ export class OdpClient implements IOdpClient {
...
+    if (!apiEndpoint || !apiKey) {
+      this._logger.log(LogLevel.ERROR, 'No ApiEndpoint or ApiKey set before attempting to send ODP events');
       return null;
     }
 
-    if (!apiEndpoint || !apiKey) {
-      this._logger.log(LogLevel.ERROR, 'No ApiEndpoint or ApiKey set before attempting to send ODP events');
+    if (events?.length === 0) {
       return null;
     }
```
jaeopt, line 128: For consistency with segments-manager, we can move this validation to the upper level (event-manager) too.

jaeopt (changes requested):
A couple of more suggestions about filtering invalid api params.

packages/optimizely-sdk/lib/plugins/odp/odp_client.ts
mikechu-optimizely: I'd like to offer one last option for centralizing validation at this below-event-manager level.  

If this does not meet your approval, I will remove all validation at this level and depend upon event manager to solely.

jaeopt (changes requested):
I see some complexity added with params pack/unpack. Also concerned about exceptions thrown while packaging. We can consider removing the pack/unpack. Let's chat.

mikechu-optimizely:
Abandoning this PR.

PR #789, opened 2022-09-30
feat: Added ODPEventManager implementation

mikechu-optimizely:
## Summary
Added `OdpEventManager` implementation.

## Test plan
- Unit tests added
- Existing tests continue to pass

## Jira
FSSDK-8443


mikechu-optimizely:
Looks like there are older unit tests (around tests/v1EventProcessor.react_native.spec.ts) that are now failing.  

Fortunately, this also matches what's going on on my local dev machine.   I'll try correcting these unit tests and create a predecessor PR that we can merge before this PR. 

I still think code reviews for Event Manager here can continue.

mikechu-optimizely:
This is ready for review @jaeopt & @zashraf1985 whenever you're available.  I'll appreciate your thorough review.

packages/optimizely-sdk/lib/plugins/odp/odp_config.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+import { ODP_CONFIG_STATE } from '../../utils/enums';
+
+export class OdpConfig {
+  /**
+   * Host of ODP audience segments API.
+   * @private
+   */
+  private _apiHost: string;
```
zashraf1985, line 24: NIT: Since its typescript and we already have access modifiers available, i believe it is safe to remove these leading underscores for private members. what do you think?

packages/optimizely-sdk/lib/plugins/odp/odp_config.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Indicates whether ODP is integrated for the project
+   * @private
+   */
+  private _odpServiceIntegrated = ODP_CONFIG_STATE.UNDETERMINED;
+
+  constructor(apiKey: string, apiHost: string, segmentsToCheck?: string[]) {
+    this._apiKey = apiKey;
+    this._apiHost = apiHost;
+    this._segmentsToCheck = segmentsToCheck ?? [] as string[];
```
zashraf1985, line 71: NIT: It might actually work without explicit casting too, not sure though.

packages/optimizely-sdk/lib/plugins/odp/odp_config.ts
```diff
@@ -0,0 +1,102 @@
...
+    this._odpServiceIntegrated = this._apiKey && this._apiHost ? ODP_CONFIG_STATE.INTEGRATED : ODP_CONFIG_STATE.UNDETERMINED;
```
zashraf1985, line 72: This is being used at multiple places. May be move it to a funciton?

packages/optimizely-sdk/lib/plugins/odp/odp_config.ts
```diff
@@ -0,0 +1,102 @@
...
+
+      return true;
+    }
+  }
+
+  /**
+   * Determines if ODP configuration has the minimum amount of information
+   */
+  public isReady(): boolean {
+    return this._apiKey !== null && this._apiKey !== '' && this._apiHost !== null && this._apiHost !== '';
```
zashraf1985, line 100:
```suggestion
    return !!this._apiKey && !!this._apiHost;
```

packages/optimizely-sdk/lib/plugins/odp/odp_event_dispatcher.ts
```diff
@@ -0,0 +1,161 @@
...
+  private odpConfig: OdpConfig;
+  private shouldStopAndDrain = false;
+
+  private readonly apiManager: RestApiManager;
+  private readonly logger: LogHandler;
+  private readonly queueSize: number;
+  private readonly batchSize: number;
+  private readonly flushInterval: number;
+
+  public constructor(odpConfig: OdpConfig,
```
zashraf1985, line 58: Since there are so many options, it might be helpful and more user friendly to use an inline de structured object instead of expecting users to pass arguments in the correct places. What do you think?

packages/optimizely-sdk/lib/plugins/odp/odp_event_dispatcher.ts
```diff
@@ -0,0 +1,161 @@
...
+                     queueSize?: number,
+                     batchSize?: number,
+                     flushInterval?: number) {
+    this.odpConfig = odpConfig;
+    this.apiManager = apiManager;
+    this.logger = logger;
+
+    this.queueSize = queueSize && queueSize > 0 ? queueSize : DEFAULT_QUEUE_SIZE;
+    this.batchSize = batchSize && batchSize > 0 ? batchSize : DEFAULT_BATCH_SIZE;
+    this.flushInterval = flushInterval && flushInterval > 0 ? flushInterval : DEFAULT_FLUSH_INTERVAL;
```
zashraf1985, line 70:
```suggestion
    this.queueSize = queueSize || DEFAULT_QUEUE_SIZE;
    this.batchSize = batchSize || DEFAULT_BATCH_SIZE;
    this.flushInterval = flushInterval || DEFAULT_FLUSH_INTERVAL;
```

packages/optimizely-sdk/lib/plugins/odp/odp_event_dispatcher.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    this.state = STATE.STOPPED;
+    // initialize this way due to different types based on execution context
+    this.intervalId = setInterval(() => {
```
zashraf1985, line 74: What is the need of empty interval callback. It will be helpful to add a detailed comment here explaining it.

packages/optimizely-sdk/lib/plugins/odp/odp_event_dispatcher.ts
```diff
@@ -0,0 +1,161 @@
...
+    });
+  }
+
+  public start(): void {
+    this.state = STATE.RUNNING;
+    (async () => await this.processQueue())();
```
zashraf1985, line 80: Since `start` does not have to wait for `processQueue`. The inline async and await do not seem to do much to me. Can simply calling `this.processQueue` serve the purpose? If there is a special reason to add async and away here, it will be helpful to have a detailed explanation in a comment.

packages/optimizely-sdk/lib/plugins/odp/odp_event_dispatcher.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    this.queue.push(event);
+  }
+
+  private async processQueue(): Promise<void> {
+    if (this.state !== STATE.RUNNING && !this.shouldStopAndDrain) {
+      return;
+    }
+
+    clearInterval(this.intervalId);
```
zashraf1985, line 111: Just curios what purpose is this serving?

packages/optimizely-sdk/lib/plugins/odp/odp_event_dispatcher.ts
```diff
@@ -0,0 +1,161 @@
...
+      if (this.shouldStopAndDrain && this.queue.length > 0) {
+        this.logger.log(LogLevel.DEBUG, 'EventDispatcher draining queue without flush interval.');
+        await this.processQueue();
+      }
+    } else {
+      this.logger.log(LogLevel.DEBUG, 'ODPConfig not ready, discarding event batch.');
+    }
+
+    if (!this.shouldStopAndDrain) {
+      this.intervalId = setInterval(() => this.processQueue(), this.flushInterval);
```
zashraf1985, line 148: setInterval if for the case where you need a recurring callback at regular intervals till you clear it. Here since you are setting it every time and clearing it after one call, its more suitable to use `setTimeout`.

packages/optimizely-sdk/lib/plugins/odp/odp_event_dispatcher.ts
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    this.state = STATE.RUNNING;
```
zashraf1985, line 151: I believe this is redundant. what do you think?

packages/optimizely-sdk/lib/plugins/odp/odp_event_dispatcher.ts
zashraf1985, line 148: Setting the next tick to flushInterval is risky because then you will miss all the events in between. For example the batch size was 10 and flush interval was 3 seconds. You initially got 5 events and flush interval passed and you flushed those and then you set interval for next 3 seconds. but in the next 2nd, 40 more events enqueued, those events will need to wait for the flush interval to pass because there is nobody looking for those events in between. this is why we needed an implementation where it either waits on `flush interval` or next event to be available whatever happens first. This is why we used blocking queue in java sdk.

My understanding can be totally wrong though.. what do you think?

packages/optimizely-sdk/lib/plugins/odp/odp_event_dispatcher.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  public async stop(): Promise<void> {
+    this.logger.log(LogLevel.DEBUG, 'EventDispatcher stop requested.');
+    this.shouldStopAndDrain = true;
+    await this.processQueue();
```
zashraf1985, line 157: What is the need to call this one more time here?

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
```diff
@@ -0,0 +1,109 @@
...
+    this.eventDispatcher.updateSettings(odpConfig);
+  }
+
+  public sendEvents(events: OdpEvent[]): void {
+    events.forEach(event => this.sendEvent(event));
+  }
+
+  public sendEvent(event: OdpEvent): void {
+    event.data = this.augmentCommonData(event.data);
+    (async () => await this.eventDispatcher.enqueue(event))();
```
zashraf1985, line 78: why is this inline async await needed? Can we simply call the method directly?

zashraf1985 (changes requested):

packages/optimizely-sdk/lib/plugins/odp/odp_config.ts
mikechu-optimizely: Based on our Teams conversation: when getter or setter is in use, we'll accept underscore-prefixed names for backing fields. 

([resource](https://www.typescriptlang.org/docs/handbook/2/classes.html#getters--setters))

packages/optimizely-sdk/tests/odpEventManager.spec.ts
```diff
@@ -0,0 +1,303 @@
...
+    })),
+  ),
+  new OdpEvent(
+    't2',
+    'a2',
+    new Map([['id-key-2', 'id-value-2']]),
+    new Map(Object.entries({
+      'key-2': 'value2',
+    })),
+  ),
```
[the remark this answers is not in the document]
mikechu-optimizely: More in the `data` like this
```js
new OdpEvent(
    't2',
    'a2',
    new Map([['id-key-2', 'id-value-2']]),
    new Map(Object.entries({
      'key-2': 'value2',
      'data_source': 'my-source',
    })),
  ),
```
or another `OdpEvent` like this
```js
new OdpEvent(
    't2',
    'a2',
    new Map([['id-key-2', 'id-value-2']]),
    new Map(Object.entries({
      'key-2': 'value2',
    })),
  ),
  new OdpEvent(
    't3',
    'a3',
    new Map([['id-key-3', 'id-value-3']]),
    new Map(Object.entries({
      'data_source': 'my-source',
    })),
  ),
```

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    const event = new OdpEvent('fullstack', 'client_initialized', identifiers);
+    this.sendEvent(event);
+  }
+
+  public updateSettings(odpConfig: OdpConfig): void {
+    this.eventDispatcher.updateSettings(odpConfig);
+  }
+
+  public sendEvents(events: OdpEvent[]): void {
```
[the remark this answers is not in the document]
mikechu-optimizely: Oops. I took this from Java.  Sorry.

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
```diff
@@ -0,0 +1,109 @@
...
+    events.forEach(event => this.sendEvent(event));
+  }
+
+  public sendEvent(event: OdpEvent): void {
+    event.data = this.augmentCommonData(event.data);
```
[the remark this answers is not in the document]
mikechu-optimizely: @jaeopt , Can you direct me to some of the checks to perform in the Swift implementation or TDD so I can understand the types of validation? 

eg 
- only primatives
- must be non-null values
- no duplicates
- value length limitations
- total number of items in the collection, etc.

Also, what do we do with data items in the collection that is considered invalid?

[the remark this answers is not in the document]
mikechu-optimizely: Oh I found it in the TDD at the end of section 3.4.1 (items 1-4 specifically 4a-c). Apologies.

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
```diff
@@ -0,0 +1,109 @@
...
+      }
+    }
+
+    const data = new Map<string, unknown>();
+    data.set('idempotence_id', uuid());
+    data.set('data_source_type', 'sdk');
+    data.set('data_source', 'javascript-sdk');
+    if (sourceVersion) {
+      data.set('data_source_version', sourceVersion);
+    }
```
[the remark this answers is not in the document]
mikechu-optimizely: This research and implementation of handling user agent data from the browser is proving to be quite complex.  

I'd like to move this to another Jira issue after talking it through with Zeeshan.   

This will allow us to move this PR forward and also help me architect this a true plugin so we don't increase the default bundle sizes.

mikechu-optimizely:
@zashraf1985, Thanks again for pair programming the round of reviews. I'm ready for another look by you and @jaeopt when you have a moment.

jaeopt (changes requested):
Changes looks good. I see a possible failure in batching and a couple of suggestions to consider.

packages/optimizely-sdk/lib/plugins/odp/odp_event_dispatcher.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    // Flush interval occurred & queue has items
+    if (shouldFlush && this.queueContainsItems()) {
+      // clear the queue completely
+      this.clearCurrentTimeout();
+
+      this.state = STATE.PROCESSING;
+
+      while (this.queueContainsItems()) {
+        await this.makeAndSendBatch();
```
zashraf1985, line 182: We should not probably wait for this here and just fire and forget the actual request. This await will cause a weird synchronization problem. I can discuss this and explain it on the call better.

packages/optimizely-sdk/lib/plugins/odp/odp_config.ts
```diff
@@ -97,6 +107,6 @@ export class OdpConfig {
    * Determines if ODP configuration has the minimum amount of information
    */
   public isReady(): boolean {
-    return this._apiKey !== null && this._apiKey !== '' && this._apiHost !== null && this._apiHost !== '';
+    return !!this._apiKey && !!this._apiHost;
```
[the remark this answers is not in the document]
mikechu-optimizely: Good eye.  I think I was copying Python too closely without full understanding of how it was used.  I think I'll remove `_odpServiceIntegrated`

packages/optimizely-sdk/lib/plugins/odp/odp_event_dispatcher.ts
```diff
@@ -101,60 +153,146 @@ export class OdpEventDispatcher implements IOdpEventDispatcher {
     }
 
     this.queue.push(event);
+
+    this.processQueue();
```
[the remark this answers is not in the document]
mikechu-optimizely: `start()` handles kicking off the timer to flush the queue no matter event count.

Here, I want to see if by adding a new event we have a batch to process.

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
```diff
@@ -0,0 +1,160 @@
...
+ * Concrete implementation of a manager for persisting events to the Optimizely Data Platform
+ */
+export class OdpEventManager implements IOdpEventManager {
+  private readonly eventDispatcher: OdpEventDispatcher;
+  private readonly logger: LogHandler;
+  private readonly clientEngine: string;
+  private readonly clientVersion: string;
+
+  public constructor({ eventDispatcher, logger, clientEngine, clientVersion }: {
+    eventDispatcher: OdpEventDispatcher,
```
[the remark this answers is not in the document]
mikechu-optimizely: Hmmm 🤔   Java had it as a subclass.  I tried replicating via a nested class in Typescript but it was really ugly and unclear.  

If you're okay with the additional time use, I can try not subclassing and maybe run it all under one class. I'll give it a try.

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
```diff
@@ -14,96 +14,152 @@
...
-      if (process) {
-        sourceVersion = process.version;
+  /**
+   * Validate event data value types
+   * @param data Event data to be validated
+   * @returns Array of event data keys that were found to be invalid
+   * @private
+   */
+  private findKeysWithInvalidData(data: Map<string, unknown>): string[] {
+    const validTypes: string[] = ['string', 'number', 'boolean', 'bigint'];
```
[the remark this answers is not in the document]
mikechu-optimizely: I was going through the Typescript primatives and just listing the ones that sort of made sense.  I'll remove `bigint`. Easy.

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * Send an event to ODP via dispatch queue
+   * @param event ODP Event to forward
+   */
+  public sendEvent(event: OdpEvent): void {
+    const foundInvalidDataInKeys = this.findKeysWithInvalidData(event.data);
+    if (foundInvalidDataInKeys.length > 0) {
+      this.logger.log(LogLevel.ERROR, `Event data found to be invalid.`);
+      this.logger.log(LogLevel.DEBUG, `Invalid event data keys (${foundInvalidDataInKeys.join(', ')})`);
```
zashraf1985, line 113: Is there a way to pass this dynamic string as parameters to log function to be concatenated later only if log level applies?

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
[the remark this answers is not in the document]
zashraf1985: I agree too. A simple merge should work. Just copy all the functions from the dispatcher to here. what do you think?

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
```diff
@@ -0,0 +1,160 @@
...
+   * @param sourceData Existing event data to augment
+   * @returns Augmented event data
+   * @private
+   */
+  private augmentCommonData(sourceData: Map<string, unknown>): Map<string, unknown> {
+    const data = new Map<string, unknown>();
+    data.set('idempotence_id', uuid());
+    data.set('data_source_type', 'sdk');
+    data.set('data_source', this.clientEngine);
+    data.set('data_source_version', this.clientVersion);
```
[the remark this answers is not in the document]
mikechu-optimizely: Finding a good architectural solution to a complex JS Browser problem was taking longer than I thought needed.  

I'd like to move this to Jira FSSDK-8651 to not delay this merge.

zashraf1985 (changes requested):
The overall logic of event queue looks very good. I just have a concern in one asynchronous scenario that we can discuss on call. Also suggested a few minor changes here and there.

packages/optimizely-sdk/tests/odpEventManager.spec.ts
```diff
@@ -56,8 +59,8 @@ const PROCESSED_EVENTS: OdpEvent[] = [
     new Map(Object.entries({
       'idempotence_id': MOCK_IDEMPOTENCE_ID,
       'data_source_type': 'sdk',
-      'data_source': 'javascript-sdk',
-      'data_source_version': MOCK_PROCESS_VERSION,
+      'data_source': clientEngine,
+      'data_source_version': clientVersion,
       'key-1': 'value1',
       'key-2': null,
       'key-3': 3.3,
```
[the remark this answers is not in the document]
mikechu-optimizely: I'll remove `bigint`.  It should be a rare case that we need to support numbers past [MAX_VALUE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_VALUE).

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
```diff
@@ -14,96 +14,152 @@
...
     data.set('data_source_type', 'sdk');
-    data.set('data_source', 'javascript-sdk');
-    if (sourceVersion) {
-      data.set('data_source_version', sourceVersion);
+
+    if (this.clientEngine) {
+      data.set('data_source', this.clientEngine);
+    }
+    if (this.clientVersion) {
+      data.set('data_source_version', this.clientVersion);
```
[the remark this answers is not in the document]
mikechu-optimizely: Sure I'll take a look.  That would save us a Jira ticket 😄

packages/optimizely-sdk/lib/plugins/odp/odp_event_dispatcher.ts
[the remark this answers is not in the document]
jaeopt: It looks like the timer set by start() will not be rescheduled after 1st interval if we do not get any event.
If the first event arrives at 11sec, then it won't be flushed until we have 10 events (otherwise never fired). Can you double check if my understanding is correct?

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
jaeopt: Or consider to remove this too-kind debug message :) They will figure out which one is invalid if we show error.

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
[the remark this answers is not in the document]
mikechu-optimizely: odp_event_dispatcher.ts deleted

packages/optimizely-sdk/lib/plugins/odp/odp_event_dispatcher.ts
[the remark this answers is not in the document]
mikechu-optimizely: Hmm maybe we should pair up on a call, so I understand your view.

In `setNewTimeout()` I'm going to `this.processQueue(**true**)`.  (vs the `this.processQueue()`)  I sort of wanted to have these as two methods processQueue and flushQueue so it was more clear

cc @zashraf1985

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
mikechu-optimizely: Removed.

packages/optimizely-sdk/lib/plugins/odp/odp_event_manager.ts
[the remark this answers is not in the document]
mikechu-optimizely: I searched the code and couldn't find where we're gathering an OS, OS version, device type.  

@zashraf1985 , Do you know if we have this already or should we move forward with FSSDK-8651?

mikechu-optimizely:
Good morning @jaeopt & @zashraf1985.  You'll never guess, but I'm ready for another code review.  👍 😄

packages/optimizely-sdk/tests/odpEventManager.spec.ts
```diff
@@ -0,0 +1,379 @@
...
+      flushInterval: 100,
+    });
+    const spiedEventManager = spy(eventManager);
+
+    eventManager.start();
+    // do not add events to the queue, but allow for...
+    await pause(400); // at least 3 flush intervals executions (giving a little longer)
+
+    verify(spiedEventManager['processQueue'](anything())).atLeast(3);
+  });
```
mikechu-optimizely, line 219: @jaeopt  Is this a good test for ensuring flush intervals?

zashraf1985 (approved):
Looks Good! Great Work!

jaeopt (approved):
LGTM

mikechu-optimizely:
Thanks @jaeopt and @zashraf1985

PR #794, opened 2023-01-17
Run project config parse inside web worker

PepijnSenders:
## Summary

The `createProjectConfig()` function is quite slow and runs sync with the application if the datafile is provided as an `object`.

This function could easily be run inside a web worker, freeing up the main thread, and we could use the same promise to track when it's done as we use for when we fetch the datafile.

For this PR to work, we'd still need:

- A way to publish a worker with this project, we could parse the worker code as a string, for that we should isolate the functions it uses a bit better to reduce the load of this string. We could also have the worker as its own file in the repo, and the consumer could pass it as a parameter to the `createInstance` function.

### DIFF
--- a/packages/optimizely-sdk/lib/core/project_config/project_config_manager.ts
+++ b/packages/optimizely-sdk/lib/core/project_config/project_config_manager.ts
@@ -18,11 +18,7 @@ import { sprintf } from '../../utils/fns';
 
 import { ERROR_MESSAGES } from '../../utils/enums';
 import { createOptimizelyConfig } from '../optimizely_config';
-import {
-  OnReadyResult,
-  OptimizelyConfig,
-  DatafileManager,
-} from '../../shared_types';
+import { OnReadyResult, OptimizelyConfig, DatafileManager } from '../../shared_types';
 import { ProjectConfig, toDatafile, tryCreatingProjectConfig } from '../project_config';
 
 const logger = getLogger();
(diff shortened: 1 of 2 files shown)


packages/optimizely-sdk/lib/core/project_config/project_config_manager.ts
```diff
@@ -197,6 +200,41 @@ export class ProjectConfigManager {
     return error;
   }
 
+  private async handleDatafileInWebWorker(newDatafile: string | object): Promise<OnReadyResult> {
+    const workerCode = '';
```
PepijnSenders, line 204: This needs to be handled by some loader inside rollup; like this one: https://www.npmjs.com/package/rollup-plugin-web-worker-loader

opti-jnguyen:
@PepijnSenders - thanks for submitting the PR! Looked through it and looks pretty slick.

As this is technically a feature request, we'll want to document it through our feature request pipeline as well.
Here's the link to the Optimizely Feedback page where you can submit a feature request in the form of feedback: https://feedback.optimizely.com/?category=7093871273740179703

Please submit a request there when you have the chance - thanks in advance!

PepijnSenders:
Hey @opti-jnguyen, I was already discussing this change with one of your Solution Architects, do you need both? Happy to do so!

russell-loube-optimizely:
Hi @PepijnSenders , I apologize for the delayed reply here. It's my understanding that you're also speaking with one of our Solution Architects. If he has not mentioned this already, we would greatly appreciate if you could please submit this through the feature request pipeline as well.

Rest assured we are tracking the issue within our Development team. If it's not too much trouble to also submit this via the feature request pipeline, this will allow for more visibility across our Product and Engineering org.

Lastly, I believe Joshua has spoken with you about next steps for evaluating this change in your environment. I cannot comment on when we can look at this in more detail past our initial review unfortunately due to prior planned work.

Thanks for your collaboration and if you have any questions please let us know.

mikechu-optimizely:
This was last mentioned in internal spike/research ticket FSSDK-8510.  I've requested a dev ticket be created.

mikechu-optimizely:
> This was last mentioned in internal spike/research ticket FSSDK-8510. I've requested a dev ticket be created.

Oops. The development ticket for this is FSSDK-9614.  We should be pointing it shortly.

jame-earnin:
Any update?

mikechu-optimizely:
> Any update?

I bumped the internal work ticket for comment.

raju-opti:
This has been open for quiet some time. Adding this support adds complexity to the maintenance of the SDK. Also, we support providing synchronous decision when a datafile is provided during construction. Considering all of these, we have decided not to add support for config parsing in web worker. Instead, we will look to improve the performance of the createProjectConfig function itself.

PR #795, opened 2023-01-27
Add Notification Registry for ODP Setting Updates

opti-jnguyen:
## Summary

(Copied from corresponding Java PR: https://github.com/optimizely/java-sdk/pull/501)

- NotificationRegistry is added to create default NotificationCenter for internal use for updating ODPConfig.
- This is to handle a corner case that if in any case user passes different `notificationCenter` to Optimizely and `ProjectConfigManager` then the Notification of `UpdateProjectConfig` will not trigger because its handler is getting added in the main `NotificationCenter` object.
- This will make sure `UpdateProjectConfig` will update `ODPConfig` every time irrespective of different NotificationCenter passed by user.

> Note: cachedConfig doesn't appear to be needed for JS implementation as getConfig() already returns a locally stored configObject: ProjectConfig in memory without the need to await for a return

## Test plan

- Added additional unit test to test NotificationRegistry
- All tests should pass.

## Issues
- [FSSDK-8779](https://jira.sso.episerver.net/browse/FSSDK-8779)

### DIFF
--- a/packages/optimizely-sdk/lib/core/notification_center/notification_registry.ts
+++ b/packages/optimizely-sdk/lib/core/notification_center/notification_registry.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { getLogger, LogHandler } from '../../modules/logging';
+import { getLogger, LogHandler, LogLevel } from '../../modules/logging';
 import { NotificationCenter, createNotificationCenter } from '../../core/notification_center';
 
 /**
@@ -25,15 +25,27 @@ export class NotificationRegistry {
 
   constructor() {}
 
-  public static getNotificationCenter(sdkKey?: string, logger?: LogHandler): NotificationCenter | null {
-    if (!sdkKey) return null;
+  /**
+   * Retrieves an SDK Key's corresponding notification center in the registry if it exists, otherwise it creates one
+   * @param sdkKey SDK Key to be used for the notification center tied to the ODP Manager
+   * @param logger Logger to be used for the corresponding notification center
+   * @returns {NotificationCenter | undefined} a notification center instance for ODP Manager if a valid SDK Key is provided, otherwise undefined
+   */
+  public static getNotificationCenter(
+    sdkKey?: string,
+    logger: LogHandler = getLogger()
+  ): NotificationCenter | undefined {
+    if (!sdkKey) {
+      logger.log(LogLevel.ERROR, 'No SDK key provided to getNotificationCenter.');
+      return undefined;
+    }
 
     let notificationCenter;
     if (this._notificationCenters.has(sdkKey)) {
-      notificationCenter = this._notificationCenters.get(sdkKey) || null;
+      notificationCenter = this._notificationCenters.get(sdkKey);
     } else {
       notificationCenter = createNotificationCenter({
-        logger: logger || getLogger(),
+        logger,
         errorHandler: { handleError: () => {} },
       });
       this._notificationCenters.set(sdkKey, notificationCenter);
@@ -43,7 +55,9 @@ export class NotificationRegistry {
   }
 
   public static removeNotificationCenter(sdkKey?: string): void {
-    if (!sdkKey) return;
+    if (!sdkKey) {
+      return;
+    }
 
     const notificationCenter = this._notificationCenters.get(sdkKey);
     if (notificationCenter) {
--- a/packages/optimizely-sdk/lib/core/notification_center/notification_registry.ts
+++ b/packages/optimizely-sdk/lib/core/notification_center/notification_registry.ts
@@ -1,0 +1,54 @@
+/**
+ * Copyright 2023, Optimizely
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { getLogger, LogHandler } from '../../modules/logging';
+import { NotificationCenter, createNotificationCenter } from '../../core/notification_center';
+
+/**
+ * Internal notification center registry for managing multiple notification centers.
+ */
+export class NotificationRegistry {
+  private static _notificationCenters = new Map<string, NotificationCenter>();
+
+  constructor() {}
+
+  public static getNotificationCenter(sdkKey?: string, logger?: LogHandler): NotificationCenter | null {
+    if (!sdkKey) return null;
+
+    let notificationCenter;
+    if (this._notificationCenters.has(sdkKey)) {
+      notificationCenter = this._notificationCenters.get(sdkKey) || null;
+    } else {
+      notificationCenter = createNotificationCenter({
+        logger: logger || getLogger(),
+        errorHandler: { handleError: () => {} },
+      });
+      this._notificationCenters.set(sdkKey, notificationCenter);
+    }
+
+    return notificationCenter;
+  }
+
+  public static removeNotificationCenter(sdkKey?: string): void {
+    if (!sdkKey) return;
+
+    const notificationCenter = this._notificationCenters.get(sdkKey);
+    if (notificationCenter) {
+      notificationCenter.clearAllNotificationListeners();
+      this._notificationCenters.delete(sdkKey);
+    }
+  }
+}
--- a/packages/optimizely-sdk/lib/core/odp/odp_config.ts
+++ b/packages/optimizely-sdk/lib/core/odp/odp_config.ts
@@ -47,20 +47,20 @@ export class OdpConfig {
    * All ODP segments used in the current datafile (associated with apiHost/apiKey).
    * @private
    */
-  private _segmentsToCheck: Set<string>;
+  private _segmentsToCheck: string[];
 
   /**
    * Getter for ODP segments to check
    * @public
    */
-  public get segmentsToCheck(): Set<string> {
+  public get segmentsToCheck(): string[] {
     return this._segmentsToCheck;
   }
 
-  constructor(apiKey?: string, apiHost?: string, segmentsToCheck?: Set<string>) {
+  constructor(apiKey?: string, apiHost?: string, segmentsToCheck?: string[]) {
     if (apiKey) this._apiKey = apiKey;
     if (apiHost) this._apiHost = apiHost;
-    this._segmentsToCheck = segmentsToCheck ?? new Set<string>('');
+    this._segmentsToCheck = segmentsToCheck ?? [];
   }
 
   /**
@@ -70,7 +70,7 @@ export class OdpConfig {
    * @param segmentsToCheck Audience segments
    * @returns true if configuration was updated successfully
    */
-  public update(apiKey?: string, apiHost?: string, segmentsToCheck?: Set<string>): boolean {
+  public update(apiKey?: string, apiHost?: string, segmentsToCheck?: string[]): boolean {
     if (this._apiKey === apiKey && this._apiHost === apiHost && this._segmentsToCheck === segmentsToCheck) {
       return false;
     } else {
@@ -98,16 +98,18 @@ export class OdpConfig {
     return (
       this._apiHost == config._apiHost &&
       this._apiKey == config._apiKey &&
-      this.segmentsToCheck.size == config._segmentsToCheck.size &&
-      this.checkSetEquality(this.segmentsToCheck, config._segmentsToCheck)
+      this.segmentsToCheck.length == config._segmentsToCheck.length &&
+      this.checkArrayEquality(this.segmentsToCheck, config._segmentsToCheck)
     );
   }
 
-  private checkSetEquality(setA: Set<string>, setB: Set<string>) {
-    let isEqual = true;
-    setA.forEach(item => {
-      if (!setB.has(item)) isEqual = false;
-    });
-    return isEqual;
+  /**
+   * Checks two string arrays for equality.
+   * @param arrayA First Array to be compared against.
+   * @param arrayB Second Array to be compared against.
+   * @returns {boolean} True if both arrays are equal, otherwise returns false.
+   */
+  private checkArrayEquality(arrayA: string[], arrayB: string[]): boolean {
+    return arrayA.length === arrayB.length && arrayA.every((item, index) => item === arrayB[index]);
   }
 }
--- a/packages/optimizely-sdk/lib/core/odp/odp_config.ts
+++ b/packages/optimizely-sdk/lib/core/odp/odp_config.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { checkArrayEquality } from '../../../lib/utils/fns';
+
 export class OdpConfig {
   /**
    * Host of ODP audience segments API.
@@ -98,18 +100,7 @@ export class OdpConfig {
     return (
       this._apiHost == config._apiHost &&
       this._apiKey == config._apiKey &&
-      this.segmentsToCheck.length == config._segmentsToCheck.length &&
-      this.checkArrayEquality(this.segmentsToCheck, config._segmentsToCheck)
+      checkArrayEquality(this.segmentsToCheck, config._segmentsToCheck)
     );
-  }
-
-  /**
-   * Checks two string arrays for equality.
-   * @param arrayA First Array to be compared against.
-   * @param arrayB Second Array to be compared against.
-   * @returns {boolean} True if both arrays are equal, otherwise returns false.
-   */
-  private checkArrayEquality(arrayA: string[], arrayB: string[]): boolean {
-    return arrayA.length === arrayB.length && arrayA.every((item, index) => item === arrayB[index]);
   }
 }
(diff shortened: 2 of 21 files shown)


packages/optimizely-sdk/lib/core/notification_center/notification_registry.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Internal notification center registry for managing multiple notification centers.
+ */
+export class NotificationRegistry {
+  private static _notificationCenters = new Map<string, NotificationCenter>();
+
+  constructor() {}
+
+  public static getNotificationCenter(sdkKey?: string, logger?: LogHandler): NotificationCenter | null {
```
mikechu-optimizely, line 28: I think `string` can be null, but could we update the signature, removing the `?` to discourage passing a null value?  

What's your opinion on `logger: LogHandler = null` too?

packages/optimizely-sdk/lib/core/odp/odp_config.ts
```diff
@@ -47,20 +47,20 @@ export class OdpConfig {
    * All ODP segments used in the current datafile (associated with apiHost/apiKey).
    * @private
    */
-  private _segmentsToCheck: string[];
+  private _segmentsToCheck: Set<string>;
```
mikechu-optimizely, line 50: 👍

packages/optimizely-sdk/lib/core/odp/odp_config.ts
```diff
@@ -98,7 +98,16 @@ export class OdpConfig {
...
       this._apiKey == config._apiKey &&
-      JSON.stringify(this.segmentsToCheck) == JSON.stringify(config._segmentsToCheck)
+      this.segmentsToCheck.size == config._segmentsToCheck.size &&
+      this.checkSetEquality(this.segmentsToCheck, config._segmentsToCheck)
     );
   }
+
+  private checkSetEquality(setA: Set<string>, setB: Set<string>) {
+    let isEqual = true;
+    setA.forEach(item => {
```
mikechu-optimizely, line 108: Above L108, would a quick check for Set length equality help speed up a decision?

packages/optimizely-sdk/lib/core/odp/odp_config.ts
```diff
@@ -98,7 +98,16 @@ export class OdpConfig {
...
       this._apiHost == config._apiHost &&
       this._apiKey == config._apiKey &&
-      JSON.stringify(this.segmentsToCheck) == JSON.stringify(config._segmentsToCheck)
+      this.segmentsToCheck.size == config._segmentsToCheck.size &&
+      this.checkSetEquality(this.segmentsToCheck, config._segmentsToCheck)
     );
   }
+
+  private checkSetEquality(setA: Set<string>, setB: Set<string>) {
+    let isEqual = true;
```
mikechu-optimizely, line 107: I like to go with a pessimistic/`false` initial value, imho.  Just a thought.

packages/optimizely-sdk/lib/core/odp/odp_config.ts
```diff
@@ -98,7 +98,16 @@ export class OdpConfig {
...
+    setA.forEach(item => {
+      if (!setB.has(item)) isEqual = false;
```
mikechu-optimizely, line 109: Do we want to `break` as soon as possible?

packages/optimizely-sdk/lib/shared_types.ts
```diff
@@ -389,6 +391,8 @@ export interface Config extends ConfigLite {
   eventMaxQueueSize?: number;
   // sdk key
   sdkKey?: string;
+  // odp manager
```
mikechu-optimizely, line 394: NIT: Do we want to shorten and remove comments?

packages/optimizely-sdk/tests/odpManager.spec.ts
```diff
@@ -94,7 +94,7 @@ describe('OdpManager', () => {
 
   it('should start ODP Event manager when ODP Manager is initialized', async () => {
     const manager = new OdpManager(false, requestHandler, logger, undefined, mockSegmentManager, mockEventManager);
-    expect(manager.eventManager?.state).to.equal(STATE.RUNNING);
+    // expect(manager.eventManager?.state).equal(STATE.RUNNING);
```
mikechu-optimizely, line 97: Is this assertion no longer valid/valuable?

packages/optimizely-sdk/tests/odpSegmentApiManager.ts
```diff
@@ -150,7 +150,7 @@ describe('OdpSegmentApiManager', () => {
 
     const segments = await manager.fetchSegments(API_key, GRAPHQL_ENDPOINT, USER_KEY, USER_VALUE, SEGMENTS_TO_CHECK);
 
-    expect(segments).toHaveLength(2);
+    if (segments) expect(segments.size).toEqual(2);
```
mikechu-optimizely, line 153: Is this equivalent `expect(segments?.size).toEqual(2);`

mikechu-optimizely (commented):
Just some thoughts.  I'm not sure I'm the official code reviewer so....

packages/optimizely-sdk/lib/core/notification_center/notification_registry.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+import { getLogger, LogHandler } from '../../modules/logging';
+import { NotificationCenter, createNotificationCenter } from '../../core/notification_center';
+
+/**
+ * Internal notification center registry for managing multiple notification centers.
+ */
+export class NotificationRegistry {
+  private static _notificationCenters = new Map<string, NotificationCenter>();
```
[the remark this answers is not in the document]
opti-jnguyen: Chatted w/Zeeshan about that last week and we came to the conclusion that we didn't due to the single-threaded synchronous nature of JS. Any thoughts on corner cases we may have missed?

packages/optimizely-sdk/lib/core/notification_center/notification_registry.ts
opti-jnguyen: Right - I initially had considered making both non-nullable values but ended up following in the footsteps of the corresponding Java PR.

Either approach has its pros/cons. I'll make some changes based on the following thoughts:

- [ ] If `sdkKey` is not defined, we should also log an error indicating it's required.
- [ ] Since we use the default logger in case a logger isn't manually passed in, perhaps `logger: LogHandler = getLogger()` might work better? We'd also avoid having a union type of `LogHandler | null` as well.
- [ ] Also forgot to add a documentation comment on this function for that juicy tooltip - will have queue that up too.

packages/optimizely-sdk/lib/core/odp/odp_config.ts
opti-jnguyen: I usually like to do that too, will reverse the conditional logic for this one.

packages/optimizely-sdk/lib/shared_types.ts
opti-jnguyen: True, for SDK Key and ODP Manager, for example, it's particularly redundant. Will remove those comments.

packages/optimizely-sdk/tests/odpManager.spec.ts
opti-jnguyen: Thanks for the catch, was playing around with ODPManager-specific tests and this snuck into the PR by accident.

packages/optimizely-sdk/tests/odpSegmentApiManager.ts
opti-jnguyen: Initially tried that and it complained - tried it again and now seems to work just fine. Switching it to `segments?.size` while it's not complaining anymore.

packages/optimizely-sdk/lib/core/odp/odp_config.ts
opti-jnguyen: So because we're limited to ES5 (AFAIK) and can't use for..of on Sets, I went with this approach to sidestep some other approaches.

I'll update this with an additional conditional for Set size check and also show what it looks like with a clean-ish short-circuit using Array.from to transform the sets first. (Unfortunately short-circuit doesn't work in .every or .forEach :c)

packages/optimizely-sdk/lib/core/odp/odp_config.ts
opti-jnguyen: Will remove this variable altogether for the short-circuit approach.

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
```diff
@@ -73,7 +74,7 @@ export class OdpManager {
     disable: boolean,
     requestHandler: RequestHandler,
     logger?: LogHandler,
-    segmentsCache?: LRUCache<string, string[]>,
+    segmentsCache?: LRUCache<string, Set<string>>,
```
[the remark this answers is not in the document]
jaeopt: Also if more convenient, we can keep the <array> definition everywhere instead of <set> after we build non-redundant array in projectConfig. After that in sdk and odp server will use good array segments (non-redundant).

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -175,6 +179,26 @@ export default class Optimizely {
...
     this.readyTimeouts = {};
     this.nextReadyTimeoutId = 0;
+
+    if (config.odpManager != null) {
+      this.odpManager = config.odpManager;
+      this.odpManager.eventManager?.start();
+      if (this.projectConfigManager.getConfig() != null) {
+        this.updateODPSettings();
+      }
+      const sdkKey = this.projectConfigManager.getConfig()?.sdkKey;
```
jaeopt, line 189: We have sdkKey here (not indirectly from projectConfigManager) in line 132.
We can consider remove "getSdkKey()" new method from projectConfigManager and use a local value.
Also, we have to change "sdkKey" is mandatory (not null), which may be a breaking change in the release.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -175,6 +179,26 @@ export default class Optimizely {
...
+      if (sdkKey != null) {
```
jaeopt, line 190: if sdkKey is null, then it should log as error.

jaeopt (changes requested):
A suggestion to clean up redundant sdkKey in 2 places.

packages/optimizely-sdk/lib/optimizely/index.ts
opti-jnguyen: Thanks for the catch - swapped out .getConfig()?sdkKey references with local config.sdkKey usage.

Regarding changing `sdkKey` to be mandatory, this will definitely be a breaking change - especially since there's some nuances (for example, the Lite bundle does not expect `sdkKey` to begin with). Brainstorming on this one.

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
```diff
@@ -154,15 +155,16 @@ export class OdpManager {
...
+   * @param {string}                          userId  - Unique identifier of a target user.
+   * @param {Set<OptimizelySegmentOption}     options - A set of OptimizelySegmentOption used to ignore and/or reset the cache.
+   * @returns {Promise<string[] | null>}      A promise holding either a list of qualified segments or null.
    */
   public async fetchQualifiedSegments(
     userKey: ODP_USER_KEY,
     userId: string,
-    options: Array<OptimizelySegmentOption>
-  ): Promise<string[] | null> {
+    options: Set<OptimizelySegmentOption>
```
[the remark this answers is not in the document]
opti-jnguyen: Swapped everything back over to Arrays and refactored previously approved "Set" related code to arrays as well for a handful of variables related to segments such as allSegments, segmentsCache, and options.

packages/optimizely-sdk/lib/core/notification_center/notification_registry.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  constructor() {}
+
+  public static getNotificationCenter(sdkKey?: string, logger?: LogHandler): NotificationCenter | null {
+    if (!sdkKey) return null;
+
+    let notificationCenter;
+    if (this._notificationCenters.has(sdkKey)) {
+      notificationCenter = this._notificationCenters.get(sdkKey) || null;
```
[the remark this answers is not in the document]
opti-jnguyen: The .get() here potentially returns "undefined" despite the .has() check above. Changed function return from `| null` to `| undefined` as a workaround in order to incorporate this suggestion.

packages/optimizely-sdk/lib/core/odp/odp_config.ts
opti-jnguyen: Refactored out Sets for Arrays as per Jae's comments below.

packages/optimizely-sdk/lib/core/odp/odp_config.ts
```diff
@@ -98,7 +98,18 @@ export class OdpConfig {
     return (
       this._apiHost == config._apiHost &&
       this._apiKey == config._apiKey &&
-      JSON.stringify(this.segmentsToCheck) == JSON.stringify(config._segmentsToCheck)
+      this.segmentsToCheck.length == config._segmentsToCheck.length &&
```
msohailhussain, line 101: we are checking this condition in `checkArrayEquality`

packages/optimizely-sdk/lib/core/odp/odp_config.ts
```diff
@@ -98,7 +98,18 @@ export class OdpConfig {
...
     );
   }
+
+  /**
+   * Checks two string arrays for equality.
+   * @param arrayA First Array to be compared against.
+   * @param arrayB Second Array to be compared against.
+   * @returns {boolean} True if both arrays are equal, otherwise returns false.
+   */
+  private checkArrayEquality(arrayA: string[], arrayB: string[]): boolean {
```
msohailhussain, line 112: what about to add in utils?

packages/optimizely-sdk/lib/modules/datafile-manager/httpPollingDatafileManager.ts
```diff
@@ -233,6 +239,7 @@ export default abstract class HttpPollingDatafileManager implements DatafileMana
           datafile,
         };
         this.emitter.emit(UPDATE_EVT, datafileUpdate);
+        NotificationRegistry.getNotificationCenter(this.sdkKey, logger)?.sendNotifications(NOTIFICATION_TYPES.OPTIMIZELY_CONFIG_UPDATE)
```
msohailhussain, line 242: Can you please bring this first and before emitting.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -175,6 +179,27 @@ export default class Optimizely {
 
     this.readyTimeouts = {};
     this.nextReadyTimeoutId = 0;
+
+    if (config.odpManager != null) {
```
msohailhussain, line 183: All this logic should come-up after the readyPromise? How config.sdkKey will be `non-null` if the datafile is still being fetched.

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -175,6 +179,27 @@ export default class Optimizely {
...
+      this.odpManager = config.odpManager;
+      this.odpManager.eventManager?.start();
+      if (this.projectConfigManager.getConfig() != null) {
```
msohailhussain, line 186: is this the blocking call ?

msohailhussain (commented):
have some questions, we may discuss offline.

packages/optimizely-sdk/lib/optimizely/index.ts
opti-jnguyen: Moving logic into Promise.all.

Also, as discussed, this should catch for both cases where a new Optimizely instance is provided either an SDK Key  or manually includes a datafile.

packages/optimizely-sdk/lib/optimizely/index.ts
opti-jnguyen: This will be resolved by the patch mentioned above.

jaeopt (changes requested):
A few changes suggested.

packages/optimizely-sdk/lib/core/odp/odp_config.ts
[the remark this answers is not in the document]
opti-jnguyen: Thanks, forgot to push this change - should be resolved now.

packages/optimizely-sdk/lib/core/project_config/index.ts
```diff
@@ -167,13 +167,13 @@ export const createProjectConfig = function (
...
+  projectConfig.allSegments = []
 
   Object.keys(projectConfig.audiencesById)
     .map((audience) => getAudienceSegments(projectConfig.audiencesById[audience]))
     .forEach(audienceSegments => {
       audienceSegments.forEach(segment => {
-        projectConfig.allSegments.add(segment)
+        projectConfig.allSegments.push(segment)
       })
     })
```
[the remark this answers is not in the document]
opti-jnguyen: Same comment as above

packages/optimizely-sdk/lib/core/project_config/index.tests.js
```diff
@@ -836,8 +836,8 @@ describe('lib/core/project_config', function () {
       })
 
       it('should contain all expected unique odp segments in allSegments', () => {
-        assert.equal(config.allSegments.size, 3)
-        assert.deepEqual(config.allSegments, new Set(['odp-segment-1', 'odp-segment-2', 'odp-segment-3']))
+        assert.equal(config.allSegments.length, 3)
+        assert.deepEqual(config.allSegments, ['odp-segment-1', 'odp-segment-2', 'odp-segment-3'])
```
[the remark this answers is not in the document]
opti-jnguyen: Same comment as above

jaeopt (approved):
LGTM

PR #797, opened 2023-02-11
[FSSDK-8452] feat: Added ODP Manager Implementation

opti-jnguyen:
## Summary
Added ODPManager Implementation which does the following.
1. Initializes and provides access to ODPEventManager and ODPSegmentManager.
2. Provides updated ODPConfig settings to event manager and segment manager.
3. Stops Event Manager when closed.
4. Uses VUID for browser flavor of javascript SDK.

## Test plan
1. Manually tested thoroughly
2. Added unit tests.

## Issues
[FSSDK-8452](https://jira.sso.episerver.net/browse/FSSDK-8452)


packages/optimizely-sdk/lib/core/odp/odp_config.ts
```diff
@@ -57,26 +59,24 @@ export class OdpConfig {
     return this._segmentsToCheck;
   }
 
-  constructor(apiKey: string, apiHost: string, segmentsToCheck?: string[]) {
-    this._apiKey = apiKey;
-    this._apiHost = apiHost;
+  constructor(apiKey?: string, apiHost?: string, segmentsToCheck?: string[]) {
+    if (apiKey) this._apiKey = apiKey;
+    if (apiHost) this._apiHost = apiHost;
```
mikechu-optimizely, line 64: Could we use `??` with your default `''` ?

packages/optimizely-sdk/lib/core/odp/odp_config.ts
```diff
@@ -57,26 +59,24 @@ export class OdpConfig {
...
+  public update(config: OdpConfig): boolean {
+    if (this.equals(config)) {
       return false;
     } else {
-      this._apiKey = apiKey;
-      this._apiHost = apiHost;
-      this._segmentsToCheck = segmentsToCheck;
+      if (config.apiKey) this._apiKey = config.apiKey;
+      if (config.apiHost) this._apiHost = config.apiHost;
+      if (config.segmentsToCheck) this._segmentsToCheck = config.segmentsToCheck;
```
mikechu-optimizely, line 79: Could we use something like `this._apiKey = config?.apiKey ?? ''`

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
```diff
@@ -0,0 +1,228 @@
...
+  enabled: boolean;
+  logger: LogHandler;
+  odpConfig!: OdpConfig;
+
+  /**
+   * ODP Segment Manager which provides an interface to the remote ODP server (GraphQL API) for audience segments mapping.
+   * It fetches all qualified segments for the given user context and manages the segments cache for all user contexts.
+   * @private
+   */
+  private _segmentManager!: OdpSegmentManager;
```
mikechu-optimizely, line 71: Is it okay to use the non-null assertion operator here on these fields?

I've not seen this done before.

packages/optimizely-sdk/lib/core/project_config/index.ts
```diff
@@ -167,13 +167,13 @@ export const createProjectConfig = function (
...
+  projectConfig.allSegments = []
 
   Object.keys(projectConfig.audiencesById)
     .map((audience) => getAudienceSegments(projectConfig.audiencesById[audience]))
     .forEach(audienceSegments => {
       audienceSegments.forEach(segment => {
-        projectConfig.allSegments.add(segment)
+        projectConfig.allSegments.push(segment)
       })
     })
```
mikechu-optimizely, line 178: NIT: I think we may need to add the `;` at least for consistency.

packages/optimizely-sdk/tests/odpSegmentApiManager.spec.ts
```diff
@@ -161,7 +162,7 @@ describe('OdpSegmentApiManager', () => {
 
     const segments = await manager.fetchSegments(API_key, GRAPHQL_ENDPOINT, ODP_USER_KEY.FS_USER_ID, USER_VALUE, []);
 
-    expect(segments).toHaveLength(0);
+    if (segments) expect(segments.length).toEqual(0);
```
mikechu-optimizely, line 165: Does `segments?.length` work here similarly to L152?

packages/optimizely-sdk/tests/odpSegmentApiManager.spec.ts
```diff
@@ -174,7 +175,7 @@ describe('OdpSegmentApiManager', () => {
 
     const segments = await manager.fetchSegments(API_key, GRAPHQL_ENDPOINT, USER_KEY, USER_VALUE, SEGMENTS_TO_CHECK);
 
-    expect(segments).toHaveLength(0);
+    if (segments) expect(segments.length).toEqual(0);
```
mikechu-optimizely, line 178: Same question as on L165.

packages/optimizely-sdk/tests/vuidManager.spec.ts
```diff
@@ -50,9 +50,9 @@ describe('VuidManager', () => {
   it('should test if a VUID is valid', async () => {
     const manager = await VuidManager.instance(instance(mockCache));
 
-    expect(manager['isVuid']('vuid_123')).toBe(true);
-    expect(manager['isVuid']('vuid-123')).toBe(false);
-    expect(manager['isVuid']('123')).toBe(false);
+    expect(VuidManager.isVuid('vuid_123')).toBe(true);
+    expect(VuidManager.isVuid('vuid-123')).toBe(false);
+    expect(VuidManager.isVuid('123')).toBe(false);
```
mikechu-optimizely, line 55: This is much nicer.  😁 👍

mikechu-optimizely (changes requested):
I have a few small questions listed.

packages/optimizely-sdk/lib/core/odp/odp_config.ts
opti-jnguyen: @mikechu-optimizely - Agree with using `??` in the constructor, however using `config?.apiKey ?? ''` wouldn't work if, say, we only wanted to update the `apiHost` but not the `apiKey`, right?

i.e. with `?? ''`

Before:
```
apiKey = 'key-a'
apiHost = 'host-a'
```

```
update({
  apiHost: 'host-b'
})
```

After:
```
apiKey = ''
apiHost='host-b'
```

This doesn't seem to be the desired behavior so I'd rather keep the current approach unless there's other reasoning, or perhaps we wanted to consider refactoring all fields to be mandatory.

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
opti-jnguyen: @mikechu-optimizely - good question.

The usage of it here is to skirt around TypeScript complaining about the short circuit return in `if (!this.enabled)` never leading to `_segmentManager` and `_eventManager` being instantiated.

If we remove the non-null assertion, TypeScript demands instantiation which doesn't seem necessary if ODPManager is not enabled. Let me know if that makes sense or if you have further thoughts - thanks!

packages/optimizely-sdk/tests/odpSegmentApiManager.spec.ts
opti-jnguyen: Yes, you're right. Will patch those - thanks!

packages/optimizely-sdk/lib/core/project_config/index.ts
opti-jnguyen: Nice catch, forgot to run format save on that one.

packages/optimizely-sdk/lib/core/odp/odp_config.ts
mikechu-optimizely: Good point.  We want to selectively update.

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
mikechu-optimizely: Can we use a union type?  `OdpSegmentManager | undefined`?   Just thinking "out loud".  We can DM over this too.

packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -0,0 +1,164 @@
...
+    const browserClientVersion = packageJSON.version;
+
+    super({
+      disable,
+      requestHandler: browserRequestHandler,
+      logger: browserLogger,
+      clientEngine: browserClientEngine,
+      clientVersion: browserClientVersion,
+      odpConfig,
+      segmentsCache,
```
jaeopt, line 66:
```suggestion
      segmentsCache || BrowerLRUCache(),
```
Or not sure if you have a plan to inject BrowserLRUCache() always.

packages/optimizely-sdk/lib/plugins/odp_manager/index.node.ts
```diff
@@ -0,0 +1,58 @@
...
+    const nodeClientEngine = NODE_CLIENT_ENGINE;
+    const nodeClientVersion = NODE_CLIENT_VERSION;
+
+    super({
+      disable,
+      requestHandler: nodeRequestHandler,
+      logger,
+      clientEngine: nodeClientEngine,
+      clientVersion: nodeClientVersion,
+      segmentsCache,
```
jaeopt, line 53:
```suggestion
      segmentsCache || ServerLRUCache,
```

packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -168,13 +174,27 @@ export default class Optimizely {
...
+        this.odpManager.eventManager?.start();
+        this.updateODPSettings();
+        const sdkKey = this.projectConfigManager.getConfig()?.sdkKey;
+        if (sdkKey != null) {
+          NotificationRegistry.getNotificationCenter(sdkKey, this.logger)
+            ?.addNotificationListener(enums.NOTIFICATION_TYPES.OPTIMIZELY_CONFIG_UPDATE, () => this.updateODPSettings());
+        } else {
+          this.logger.log(LOG_LEVEL.ERROR, ERROR_MESSAGES.ODP_SDK_KEY_MISSING_NOTIFICATION_CENTER_FAILURE);
+        }
+      }
```
jaeopt, line 189: What if config does not have odpManager? Is it coming in a separate PR?

jaeopt (changes requested):
Looks good! A few changes suggested.

packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -0,0 +1,164 @@
...
+      eventManager,
+      segmentManager,
+    });
+
+    this.logger = browserLogger;
+
+    // TODO: Think of better way to trigger initializeVuid after this.eventManager is guaranteed to exist. Promise?
+    setTimeout(() => {
```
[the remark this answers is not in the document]
opti-jnguyen: Been trying to wrap my head around this for the past day or so - both manager eventManager should be ready at this point, but I've been having issues verifying that the eventApiManager's `sendEvent` is successfully called in the test suite, even when manually flushing the queue to trigger sendEvent.

Will push what I have for now and keep on working at getting these last tests to pass, I'm probably missing something obvious.

packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -0,0 +1,164 @@
...
+      if (VuidManager.isVuid(fsUserId)) {
+        vuid = fsUserId;
+        fsUserId = undefined;
+      }
+      this.eventManager.identifyUser(vuid, fsUserId);
+    } catch (e) {
+      this.logger.log(LogLevel.ERROR, LOG_MESSAGES.ODP_IDENTIFY_FAILED_VUID_MISSING);
+    }
+  }
+
```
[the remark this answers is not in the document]
opti-jnguyen: Note: Extending a method in TS forces the parameters to remain the same between parent and child classes.

In this case, `userKey` will be kept, though it won't be used since we'll know if the key should either be VUID or FS_USER_ID based on .isVuid. Thoughts on making this a new method instead of extending the parent?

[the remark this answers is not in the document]
jaeopt: @opti-jnguyen I see that a separate API (userId) will be clean here, but OptimizelyUserContext should call different APIs for browser and node. I think we can keep the current (userKey, userId) though userKey is ignored here.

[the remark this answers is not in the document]
jaeopt: @opti-jnguyen Alternatively, we can change (userKey, userId) to (userId) in the parent OdpManager and decide "userKey" inside the parent OdpManager using 'isVuid`.

packages/optimizely-sdk/lib/optimizely/index.ts
opti-jnguyen: @jaeopt - those changes will be reserved for https://jira.sso.episerver.net/browse/FSSDK-8475 which includes User Context and Optimizely Client changes.

packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -96,29 +113,14 @@ export class BrowserOdpManager extends OdpManager {
...
-      this.logger.log(LogLevel.DEBUG, LOG_MESSAGES.ODP_IDENTIFY_FAILED_ODP_DISABLED);
-      return;
-    }
-
-    if (!this.odpConfig.isReady()) {
-      this.logger.log(LogLevel.DEBUG, LOG_MESSAGES.ODP_IDENTIFY_FAILED_ODP_NOT_INTEGRATED);
+  public identifyUser(userId: string): void {
+    if (VuidManager.isVuid(userId)) {
+      this.vuid = userId;
+      super.identifyUser('', userId);
```
jaeopt, line 119: We should not send empty fsUserId string. Please confirm this empty fsUserId is filtered out from the final event.

packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -96,29 +113,14 @@ export class BrowserOdpManager extends OdpManager {
...
-    if (!this.enabled) {
-      this.logger.log(LogLevel.DEBUG, LOG_MESSAGES.ODP_IDENTIFY_FAILED_ODP_DISABLED);
-      return;
-    }
-
-    if (!this.odpConfig.isReady()) {
-      this.logger.log(LogLevel.DEBUG, LOG_MESSAGES.ODP_IDENTIFY_FAILED_ODP_NOT_INTEGRATED);
+  public identifyUser(userId: string): void {
+    if (VuidManager.isVuid(userId)) {
+      this.vuid = userId;
```
jaeopt, line 118: I think we should not override the vuid. Other modules can call this api with other vuid values (different from the device vuid) as well (not likely, but possible in future).

jaeopt (changes requested):
All changes look great. Some small changes suggested.

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
```diff
@@ -119,10 +105,10 @@ export class OdpManager {
 
     // Set up Events Manager (Events REST API Interface)
     if (eventManager) {
-      this._eventManager = eventManager;
-      this._eventManager.updateSettings(this.odpConfig);
+      this.eventManager = eventManager;
+      this.eventManager.updateSettings(this.odpConfig);
```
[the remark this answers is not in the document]
opti-jnguyen: Went back and changed line 97 to match instead to acknowledge the `private` nature of event manager's ODP Config.

Refactored Segment Manager's ODP Config to be a `private` variable instead.

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
```diff
@@ -169,16 +182,24 @@ export class OdpManager {
       return null;
     }
 
-    return this.segmentManager.fetchQualifiedSegments(userKey, userId, options);
+    if (!this.segmentManager) {
+      throw new Error(ERROR_MESSAGES.ODP_FETCH_QUALIFIED_SEGMENTS_SEGMENTS_MANAGER_MISSING);
+    }
+
+    if (VuidManager.isVuid(userId)) {
```
jaeopt, line 189: I like this - "isVuid" check in odp_manager (instead of index.browser.ts), so we can remove most extra logic from index.browser.ts and index.node.ts.
I suggest to do the same thing for "identifyUser(userId)" (remove "vuid" from parameters) and isVuid differentiation here. It'll help to make it clean and remove extra codes from index.browser.ts and index.node.ts.

jaeopt (changes requested):
A couple of more suggestions.

packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -78,49 +78,32 @@ export class BrowserOdpManager extends OdpManager {
...
-   * @param userId Unique identifier of a target user.
+   * @param fsUserId Unique identifier of a target user.
    */
-  public identifyUser(userId: string): void {
-    if (VuidManager.isVuid(userId)) {
-      this.vuid = userId;
-      super.identifyUser('', userId);
+  public identifyUser(fsUserId?: string, vuid?: string): void {
+    if (fsUserId && VuidManager.isVuid(fsUserId)) {
+      this.vuid = fsUserId;
```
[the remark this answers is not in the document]
opti-jnguyen: Makes sense, will do!

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
opti-jnguyen: Agreed, will make the delegation of VUID check consistent - thanks!

jaeopt (approved):
LGTM

opti-jnguyen:
Merging without passing FSC for velocity, will address at a later point.

PR #805, opened 2023-03-17
Update OdpOptions to handle additional granular configuration options for ODP Segment and Event Managers

opti-jnguyen:
## Summary
- Add additional configuration options in `OdpOptions` to support granular configuration of:
  - `segmentsApiTimeout`
  - `segmentsRequestHandler`
  - `eventApiTimeout`
  - `eventFlushInterval`
  - `eventQueueSize`
  - `eventRequestHandler`

## Test plan
- TODO: Test permutations of new `OdpOptions` passed into the `createBrowserOdpManager()` method.

## Issues
- [FSSDK-8986](https://jira.sso.episerver.net/browse/FSSDK-8986)

Will unblock:
- [FSSDK-8793](https://jira.sso.episerver.net/browse/FSSDK-8793)
- [FSSDK-8690](https://jira.sso.episerver.net/browse/FSSDK-8690)

### DIFF
--- a/packages/optimizely-sdk/lib/core/odp/odp_event_manager.ts
+++ b/packages/optimizely-sdk/lib/core/odp/odp_event_manager.ts
@@ -143,6 +143,7 @@ export class OdpEventManager implements IOdpEventManager {
     let defaultQueueSize = DEFAULT_BROWSER_QUEUE_SIZE;
 
     try {
+      // TODO: Consider refactoring to use typeof process and combine w/above line
       if (process) {
         defaultQueueSize = DEFAULT_SERVER_QUEUE_SIZE;
       }
--- a/packages/optimizely-sdk/lib/core/odp/odp_manager.ts
+++ b/packages/optimizely-sdk/lib/core/odp/odp_manager.ts
@@ -19,8 +19,9 @@ import { getLogger, LogHandler, LogLevel } from '../../modules/logging';
 import { ERROR_MESSAGES, ODP_USER_KEY } from '../../utils/enums';
 
 import { RequestHandler } from './../../utils/http_request_handler/http';
+import { BrowserRequestHandler } from '../../utils/http_request_handler/browser_request_handler';
+
 import { BrowserLRUCache } from './../../utils/lru_cache/browser_lru_cache';
-import { LRUCache } from './../../utils/lru_cache/lru_cache';
 
 import { VuidManager } from '../../plugins/vuid_manager';
 
@@ -32,6 +33,7 @@ import { OdpEventApiManager } from './odp_event_api_manager';
 import { OptimizelySegmentOption } from './optimizely_segment_option';
 import { invalidOdpDataFound } from './odp_utils';
 import { OdpEvent } from './odp_event';
+import { OdpOptions } from '../../shared_types';
 
 /**
  * @param {boolean}                     disable Flag for disabling ODP Manager.
@@ -45,13 +47,11 @@ import { OdpEvent } from './odp_event';
  */
 interface OdpManagerConfig {
   disable: boolean;
-  requestHandler: RequestHandler;
+  defaultRequestHandler: RequestHandler;
   logger?: LogHandler;
   clientEngine?: string;
   clientVersion?: string;
-  segmentsCache?: LRUCache<string, string[]>;
-  eventManager?: OdpEventManager;
-  segmentManager?: OdpSegmentManager;
+  odpOptions?: OdpOptions;
 }
 
 /**
@@ -74,47 +74,53 @@ export class OdpManager {
    */
   public eventManager: OdpEventManager | undefined;
 
-  constructor({
-    disable,
-    requestHandler,
-    logger,
-    clientEngine,
-    clientVersion,
-    segmentsCache,
-    eventManager,
-    segmentManager,
-  }: OdpManagerConfig) {
+  constructor({ disable, defaultRequestHandler, logger, clientEngine, clientVersion, odpOptions }: OdpManagerConfig) {
     this.enabled = !disable;
     this.logger = logger || getLogger();
 
-    if (!this.enabled) {
+    if (disable) {
       this.logger.log(LogLevel.INFO, LOG_MESSAGES.ODP_DISABLED);
       return;
     }
 
     // Set up Segment Manager (Audience Segments GraphQL API Interface)
-    if (segmentManager) {
-      this.segmentManager = segmentManager;
+    if (odpOptions?.segmentManager) {
+      this.segmentManager = odpOptions.segmentManager;
       this.segmentManager.updateSettings(this.odpConfig);
     } else {
       this.segmentManager = new OdpSegmentManager(
         this.odpConfig,
-        segmentsCache || new BrowserLRUCache<string, string[]>(),
-        new OdpSegmentApiManager(requestHandler, this.logger)
+        odpOptions?.segmentsCache ||
+          new BrowserLRUCache<string, string[]>({
+            maxSize: odpOptions?.segmentsCacheSize,
+            timeout: odpOptions?.segmentsCacheTimeout,
+          }),
+        new OdpSegmentApiManager(odpOptions?.segmentsRequestHandler || defaultRequestHandler, this.logger)
       );
     }
 
     // Set up Events Manager (Events REST API Interface)
-    if (eventManager) {
-      this.eventManager = eventManager;
+    if (odpOptions?.eventManager) {
+      this.eventManager = odpOptions.eventManager;
       this.eventManager.updateSettings(this.odpConfig);
     } else {
+      let customEventRequestHandler;
+
+      if (odpOptions?.eventRequestHandler) {
+        customEventRequestHandler = odpOptions.eventRequestHandler;
+      } else if (odpOptions?.eventApiTimeout) {
+        customEventRequestHandler = new BrowserRequestHandler(this.logger, odpOptions.eventApiTimeout);
+      }
+
       this.eventManager = new OdpEventManager({
         odpConfig: this.odpConfig,
-        apiManager: new OdpEventApiManager(requestHandler, this.logger),
+        apiManager: new OdpEventApiManager(customEventRequestHandler || defaultRequestHandler, this.logger),
         logger: this.logger,
         clientEngine: clientEngine || 'javascript-sdk',
         clientVersion: clientVersion || BROWSER_CLIENT_VERSION,
+        flushInterval: odpOptions?.eventFlushInterval,
+        batchSize: odpOptions?.eventBatchSize,
+        queueSize: odpOptions?.eventQueueSize,
       });
     }
 
--- a/packages/optimizely-sdk/lib/core/odp/odp_manager.ts
+++ b/packages/optimizely-sdk/lib/core/odp/odp_manager.ts
@@ -47,7 +47,8 @@ import { OdpOptions } from '../../shared_types';
  */
 interface OdpManagerConfig {
   disable: boolean;
-  defaultRequestHandler: RequestHandler;
+  segmentRequestHandler: RequestHandler;
+  eventRequestHandler: RequestHandler;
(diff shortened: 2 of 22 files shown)


packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -134,6 +137,41 @@ export class BrowserOdpManager extends OdpManager {
       return new BrowserOdpManager({ disable: false, logger });
     }
 
+    let browserSegmentManager;
+    if (odpOptions.segmentsRequestHandler || odpOptions.segmentsApiTimeout) {
+      browserSegmentManager = new OdpSegmentManager(
+        new OdpConfig(),
+        new BrowserLRUCache<string, string[]>(),
```
jaeopt, line 144:
```suggestion
        odpOptions?.segmentsCache || new BrowserLRUCache<string, string[]>(maxSize: odpOptions.segmentsCacheSize,
          timeout: odpOptions.segmentsCacheTimeout,),
```

It's possible they want to config Cache + ApiTimeout.

packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -134,6 +137,41 @@ export class BrowserOdpManager extends OdpManager {
       return new BrowserOdpManager({ disable: false, logger });
     }
 
+    let browserSegmentManager;
+    if (odpOptions.segmentsRequestHandler || odpOptions.segmentsApiTimeout) {
+      browserSegmentManager = new OdpSegmentManager(
+        new OdpConfig(),
```
jaeopt, line 143: This dummy odpConfig injection which will be overwritten on OdpManager constructor sounds odd.
We can consider to remove this parameter and instead, change to always call "updateSettings(opdConfig)" when odpEventManager and odpSegmentManager are instantiated.

packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -134,6 +137,41 @@ export class BrowserOdpManager extends OdpManager {
...
+
+    let browserEventManager;
+    if (
+      odpOptions.eventRequestHandler ||
+      odpOptions.eventApiTimeout ||
+      odpOptions.eventFlushInterval ||
+      odpOptions.eventBatchSize ||
+      odpOptions.eventQueueSize
+    ) {
+      browserEventManager = new OdpEventManager({
```
jaeopt, line 160: We repeat this same EventManager instantiation in OdpManager as well, which looks redundant. Line 58 and 62 passing `RequestHandler` and `cache` also look redundant. We can consider passing `options` to `OdpManager` where we build all `OdpEventManager` and `OdpSegmentManager` in there. Let's discuss about other options.

packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -134,6 +137,41 @@ export class BrowserOdpManager extends OdpManager {
...
+        apiManager: new OdpEventApiManager(
+          odpOptions.eventRequestHandler || new BrowserRequestHandler(logger, odpOptions.eventApiTimeout),
+          logger
+        ),
+        logger: logger,
+        clientEngine: 'javascript-sdk',
+        clientVersion: BROWSER_CLIENT_VERSION,
+        flushInterval: odpOptions.eventFlushInterval,
+        batchSize: odpOptions.eventBatchSize,
+        queueSize: odpOptions.eventQueueSize,
```
jaeopt, line 171: These size and timeout are platform dependent (browser or node). Do we determine their default values (when not defined in odpOptions) in the `OdpEventManager`, depending on the target platform? If so, the value of this separate `browser-odp-manger` here is not that significant.

jaeopt (changes requested):
It looks good, with full flexibility in config supported.  A concern is that I see redundant event and segment manager instantiations in multiple levels. Let's discuss about options to clean them up.

packages/optimizely-sdk/lib/index.browser.ts
```diff
@@ -125,7 +125,7 @@ const createInstance = function(config: Config): Client | null {
         : undefined,
       notificationCenter,
       isValidInstance: isValidInstance,
-      odpManager: BrowserOdpManager.createBrowserOdpManager({ logger, odpOptions: config.odpOptions }),
+      odpManager: new BrowserOdpManager({ disable: true, logger, odpOptions: config.odpOptions }),
```
jaeopt, line 128: Why disable is true by default?  It is a part of odpOptions. Do we need this separate parameter?

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
```diff
@@ -74,47 +74,53 @@ export class OdpManager {
...
-    requestHandler,
-    logger,
-    clientEngine,
-    clientVersion,
-    segmentsCache,
-    eventManager,
-    segmentManager,
-  }: OdpManagerConfig) {
+  constructor({ disable, defaultRequestHandler, logger, clientEngine, clientVersion, odpOptions }: OdpManagerConfig) {
     this.enabled = !disable;
```
jaeopt, line 78: Can `odpOptions.disable` override this?  Still wondering if we need this separate `disable` parameter.

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
```diff
@@ -74,47 +74,53 @@ export class OdpManager {
...
+    if (odpOptions?.segmentManager) {
+      this.segmentManager = odpOptions.segmentManager;
       this.segmentManager.updateSettings(this.odpConfig);
     } else {
       this.segmentManager = new OdpSegmentManager(
         this.odpConfig,
-        segmentsCache || new BrowserLRUCache<string, string[]>(),
-        new OdpSegmentApiManager(requestHandler, this.logger)
+        odpOptions?.segmentsCache ||
+          new BrowserLRUCache<string, string[]>({
```
jaeopt, line 94: If we have `BrowserLRUCache` here at the common odp_manager, how can we support the node version?
We may use `LRUCache` (instead of `BrowserLRUCache`) since it can be different from the node version only by the size and timeout. We can feed target-specific size and timeout form `browser_odp_manager`.

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
```diff
@@ -74,47 +74,53 @@ export class OdpManager {
...
+            maxSize: odpOptions?.segmentsCacheSize,
+            timeout: odpOptions?.segmentsCacheTimeout,
+          }),
+        new OdpSegmentApiManager(odpOptions?.segmentsRequestHandler || defaultRequestHandler, this.logger)
```
jaeopt, line 98: I do not see support for `odpOptions.segmentApiTimeout`.

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
```diff
@@ -74,47 +74,53 @@ export class OdpManager {
...
+    if (odpOptions?.eventManager) {
+      this.eventManager = odpOptions.eventManager;
       this.eventManager.updateSettings(this.odpConfig);
     } else {
+      let customEventRequestHandler;
+
+      if (odpOptions?.eventRequestHandler) {
+        customEventRequestHandler = odpOptions.eventRequestHandler;
+      } else if (odpOptions?.eventApiTimeout) {
+        customEventRequestHandler = new BrowserRequestHandler(this.logger, odpOptions.eventApiTimeout);
```
jaeopt, line 112: Same here. It's awkward to see `Browser-xxx` at this level. We can push this custom requesthandler to browser-odp-manager.

jaeopt (changes requested):
Looks much cleaner! A few more suggestions.

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
opti-jnguyen: True, the disable param is redundant - will look into moving all disable logic into `odpOptions`.

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
opti-jnguyen: @jaeopt - I defaulted to `BrowserLRUCache` for now as LRUCache has no defaults for size and timeout.

Two possible approaches:
1. Do something similar to `defaultRequestHandler` and have a `defaultLRUCache` passed down from the implementation level as a required field.
2. Instantiate with LRUCache and consider setting the defaults to be (minimum?) values.

packages/optimizely-sdk/lib/core/odp/odp_manager.ts
opti-jnguyen: Oops - let me add that in with the same logic as the custom Event Request Handler.

packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -40,16 +39,37 @@ export class BrowserOdpManager extends OdpManager {
...
-      disable,
-      defaultRequestHandler: browserDefaultRequestHandler,
+      segmentLRUCache:
+        odpOptions?.segmentsCache ||
+        new BrowserLRUCache<string, string[]>({
+          maxSize: odpOptions?.segmentsCacheSize,
+          timeout: odpOptions?.segmentsCacheTimeout,
+        }),
+      segmentRequestHandler: customSegmentRequestHandler || new BrowserRequestHandler(browserLogger),
+      eventRequestHandler: new BrowserRequestHandler(browserLogger),
```
jaeopt, line 72:
```suggestion
      eventRequestHandler: customEventRequestHandler || new BrowserRequestHandler(browserLogger),
```

jaeopt: Do we have all test cases covering different config options?  It won't be covered by FSC, so please make sure we have full coverage of config combination testing.

packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -40,16 +39,37 @@ export class BrowserOdpManager extends OdpManager {
...
+    } else if (odpOptions?.eventApiTimeout) {
+      customEventRequestHandler = new BrowserRequestHandler(browserLogger, odpOptions.eventApiTimeout);
+    }
+
     super({
-      disable,
-      defaultRequestHandler: browserDefaultRequestHandler,
+      segmentLRUCache:
+        odpOptions?.segmentsCache ||
+        new BrowserLRUCache<string, string[]>({
```
jaeopt, line 67: We do not have to fix for now, but I suggest to remove `BrowserLRUCache` and `NodeLRUCache` since they're identical in implementation with size/timeout injected.

jaeopt (changes requested):
Looks good! A couple of suggestions.

packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
opti-jnguyen: Agreed, perhaps the variant-specific default values can be handled by the variant-specific OdpManager implementations in the future instead.

packages/optimizely-sdk/lib/core/odp/odp_event_manager.ts
```diff
@@ -143,6 +143,7 @@ export class OdpEventManager implements IOdpEventManager {
     let defaultQueueSize = DEFAULT_BROWSER_QUEUE_SIZE;
 
     try {
+      // TODO: Consider refactoring to use typeof process and combine w/above line
```
jaeopt, line 146: Alternatively, we can later change to push variant size down from the browser-odp-manager level

jaeopt (approved):
LGTM!

PR #806, opened 2023-03-17
[FSSDK-8994] Fix Identify User

opti-jnguyen:
## Summary
- Patched the `identifyUser()` method to also pass in the `vuid` along with `fs_user_id` in the initial call when instantiating a new user context with a non-vuid identifier.

## Test plan

## Issues
- [FSSDK-8994](https://jira.sso.episerver.net/browse/FSSDK-8994)

### DIFF
--- a/packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
+++ b/packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
@@ -1,4 +1,3 @@
-import { LOG_LEVEL } from './../../utils/enums/index';
 /**
  * Copyright 2023, Optimizely
  *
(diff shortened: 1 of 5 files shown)


packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -96,13 +97,23 @@ export class BrowserOdpManager extends OdpManager {
...
+      return super.identifyUser(undefined, fsUserId);
     }
 
-    super.identifyUser(fsUserId, vuid);
+    if (fsUserId && vuid && VuidManager.isVuid(vuid)) {
+      return super.identifyUser(fsUserId, vuid);
+    }
+
+    try {
+      const vuidManager = await VuidManager.instance(BrowserOdpManager.cache);
```
jaeopt, line 110: Can't we reuse the `vuid` in line 77?

packages/optimizely-sdk/lib/plugins/odp_manager/index.browser.ts
```diff
@@ -96,13 +97,23 @@ export class BrowserOdpManager extends OdpManager {
...
+
+      return super.identifyUser(fsUserId, vuid || vuidManager.vuid);
+    } catch (e) {
+      this.logger.log(LogLevel.WARNING, ERROR_MESSAGES.ODP_IDENTIFY_USER_FAILED);
+      return;
+    }
```
jaeopt, line 116: Then, we may not need this error case.

jaeopt (changes requested):
Looks good. Clarification for cleaning up.

jaeopt (approved):
LGTM - remove async from identifyUser

PR #809, opened 2023-03-23
feat(ats): [FSSDK-8793] add support for disabling odp event batching

jaeopt:
## Summary
Allow to disable ODP event batching by setting flushInterval to zero.

## Test plan
- Test flushInterval and batchSize when configured with odpOptions.flushInterval = 0

## Issues
- [FSSDK-8793](https://jira.sso.episerver.net/browse/FSSDK-8793)


opti-jnguyen:
Note that for the JS Browser implementation, batchSize will be hard-coded to `1` going forward due to the new requirement of using the Zaius Pixel API endpoint which doesn't support batched event payloads.

opti-jnguyen (approved):

PR #812, opened 2023-03-30
fix(ats): [FSSDK-9023] Fix ODP Exports

opti-jnguyen:
## Summary
- Patched `fetchQualifiedSegments` method signature in the `OptimizelyUserContext` interface export to include optional parameter `options` to match expected external-facing interface for `OptimizelyUserContext`.
- Added `qualifiedSegments` to `OptimizelyUserContext` interface to match expected exported interface for external-facing interface for `OptimizelyUserContext`.
- Refactored `OptimizelyUserContext` in `shared_types.ts` and `optimizely_user_context/index.ts` to couple class implementation with interface in order to avoid future interface mismatches. Renamed the interface to `IOptimizelyUserContext` while keeping the export name the same as before (`OptimizelyUserContext`) in order to avoid external regression.
- Added enum `OptimizelySegmentsOption` to exported values to be exposed to external consumers.
- Added `BrowserClient` to exported values to be exposed to external consumers.

The above changes fix the SDK to expose the expected interfaces to consumers for `fetchQualifiedSegments`, `qualifiedSegments`, `OptimizelySegmentsOption`, and `BrowserClient`.

## Test plan

- All existing tests should pass. No tests within this SDK should be affected, though this will enable corresponding FSC tests to be implemented.

## Issues
- [FSSDK-9023](https://jira.sso.episerver.net/browse/FSSDK-9023)


packages/optimizely-sdk/lib/optimizely/index.ts
```diff
@@ -1662,23 +1682,17 @@ export default class Optimizely {
...
-    action,
-    identifiers,
-    data,
-  }: {
-    type?: string;
-    action: ODP_EVENT_ACTION;
-    identifiers?: Map<string, string>;
-    data?: Map<string, unknown>;
-  }): Promise<void> {
+  public async sendOdpEvent(
```
jaeopt, line 1690: It looks like we need to make it "sync".

jaeopt: line 1705, it await for sync odpManager.sendOdpManager. Is it allowed or maybe some interface error?

jaeopt (changes requested):
Looks good. A quick for sendOdpEvent async.

packages/optimizely-sdk/lib/optimizely/index.ts
jaeopt: I see "async" on identifyUser as well. It can be cleaned up too.

jaeopt (changes requested):
One more async issue.

PR #820, opened 2023-05-02
[FSSDK-9152] Fix "client_initialized" event not firing

mikechu-optimizely:
## Summary
- Fix "client_initialized" event not firing

## Test plan

- Needs a more detailed set of tests wrapped around for coverage. 
- Existing tests should continue to succeed along with FSC

## Issues
[FSSDK-9152](https://jira.sso.episerver.net/browse/FSSDK-9152)

### DIFF
--- a/packages/optimizely-sdk/lib/index.browser.tests.js
+++ b/packages/optimizely-sdk/lib/index.browser.tests.js
@@ -907,7 +907,7 @@ describe('javascript-sdk (Browser)', function() {
       it('should send odp client_initialized on client instantiation', async () => {
         const odpConfig = new OdpConfig();
         const apiManager = new BrowserOdpEventApiManager(mockRequestHandler, logger);
-        apiManager.sendEvents = sinon.spy();
+        sinon.spy(apiManager, 'sendEvents');
         const eventManager = new BrowserOdpEventManager({
            odpConfig,
            apiManager,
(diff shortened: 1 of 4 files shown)


packages/optimizely-sdk/lib/index.browser.tests.js
```diff
@@ -903,6 +903,35 @@ describe('javascript-sdk (Browser)', function() {
...
+            odpConfig,
+            eventManager,
+          },
+        });
+
+        const readyData = await client.onReady();
+        assert.equal(readyData.success, true);
+        assert.isUndefined(readyData.reason);
+
+        setTimeout(() => sinon.assert.callCount(apiManager.sendEvents, 1), 100);
```
mikechu-optimizely, line 933: This really needs to check sendEvents(events) has a "client_initialized" in the array.

jaeopt: I believe we have a separate unit test for "registerVuid" which validates "client_initialized" event. If so, this may be enough to validate we have one event generated here.

jaeopt (approved):
LGTM

mikechu-optimizely:
@opti-jnguyen I found the test that was failing (doing its job dutifully despite me ignoring it).  I'm good to go especially with @jaeopt 's approval.

opti-jnguyen (approved):
LGTM overall. Agree with your note on the test, you may be able to grab the args from the spy and do an assert equal.

Maybe something along the lines of:
```
const sendEventArgs1 = fakeEventManager.sendEvent.args;
assert.deepEqual(
  sendEventArgs1[0].toString(),
  new OdpEvent('fullstack', 'client_initialized', new Map([['fs_user_id', 'fsUserA']]), new Map()).toString()
);
```

That being said, if that doesn't work I'm down to merge and address later.

PR #822, opened 2023-05-03
[FSSDK-9096] chore: Prepare release 5.0.0-beta

opti-jnguyen:
## Summary
- Incremented Optimizely SDK version from 4.9.3 to 5.0.0-beta.1

Prepare for first v5 Beta Pre-release.

## Test plan
- New tests introduced in this version should also pass.
- The known FSC failure will be addressed in a future patch.

## Issues
- [FSSDK-9096](https://jira.sso.episerver.net/browse/FSSDK-9096)

### DIFF
--- a/packages/optimizely-sdk/CHANGELOG.md
+++ b/packages/optimizely-sdk/CHANGELOG.md
@@ -7,12 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
-## 5.0.0-beta.1
+## 5.0.0-beta
 May 4, 2023
 
 ### New Features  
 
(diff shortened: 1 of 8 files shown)


packages/optimizely-sdk/CHANGELOG.md
```diff
@@ -7,7 +7,49 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
-### Breaking Changes
+## 5.0.0-beta.1
```
jaeopt, line 10: `5.0.0-beta` to be consistent with other sdks?

jaeopt (changes requested):
Can we rename to 5.0.0-beta to be consistent with others? 
I expect that this PR will pass all FSC tests, correct?

packages/optimizely-sdk/CHANGELOG.md
opti-jnguyen: Updated to be consistent. This PR should pass all FSC tests minus two which involve firing queued events upon config change.

Upon further investigation, unsure if the FSC test should apply to the Browser variant as events are always immediately sent in a fire-and-forget method due to use of the Pixel API. Either way, will be publishing the Beta version as is, noting down all issues such as those, and addressing them in future patches.

jaeopt (approved):
LGTM

PR #823, opened 2023-05-04
[FSSDK-8950] change identifiers to be required for sendOdpEvent

mikechu-optimizely:
## Summary
- SendODPEvent API has "identifiers" parameter as optional. For server-sdks (aka Node context), this should be changed to required (at least one key-value pair).

## Test plan

- 2 new unit tests added
- Existing unit tests should continue to pass
- FSC tests should pass

## Issues
- [FSSDK-8950](https://jira.sso.episerver.net/browse/FSSDK-8950)

### DIFF
--- a/packages/optimizely-sdk/lib/plugins/odp/event_manager/index.browser.ts
+++ b/packages/optimizely-sdk/lib/plugins/odp/event_manager/index.browser.ts
@@ -1,5 +1,6 @@
 import { IOdpEventManager, OdpEventManager } from '../../../../lib/core/odp/odp_event_manager';
 import { LogLevel } from '../../../modules/logging';
+import {OdpEvent} from "../../../core/odp/odp_event";
 
 const DEFAULT_BROWSER_QUEUE_SIZE = 100;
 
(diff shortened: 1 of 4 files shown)


opti-jnguyen (approved):
A couple nits, but overall LGTM!

packages/optimizely-sdk/lib/plugins/odp/event_manager/index.browser.ts
```diff
@@ -28,4 +29,6 @@ export class BrowserOdpEventManager extends OdpEventManager implements IOdpEvent
     // in Browser/client-side context, give debug message but leave events in queue
     this.getLogger().log(LogLevel.DEBUG, 'ODPConfig not ready. Leaving events in queue.');
   }
+
+  protected hasNecessaryIdentifiers = (event: OdpEvent): boolean => event.identifiers.size >= 0;
```
jaeopt, line 33: browser version is expected to insert vuid automatically to all events. I expect that this check is not necessary. I cannot find this auto insertion for browser. @opti-jnguyen can you check?

jaeopt (commented):
Can we check if we have a separate requirement for browser (vuid auto inserted)?

packages/optimizely-sdk/lib/plugins/odp/event_manager/index.browser.ts
mikechu-optimizely: I was following an existing pattern (which I liked).  

I agree; this particular method is quite unnecessary unless later on we decide to change the logic and require 1 identifier or maybe identifier in browser requirements change.   🤷

PR #834, opened 2023-07-07
[FSSDK-9416] Refactor dependent promises typing in client onReady

raju-opti:
## Summary
- eventProcessorStartedPromise actually is typed as void in the Managed interface, but has a Promise<void> implementation in the extended EventProcessor interface. This PR fixes that.

## Test plan

## Issues
- [FSSDK-9416](https://jira.sso.episerver.net/browse/FSSDK-9416)


jaeopt (approved):
LGTM

PR #836, opened 2023-07-21
[FSSDK-8983] Specify package entry point using exports field 

raju-opti:
## Summary
- specified package entry points  using the modern [exports field in package.json](https://nodejs.org/api/packages.html#package-entry-points)
- added an export for lite bundle. lite bundle can be now imported using `import optimizely from '@optimizely/optimizely-sdk/lite`

## Test plan
- Tested that the new export entry points are working as expected. Added specific logs to the built files in the dist folder for each specific variant. Then tested that the specific variant is being loaded by installing that locally built sdk in a test project and inspecting the console log. Tested for both browser and node. 

## Issues
- FSSDK-8983

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,27 @@
   "browser": "dist/optimizely.browser.min.js",
   "react-native": "dist/optimizely.react_native.min.js",
   "typings": "dist/index.browser.d.ts",
+  "exports": {
+    ".": {
+      "node": {
+        "types": "./dist/index.node.d.ts",
+        "default": "./dist/optimizely.node.min.js"
+      },
+      "browser": {
+        "types": "./dist/index.browser.d.ts",
+        "default": "./dist/optimizely.browser.min.js"
+      },
+      "react-native": {
+        "types": "./dist/index.react_native.d.ts",
+        "default": "./dist/optimizely.react_native.min.js"
+      },
+      "default": "./dist/optimizely.node.min.js"
+    },
+    "./lite":  {
+      "types": "./dist/index.lite.d.ts",
+      "default": "./dist/optimizely.lite.es.js"
+    }
+  },
   "scripts": {
     "clean": "rm -rf dist",
     "clean:win": "(if exist dist rd /s/q dist)",
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
       },
       "browser": {
         "types": "./dist/index.browser.d.ts",
+        "import": "./dist/optimizely.browser.es.min.js",
         "default": "./dist/optimizely.browser.min.js"
       },
       "react-native": {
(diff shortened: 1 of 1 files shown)


packages/optimizely-sdk/package.json
mikechu-optimizely, line 1: Please update the Description > Tests section to include how you tested the imports (like with "lite") and the results of those test (did the prod bundle include just the expected dist files)

packages/optimizely-sdk/package.json
```diff
@@ -7,6 +7,55 @@
...
+      "react-native": {
+        "types": "./dist/index.react_native.d.ts",
+        "default": "./dist/optimizely.react_native.min.js"
+      },
+      "default": "./dist/optimizely.node.min.js"
+    },
+    "./lite":  {
+      "types": "./dist/index.lite.d.ts",
+      "node": "./dist/optimizely.lite.min.js",     
+      "import": "./dist/optimizely.lite.es.min.js",
```
mikechu-optimizely, line 30: I love that we can `import` specific part of the dist.  That really should help bundlers with shaking out unneeded code paths & files.

mikechu-optimizely (changes requested):
Just a couple of items/thoughts.

mikechu-optimizely:
@zashraf1985 Can you do a light review when you get a chance?

packages/optimizely-sdk/package.json
```diff
@@ -7,6 +7,55 @@
...
+      "default": "./dist/optimizely.lite.min.js"
+    },
+    "./dist/optimizely.lite.es": {
+      "types": "./dist/index.lite.d.ts",
+      "default": "./dist/optimizely.lite.es.js"
+    },
+    "./dist/optimizely.lite.es.js": {
+      "types": "./dist/index.lite.d.ts",
+      "default": "./dist/optimizely.lite.es.js"
+    },
```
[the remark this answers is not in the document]
raju-opti: Probably. We should at some point clean up the public api of the package. Currently kept these cause someone might be importing them.

packages/optimizely-sdk/package.json
raju-opti: updated

zashraf1985 (approved):
Generally looks great! But i have no idea about this new feature so i would recommend thorough testing using all possible known variants of user applications / supported JS versions before moving forward.

mikechu-optimizely (approved):
If Zeeshan says ok 😁

Be sure all checks are passing

mikechu-optimizely:
@raju-opti is this a good time to bring this PR into the code before AAT release?

raju-opti:
> @raju-opti is this a good time to bring this PR into the code before AAT release?

Yeah, I think it is!

PR #850, opened 2023-08-16
[FSSDK-8219] consolidate tsconfig.json

raju-opti:
## Summary
- previously there was a tsconfig.json in the repository root which was being extended by packages/optimizely-sdk/tsconfig.json. This PR removes the root config and moves all settings to packages/optimizely-sdk/tsconfig.json

## Test plan

## Issues
- FSSDK-8219


tsconfig.json
```diff
@@ -1,63 +0,0 @@
...
-    // "sourceRoot": "",                      /* Specify the location where debugger should locate TypeScript files instead of source locations. */
-    // "mapRoot": "",                         /* Specify the location where debugger should locate map files instead of generated locations. */
-    // "inlineSourceMap": true,               /* Emit a single file with source maps instead of having a separate file. */
-    // "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file; requires '--inlineSourceMap' or '--sourceMap' to be set. */
-
-    /* Experimental Options */
-    // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
-    // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
-  },
-  "exclude": ["node_modules", "src/**/*.spec.ts"]
```
mikechu-optimizely, line 62: Do we need to add `**/*.spec.ts` to the `"exclude"` in packages/optimizely-sdk/tsconfig.json

mikechu-optimizely (approved):
LGTM. 1 question to review but not show stopper I'd think

tsconfig.json
raju-opti: https://www.typescriptlang.org/tsconfig#exclude

most likely no. excludes are files that would be excluded when resolving includes.  it seems to me that all .spec.ts files are in optimizely-sdk/tests directory which is not included in the first place.

PR #852, opened 2023-08-17
[FSSDK-9562] set pixel api endpoint dynamically

raju-opti:
## Summary
- Previously, pixel api endpoint was fixed. Now pixel api endpoint is generated dynamically from odp apiHost. This is needed cause pixel api endpoints are different for different odp regions.

## Test plan

## Issues
- FSSDK-9562

### DIFF
--- a/lib/plugins/odp/event_api_manager/index.browser.ts
+++ b/lib/plugins/odp/event_api_manager/index.browser.ts
@@ -1,10 +1,11 @@
 import { OdpEvent } from '../../../core/odp/odp_event';
 import { OdpEventApiManager } from '../../../core/odp/odp_event_api_manager';
 import { LogLevel } from '../../../modules/logging';
-import { ODP_EVENT_BROWSER_ENDPOINT } from '../../../utils/enums';
 import { ODP_CONFIG_NOT_READY_MESSAGE } from '../../../core/odp/odp_event_api_manager';
 
 const EVENT_SENDING_FAILURE_MESSAGE = 'ODP event send failed';
+
+const pixelApiPath = 'v2/zaius.gif';
 
 export class BrowserOdpEventApiManager extends OdpEventApiManager {
   protected shouldSendEvents(events: OdpEvent[]): boolean {
@@ -15,6 +16,15 @@ export class BrowserOdpEventApiManager extends OdpEventApiManager {
     return false;
   }
 
+  private getPixelApiEndpoint(): string {
+    if (!this.odpConfig?.isReady()) {
+      throw new Error(ODP_CONFIG_NOT_READY_MESSAGE);
+    }
+    const apiHost = this.odpConfig.apiHost;
+    const pixelApiEndpoint = new URL(pixelApiPath, apiHost.replace('api', 'jumbe')).href;
+    return pixelApiEndpoint;
+  }
+
   protected generateRequestData(
     events: OdpEvent[]
   ): { method: string; endpoint: string; headers: { [key: string]: string }; data: string } {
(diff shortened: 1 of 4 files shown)


lib/plugins/odp/event_api_manager/index.browser.ts
```diff
@@ -1,11 +1,12 @@
 import { OdpEvent } from '../../../core/odp/odp_event';
 import { OdpEventApiManager } from '../../../core/odp/odp_event_api_manager';
 import { LogLevel } from '../../../modules/logging';
-import { ODP_EVENT_BROWSER_ENDPOINT } from '../../../utils/enums';
 import { ODP_CONFIG_NOT_READY_MESSAGE } from '../../../core/odp/odp_event_api_manager';
 
 const EVENT_SENDING_FAILURE_MESSAGE = 'ODP event send failed';
 
+const pixelApiPath = 'v2/zaius.gif';
```
mikechu-optimizely, line 8: Is it worth putting these in './utils/enums'?

lib/plugins/odp/event_api_manager/index.browser.ts
```diff
@@ -15,6 +16,15 @@ export class BrowserOdpEventApiManager extends OdpEventApiManager {
     return false;
   }
 
+  private getPixelApiEndpoint(): string {
+    if (!this.odpConfig?.isReady()) {
+      throw new Error(ODP_CONFIG_NOT_READY_MESSAGE);
+    }
+    const apiHost = this.odpConfig.apiHost;
+    const pixelApiEndpoint = new URL(pixelApiPath, apiHost.replace('api', 'jumbe')).href;
```
mikechu-optimizely, line 24: Yeah....just a thought. I wonder if it'd be helpful for constants in './utils/enums' or elsewhere. 🤔

lib/plugins/odp/event_api_manager/index.browser.ts
```diff
@@ -23,11 +33,17 @@ export class BrowserOdpEventApiManager extends OdpEventApiManager {
...
+    // and can be updated in place and it is done so in odp
+    // manager. We should make OdpConfig immutable and
+    // refacator later
+    const pixelApiEndpoint = this.getPixelApiEndpoint();
 
     const apiKey = this.odpConfig.apiKey;
     const method = 'GET';
     const event = events[0];
-    const url = new URL(ODP_EVENT_BROWSER_ENDPOINT);
+    const url = new URL(pixelApiEndpoint!);
```
mikechu-optimizely, line 46: Is there an alternative to `!`?

lib/plugins/odp/event_api_manager/index.browser.ts
mikechu-optimizely, line 1: Do we need to do the same edits for index.node.ts and its tests?

mikechu-optimizely: ...or maybe it's just a problem with the pixel.

mikechu-optimizely (changes requested):
Some thoughts/questions 👇

lib/plugins/odp/event_api_manager/index.browser.ts
raju-opti: I personally don't think there should be one giant file for all constants, rather each module should own it's own constants especially if it is part of the internal detail of the module. Here, no other module should be concerned with browser event api path, that is the responsibility of the BrowserOdpEventApiManager.

lib/plugins/odp/event_api_manager/index.browser.ts
raju-opti: My opinion is similar to previous comment. Maybe we can declare the constants in the same module, but finding proper names seemed a bit hard

lib/plugins/odp/event_api_manager/index.browser.ts
raju-opti: Yep. This is totally unnecessary here. Previously the code was such that it was needed. Later I changed to code and missed to remove this. Removing

lib/plugins/odp/event_api_manager/index.browser.ts
raju-opti: Yeah, just the pixel api

mikechu-optimizely (approved):
LGTM

PR #845, opened 2023-08-14
[FSSDK-8219] removed unnecessary subpackage dependency 

raju-opti:
## Summary
- remove js-sdk-datafile-manager subpackage from package.json

## Test plan
- existing tests should pass

## Issues
- FSSDK-8219

PR #841, opened 2023-08-01
[FSSDK-9506] Allow any polling interval but implement warning log 

mikechu-optimizely:
## Summary
- Remove 1 second polling floor to allow datafile polling at any frequency
- For intervals under 30 seconds, log a warning
- Add JSDoc around updateInterval and DEFAULT constants for clarity
- Added dev container and supporting configs
- Light formatting updates
- Updates to package-lock

## Test plan

- Added a new unit test file httpPollingDatafileManagerPolling to test polling & warning.
- Existing unit & e2e tests should pass

## Issues
- [FSSDK-9506](https://jira.sso.episerver.net/browse/FSSDK-9506)

PR #853, opened 2023-08-18
[FSSDK-9562] prepare for release 5.0.0-beta3

raju-opti:
## Summary
- prepare for release 5.0.0-beta3

## Test plan
-
## Issues
- FSSDK-9562

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [5.0.0-beta3] - August 18, 2023
 
 ### Bug fixes
-- Fixed odp event sending not working for Europe and Asia region ([#852](https://github.com/optimizely/javascript-sdk/pull/852))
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
```
mikechu-optimizely, line 9: Do we want to mention the re-organization of the repo in the event beta testers find something we missed?

mikechu-optimizely: PRs #845 to #851

mikechu-optimizely: Maybe also #844 and #841  ?

mikechu-optimizely (approved):
LGTM though I added some comments/questions.

CHANGELOG.md
raju-opti: "Do we want to mention the re-organization of the repo in the event beta testers find something we missed?" - I think no. This changelog is for the optimizely-sdk package which just moved to root now. This is not concerned with other packages in the now deprecated monorepo.

Apart from PR 841, others are internal reorganization, I would prefer to skip them. I'll add 841 as it has changes in the behavior of the public api.

PR #858, opened 2023-08-25
[FSSDK-9588] export logging types and values

raju-opti:
## Summary
- Clients like the testapp and react-sdk needs logging types and values, currently they have to import it from deep paths from the lib directory. this PR will allow imports directly for the entrypoint. We will eventually remove lib directory from the published  package bundle.

## Test plan

## Issues
- FSSDK-9588

### DIFF
--- a/lib/common_exports.ts
+++ b/lib/common_exports.ts
@@ -1,0 +1,19 @@
+/**
+ * Copyright 2023 Optimizely
+ *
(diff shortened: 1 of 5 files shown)


mikechu-optimizely (approved):
Great idea. Thanks.  React will be able to log through to JS. LGTM.

PR #861, opened 2023-08-30
[FSSDK-8687] update uuid and fix import paths

raju-opti:
## Summary
- update uuid and fix a few relative import paths

## Test plan

## Issues
- FSSDK-8687

### DIFF
--- a/packages/optimizely-sdk/package.json
+++ b/packages/optimizely-sdk/package.json
@@ -45,7 +45,7 @@
     "@optimizely/js-sdk-logging": "^0.3.1",
     "json-schema": "^0.4.0",
     "murmurhash": "0.0.2",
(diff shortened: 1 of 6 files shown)


mikechu-optimizely (approved):
Woo. Big jump, but LGTM assuming tests pass.

packages/optimizely-sdk/package.json
mikechu-optimizely, line 1: @raju-opti Is it time to remove this extra `uuid` package and use the native `crypto.randomUUID()` functions from [Node](https://nodejs.org/docs/latest-v14.x/api/crypto.html#crypto_crypto_randomuuid_options) and [Web](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) JS?

PR #864, opened 2023-09-14
[FSSDK-9654] Refactorings of ODP config & managers

mikechu-optimizely:
## Summary
- Refactor ODPConfig, segment manager and segment API manager for consistency with ODP Event Managers
- Add JS Docs & TS accessors

## Test plan

## Issues
- FSSDK-9654

### DIFF
--- a/lib/core/odp/odp_config.ts
+++ b/lib/core/odp/odp_config.ts
@@ -65,6 +65,8 @@ export class OdpConfig {
 
   /**
    * Determines if ODP configuration has the minimum amount of information
+   * @public
+   * @returns boolean indicating if ODP configuration is valid
    */
   public isValid(): boolean {
     return !!this._apiKey && !!this._apiHost;
--- a/lib/core/odp/odp_config.ts
+++ b/lib/core/odp/odp_config.ts
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 
-import { checkArrayEquality } from '../../utils/fns';
-
 export class OdpConfig {
   /**
    * Host of ODP audience segments API.
@@ -27,7 +25,7 @@ export class OdpConfig {
    * Getter to retrieve the ODP server host
    * @public
    */
-  get apiHost(): string {
+  public get apiHost(): string {
     return this._apiHost;
   }
 
@@ -41,7 +39,7 @@ export class OdpConfig {
    * Getter to retrieve the ODP API key
    * @public
    */
-  get apiKey(): string {
+  public get apiKey(): string {
     return this._apiKey;
   }
 
@@ -55,7 +53,7 @@ export class OdpConfig {
    * Getter for ODP segments to check
    * @public
    */
-  get segmentsToCheck(): string[] {
+  public get segmentsToCheck(): string[] {
     return this._segmentsToCheck;
   }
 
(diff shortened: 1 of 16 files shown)


lib/core/odp/odp_config.ts
```diff
@@ -65,40 +63,10 @@ export class OdpConfig {
...
+  public isValid(): boolean {
     return !!this._apiKey && !!this._apiHost;
   }
-
-  /**
-   * Detects if there are any changes between the current and incoming ODP Configs
-   * @param configToCompare ODP Configuration to check self against for equality
-   * @returns Boolean based on if the current ODP Config is equivalent to the incoming ODP Config
-   */
-  equals(configToCompare: OdpConfig): boolean {
```
raju-opti, line 97: this might be useful to keep

lib/core/odp/odp_manager.ts
```diff
@@ -61,43 +63,72 @@ export interface IOdpManager {
...
+   */
+  private _odpConfig: OdpConfig;
+
+  /**
+   * Getter for ODP configuration settings
+   * @public
+   * @returns ODP configuration settings in use
    */
-  odpConfig: OdpConfig = new OdpConfig(); // TODO: Consider making private and adding public accessors
+  public get odpConfig(): OdpConfig {
```
raju-opti, line 107: we probably don't need to get the config out of odp_manager. it can stay private

lib/core/odp/odp_manager.ts
```diff
@@ -61,43 +63,72 @@ export interface IOdpManager {
...
    * @protected
    */
-  protected logger: LogHandler = getLogger(); // TODO: Consider making private and moving instantiation to constructor
+  protected logger: LogHandler;
 
   /**
    * ODP configuration settings for identifying the target API and segments
+   * @private
+   */
+  private _odpConfig: OdpConfig;
```
raju-opti, line 100: this should be optional

lib/core/odp/odp_manager.ts
```diff
@@ -61,43 +63,72 @@ export interface IOdpManager {
...
+   * @returns ODP configuration settings in use
    */
-  odpConfig: OdpConfig = new OdpConfig(); // TODO: Consider making private and adding public accessors
+  public get odpConfig(): OdpConfig {
+    return this._odpConfig;
+  }
+
+  constructor(config?: OdpConfig, logger?: LogHandler) {
+    this.logger = logger ?? getLogger();
+    this._odpConfig = config ?? new OdpConfig();
```
raju-opti, line 113: no empty config should be set. odp_manager should ensure that if the config field is set, then the config is valid

lib/core/odp/odp_manager.ts
```diff
@@ -114,22 +145,18 @@ export abstract class OdpManager implements IOdpManager {
 
     this.eventManager.flush();
 
-    const newConfig = new OdpConfig(apiKey, apiHost, segmentsToCheck);
-    const configDidUpdate = this.odpConfig.update(newConfig);
+    this._odpConfig = newConfig;
```
raju-opti, line 148: before setting and propagating the updateSettings() down to the child components, we should check whether this new config is equal to the previous config, in which case we should just return without doing anything. 
```
if (newConfig.equals(this._odpConfig) {
  return
 }
```

lib/core/odp/odp_manager.ts
```diff
@@ -114,22 +145,18 @@ export abstract class OdpManager implements IOdpManager {
...
 
-    if (configDidUpdate) {
-      this.odpConfig.update(newConfig);
-      this.segmentManager?.reset();
```
raju-opti, line 122: segmentManger reset should probably stay

lib/core/odp/odp_manager.ts
```diff
@@ -114,22 +145,18 @@ export abstract class OdpManager implements IOdpManager {
...
-      return true;
-    }
+    this.segmentManager?.updateSettings(newConfig);
```
raju-opti, line 150: should probably also add `eventManager.updateSettings()` ?

raju-opti (commented):
Added a few comments

lib/core/odp/odp_config.ts
mikechu-optimizely: This was a bit I added early on and was only used in one place for low value added.  I was thinking now that it's not worth the bundle size. YAGNI.

lib/core/odp/odp_manager.ts
mikechu-optimizely: Done.

lib/core/odp/odp_manager.ts
mikechu-optimizely: Done.

lib/core/odp/odp_manager.ts
mikechu-optimizely: Done.

lib/core/odp/odp_manager.ts
mikechu-optimizely: Ohhh Ok. I'll add back the `.equals()`

lib/core/odp/odp_manager.ts
mikechu-optimizely: segmentManager.updateSettings() does a reset of its own so I figured let that responsibility be there.

lib/core/odp/odp_manager.ts
mikechu-optimizely: Done.

mikechu-optimizely:
Close temporarily to save CI runs

PR #866, opened 2023-09-26
[FSSDK-9605] add sendBeacon event dispatcher support

raju-opti:
## Summary
-  add support for event dispatcher using sendBeacon which will be used in browser on optimizely client close.

## Test plan
- added new tests for the new dispatcher and event processor
- manually tested that event are being sent using sendBeacon on close

## Issues
- FSSDK-9605

### DIFF
--- a/lib/index.browser.ts
+++ b/lib/index.browser.ts
@@ -91,11 +91,7 @@ const createInstance = function(config: Config): Client | null {
       eventDispatcher = config.eventDispatcher;
     }
 
-    let closingDispatcher = config.closingEventDispatcher;
-
-    if (!closingDispatcher && window.navigator && 'sendBeacon' in window.navigator) {
-      closingDispatcher = sendBeaconEventDispatcher;
-    }
+    const closingDispatcher = config.closingEventDispatcher;
 
     let eventBatchSize = config.eventBatchSize;
     let eventFlushInterval = config.eventFlushInterval;
--- a/lib/index.browser.ts
+++ b/lib/index.browser.ts
@@ -15,10 +15,11 @@
  */
 import logHelper from './modules/logging/logger';
 import { getLogger, setErrorHandler, getErrorHandler, LogLevel } from './modules/logging';
-import { LocalStoragePendingEventsDispatcher } from './modules/event_processor';
+import { EventDispatcher, LocalStoragePendingEventsDispatcher } from './modules/event_processor';
 import configValidator from './utils/config_validator';
 import defaultErrorHandler from './plugins/error_handler';
 import defaultEventDispatcher from './plugins/event_dispatcher/index.browser';
+import sendBeaconEventDispatcher from './plugins/event_dispatcher/send_beacon_dispatcher';
 import * as enums from './utils/enums';
 import * as loggerPlugin from './plugins/logger';
 import eventProcessorConfigValidator from './utils/event_processor_config_validator';
@@ -90,6 +91,9 @@ const createInstance = function(config: Config): Client | null {
       eventDispatcher = config.eventDispatcher;
     }
 
+    const closingDispatcher = (window.navigator && 'sendBeacon' in window.navigator) ?
+      sendBeaconEventDispatcher : undefined;
+
     let eventBatchSize = config.eventBatchSize;
     let eventFlushInterval = config.eventFlushInterval;
 
@@ -111,6 +115,7 @@ const createInstance = function(config: Config): Client | null {
 
     const eventProcessorConfig = {
       dispatcher: eventDispatcher,
+      closingDispatcher,
       flushInterval: eventFlushInterval,
       batchSize: eventBatchSize,
       maxQueueSize: config.eventMaxQueueSize || DEFAULT_EVENT_MAX_QUEUE_SIZE,
@@ -172,6 +177,7 @@ export {
   loggerPlugin as logging,
   defaultErrorHandler as errorHandler,
   defaultEventDispatcher as eventDispatcher,
+  sendBeaconEventDispatcher,
   enums,
   setLogHandler as setLogger,
   setLogLevel,
@@ -189,6 +195,7 @@ export default {
   logging: loggerPlugin,
   errorHandler: defaultErrorHandler,
   eventDispatcher: defaultEventDispatcher,
+  sendBeaconEventDispatcher,
   enums,
   setLogger: setLogHandler,
   setLogLevel,
--- a/lib/index.browser.ts
+++ b/lib/index.browser.ts
@@ -91,7 +91,11 @@ const createInstance = function(config: Config): Client | null {
       eventDispatcher = config.eventDispatcher;
     }
 
-    const closingDispatcher = config.closingEventDispatcher;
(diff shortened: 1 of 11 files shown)


lib/index.browser.ts
```diff
@@ -111,6 +115,7 @@ const createInstance = function(config: Config): Client | null {
 
     const eventProcessorConfig = {
       dispatcher: eventDispatcher,
+      closingDispatcher,
```
jaeopt, line 118: Wondering if we see any possible side-effect and need to make this configurable.

lib/modules/event_processor/eventProcessor.ts
```diff
@@ -57,13 +57,20 @@ export function validateAndGetBatchSize(batchSize: number): number {
...
 }
 
-export function getQueue(batchSize: number, flushInterval: number, sink: EventQueueSink<ProcessableEvent>, batchComparator: (eventA: ProcessableEvent, eventB: ProcessableEvent) => boolean ): EventQueue<ProcessableEvent> {
+export function getQueue(
+  batchSize: number, 
+  flushInterval: number, 
+  batchComparator: (eventA: ProcessableEvent, eventB: ProcessableEvent) => boolean,
+  sink: EventQueueSink<ProcessableEvent>,
+  closingSink?: EventQueueSink<ProcessableEvent>
+): EventQueue<ProcessableEvent> {
```
jaeopt, line 66: Don't we need the same beacon support for ODPEventDispatcher as well for consistency?

jaeopt (changes requested):
Looks good! A couple of clarifications.

lib/modules/event_processor/eventProcessor.ts
raju-opti: Currently, ODPEventManager for browser does not queue any events, it sends them immediately, so we don't have a similar use case.

lib/index.browser.ts
raju-opti: Yeah, probably should make it configurable as well.

lib/modules/event_processor/eventProcessor.ts
jaeopt: what is expected if we have back-2-back ODP events? 2nd one is discarded or blocking?

raju-opti: @jaeopt in that case both event will be sent immediately asynchronously, without blocking or waiting.

alexparish:
This is really great. 

In order to synchronise activations to a data platform I have created a custom event dispatcher that I pass to `createInstance`. Ultimately this custom event dispatcher calls `eventDispatcher.dispatchEvent(eventObject, callback)` with `eventDispatcher` imported from the SDK package.

Will this pull request support my use case or will I have to integrate `sendBeaconDispatcher` into my custom event dispatcher?

raju-opti:
> This is great, thanks for implementing the Beacon API.
> 
> In order to synchronise activations to a data platform I am using a custom event dispatcher that I pass to `createInstance`. Ultimately this custom event dispatcher calls `eventDispatcher.dispatchEvent(eventObject, callback)` with `eventDispatcher` imported from the SDK package.
> 
> Will people like myself using custom event dispatchers need to manually integrate`sendBeaconDispatcher`?

@alexparish yes, you will need to wrap the `sendBeaconEventDispatcher` similarly and pass it as `closingEventDispatcher` option to `createInstance`.

mikechu-optimizely:
@raju-opti and I are finishing our manual e2e acceptance testing (bug bashing) after the code review....

mikechu-optimizely (commented):
Some questions

lib/modules/event_processor/eventQueue.ts
```diff
@@ -14,149 +14,149 @@
...
     batchComparator,
   }: {
-    flushInterval: number
-    maxQueueSize: number
-    sink: EventQueueSink<K>
-    batchComparator: (eventA: K, eventB: K) => boolean
+    flushInterval: number;
+    maxQueueSize: number;
+    sink: EventQueueSink<K>;
+    closingSink?: EventQueueSink<K>;
```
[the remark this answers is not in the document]
raju-opti: This is just a field in an object, so order should not matter.

lib/modules/event_processor/v1/v1EventProcessor.ts
```diff
@@ -36,31 +36,41 @@ const logger = getLogger('LogTierV1EventProcessor')
...
 
   constructor({
     dispatcher,
+    closingDispatcher,
     flushInterval = DEFAULT_FLUSH_INTERVAL,
     batchSize = DEFAULT_BATCH_SIZE,
     notificationCenter,
   }: {
     dispatcher: EventDispatcher
+    closingDispatcher?: EventDispatcher
```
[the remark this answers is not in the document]
raju-opti: Order of optional arguments to a function matters in TS, they must come after all mandatory params. However, here closingDispatcher is a field of an object, not a parameter of the constructor itself. So order should not matter. And as it is an optional field, the change to API is non-breaking.

lib/plugins/event_dispatcher/send_beacon_dispatcher.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Users of the SDK can provide their own implementation
+ * @param  {Event}    eventObj
+ * @param  {Function} callback
+ */
+export const dispatchEvent = function(
+  eventObj: Event,
+  callback: (response: { statusCode: number; }) => void
+): void {
+  const params = eventObj.params;
+  const url: string = eventObj.url;
```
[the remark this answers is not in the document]
raju-opti: updated

mikechu-optimizely (approved):
LGTM. 👍   Thanks.

mikechu-optimizely:
@jaeopt, Are you good with this PR?

jaeopt (changes requested):
Changes look good to me! 
With this change, sendBeacon is not default any more, so clients should instantiate the custom handler to use it. Other option is to use it as the default and they can turn it off if they do not like it. WDYT?

raju-opti:
> Changes look good to me! With this change, sendBeacon is not default any more, so clients should instantiate the custom handler to use it. Other option is to use it as the default and they can turn it off if they do not like it. WDYT?

@jaeopt As this PR introduces option to use a separate dispatcher when closing the instance, this would break users using custom event dispatcher if used by default when closing, that's why we have to keep it optional.

alexparish:
Is there a reason why events cannot always be sent using `sendBeacon` in the browser? It would simplify the change, for example `closingEventDispatcher` would not need to be specified when using `createInstance`,.

raju-opti:
> Is there a reason why events cannot always be sent using `sendBeacon` in the browser? It would simplify the change, for example `closingEventDispatcher` would not need to be specified when calling `createInstance`.

Yes, sendBeacon does not provide any way of checking the HTTP response status. By default, we will want to check the response status and retry in case of 5xx status.

mikechu-optimizely:
Last commit, still LGTM.

jaeopt (approved):
LGTM

mikechu-optimizely:
Woot 🎉

PR #882, opened 2023-12-05
FSSDK-9843: Add infinity check

mandytrinh:
A customer in [BUG-7104](https://jira.sso.episerver.net/browse/BUG-7104) was sending 1/0 values in the tags for their events. Because the NodeJS SDK [doesn't have a check](https://github.com/optimizely/javascript-sdk/blob/master/lib/utils/event_tag_utils/index.ts#L67) against infinity, this allowed the value to be passed through subsequently broke the results API calculations. 

Other SDKs ([python](https://github.com/optimizely/python-sdk/blob/master/optimizely/helpers/event_tag_utils.py#L106), [java](https://github.com/optimizely/java-sdk/blob/master/core-api/src/main/java/com/optimizely/ab/internal/EventTagUtils.java#L64)) have this check, so we should add it to Node.

`isNaN()` can be simplified to just `!isFinite()` which checks for +/- infinity as well as NaN

### DIFF
--- a/lib/utils/event_tag_utils/index.ts
+++ b/lib/utils/event_tag_utils/index.ts
@@ -41,7 +41,7 @@ export function getRevenueValue(eventTags: EventTags, logger: LoggerFacade): num
     let parsedRevenueValue;
     if (typeof rawValue === 'string') {
       parsedRevenueValue = parseInt(rawValue);
-      if (isNaN(parsedRevenueValue)) {
+      if (!isFinite(parsedRevenueValue)) {
         logger.log(LOG_LEVEL.INFO, LOG_MESSAGES.FAILED_TO_PARSE_REVENUE, MODULE_NAME, rawValue);
         return null;
       }
@@ -70,7 +70,7 @@ export function getEventValue(eventTags: EventTags, logger: LoggerFacade): numbe
     let parsedEventValue;
     if (typeof rawValue === 'string') {
       parsedEventValue = parseFloat(rawValue);
-      if (isNaN(parsedEventValue)) {
+      if (!isFinite(parsedEventValue)) {
         logger.log(LOG_LEVEL.INFO, LOG_MESSAGES.FAILED_TO_PARSE_VALUE, MODULE_NAME, rawValue);
         return null;
       }


lib/utils/event_tag_utils/index.ts
```diff
@@ -41,7 +41,7 @@ export function getRevenueValue(eventTags: EventTags, logger: LoggerFacade): num
     let parsedRevenueValue;
     if (typeof rawValue === 'string') {
       parsedRevenueValue = parseInt(rawValue);
-      if (isNaN(parsedRevenueValue)) {
+      if (!isFinite(parsedRevenueValue)) {
```
mikechu-optimizely, line 44: Should we use `Number.isFinite` ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite))?  🤔  I mean, maybe not since we're doing `parseInt()` and `parseFloat()` later.

raju-opti: the previous line does a `parseInt`. So if `rawValue` contains some string like "1/0", `parsedRevenueValue` will just contain 1. Replacing `isNaN` with `!isFinite` is nice, but the error the customer faced is probably due to the `number` type of `rawValue` where the client passed 1/0 as a number in which case it will be Infinite. So we need to add this `if (!isFinite(parsedRevenueValue)) `  check there too (After [line 52](https://github.com/optimizely/javascript-sdk/blob/fc1efa088298569f1001eb07a0eae7cb09e83314/lib/utils/event_tag_utils/index.ts#L52C37-L52C37))

lib/utils/event_tag_utils/index.ts
```diff
@@ -70,7 +70,7 @@ export function getEventValue(eventTags: EventTags, logger: LoggerFacade): numbe
     let parsedEventValue;
     if (typeof rawValue === 'string') {
       parsedEventValue = parseFloat(rawValue);
-      if (isNaN(parsedEventValue)) {
+      if (!isFinite(parsedEventValue)) {
```
raju-opti, line 73: similarly, we need to add this check after [line 81](https://github.com/optimizely/javascript-sdk/blob/fc1efa088298569f1001eb07a0eae7cb09e83314/lib/utils/event_tag_utils/index.ts#L81)

raju-opti (changes requested):

lib/utils/event_tag_utils/index.ts
mandytrinh: thanks @raju-opti, did not know that javascript evaluates `typeof(1/0)` to be a 'number'. Would think it would be 'undefined'. 
I added the check and refactored the entire function to be more readable and concise

raju-opti (approved):
LGTM

PR #888, opened 2024-01-12
[FSSDK-9607] add pixelUrl to OdpConfig

andrewleap-optimizely:
## Summary
- Changed the ODP pixelUrl to be read from the datafile instead of being hard coded in the SDK.

## Ticket
[FSSDK-9607](https://jira.sso.episerver.net/browse/FSSDK-9607)

### DIFF
--- a/lib/core/project_config/project_config_schema.ts
+++ b/lib/core/project_config/project_config_schema.ts
@@ -290,6 +290,9 @@ var schemaDefinition = {
           },
           publicKey: {
             type: 'string'
+          },
+          pixelUrl: {
+            type: 'string'
           }
         }
       }
--- a/lib/core/project_config/project_config_schema.ts
+++ b/lib/core/project_config/project_config_schema.ts
@@ -293,10 +293,10 @@ var schemaDefinition = {
           },
           pixelUrl: {
             type: 'string'
-          }
-        }
-      }
-    }
+          },
+        },
+      },
+    },
   },
 };
 
--- a/lib/index.browser.tests.js
+++ b/lib/index.browser.tests.js
@@ -1116,9 +1116,9 @@ describe('javascript-sdk (Browser)', function() {
         clock.tick(100);
 
         let publicKey = datafile.integrations[0].publicKey;
-        let pixelHost = datafile.integrations[0].pixelUrl;
+        let pixelUrl = datafile.integrations[0].pixelUrl;
 
-        const pixelApiEndpoint = `${pixelHost}/v2/zaius.gif`;
+        const pixelApiEndpoint = `${pixelUrl}/v2/zaius.gif`;
         let requestEndpoint = new URL(requestParams.get('endpoint'));
         assert.equal(requestEndpoint.origin + requestEndpoint.pathname, pixelApiEndpoint);
         assert.equal(requestParams.get('method'), 'GET');
--- a/lib/index.browser.tests.js
+++ b/lib/index.browser.tests.js
@@ -879,7 +879,7 @@ describe('javascript-sdk (Browser)', function() {
 
         client.sendOdpEvent('test', '', new Map([['eamil', 'test@test.test']]), new Map([['key', 'value']]));
         clock.tick(10000);
-        
+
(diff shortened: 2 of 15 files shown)


lib/index.browser.tests.js
```diff
@@ -1116,8 +1116,9 @@ describe('javascript-sdk (Browser)', function() {
         clock.tick(100);
 
         let publicKey = datafile.integrations[0].publicKey;
+        let pixelHost = datafile.integrations[0].pixelUrl;
 
-        const pixelApiEndpoint = 'https://jumbe.zaius.com/v2/zaius.gif';
+        const pixelApiEndpoint = `${pixelHost}/v2/zaius.gif`;
```
mikechu-optimizely, line 1121: Are we getting the full pixel URL (all the way to `zaius.gif`) from the datafile or just the host?  This test made me question.  

I mean I'm in favor of getting the FULL host + path + resource from datafile letting the server control this.

andrewleap-optimizely: Currently we're just getting the host- I'm not opposed to making that change though, can't think of any good reason not to...

lib/core/project_config/project_config_schema.ts
```diff
@@ -290,6 +290,9 @@ var schemaDefinition = {
           },
           publicKey: {
             type: 'string'
+          },
+          pixelUrl: {
+            type: 'string'
           }
```
mikechu-optimizely, line 296: NIT: Cool if we add a trailing `,` ? Maybe it's a linter prob 🤷

lib/index.browser.tests.js
mikechu-optimizely: If we're only getting & using the pixelHost, is it worth renaming vars to pixelHost elsewhere? Just 🤔

lib/plugins/odp/event_api_manager/index.browser.ts
```diff
@@ -20,8 +20,8 @@ export class BrowserOdpEventApiManager extends OdpEventApiManager {
     if (!this.odpConfig?.isReady()) {
       throw new Error(ODP_CONFIG_NOT_READY_MESSAGE);
     }
-    const apiHost = this.odpConfig.apiHost;
-    const pixelApiEndpoint = new URL(pixelApiPath, apiHost.replace('api', 'jumbe')).href;
+    const pixelUrl = this.odpConfig.pixelUrl;
```
mikechu-optimizely, line 23: NIT: Yeah maybe this should be `pixelHost`  🤷

mikechu-optimizely (approved):
Just some thoughts but nothing show-stopping.  LGTM.

lib/plugins/odp/event_api_manager/index.browser.ts
```diff
@@ -20,8 +20,8 @@ export class BrowserOdpEventApiManager extends OdpEventApiManager {
...
+    const pixelApiEndpoint = new URL(pixelApiPath, pixelUrl).href;
```
raju-opti, line 24: Is it currently possible to have a datafile that has odp apiHost but not the pixelUrl? (for example: some project whose datafile was generated a long time ago and not updated since? or someone who are are calling createInstance with a datafile generated before adding pixelUrl?). In that case, should we check if pixelUrl exists, and if not, fallback to `apiHost.replace('api', 'jumbe'))`?

raju-opti (changes requested):

lib/plugins/odp/event_api_manager/index.browser.ts
andrewleap-optimizely: You know I don't disagree, however it seems like the real fix is to make the change in the datafile.

Since we don't modify the value from the datafile, then it seems like the least confusing thing is to match the name from the datafile.

lib/plugins/odp/event_api_manager/index.browser.ts
andrewleap-optimizely: Good callout. This would have to be a beta customer- I just went through and checked all the beta datafiles and we're all good!

raju-opti (approved):
LGTM

PR #895, opened 2024-02-06
[FSSDK-9987] fix: Conditional ODP instantiation

mikechu-optimizely:
## Summary
- When explicit `odpOptions.disabled = true` is passed, avoid ODP instantiation
- Fix Jest test setting in VSCode
- Lint updates

## Test plan

- Existing unit and e2e test suite are expected to pass

## Issues
- FSSDK-9987


mikechu-optimizely:
Developer review completed.

jaeopt (changes requested):
Looks good to me. We need to double check of its side-effects.

For example - https://github.com/optimizely/javascript-sdk/blob/961f6a124f132bc49bee0b318acdca2f7f39ad2b/lib/optimizely/index.ts#L1761

They do not want to see many "ERROR"-level logging. Better stay silent except for "ODP disabled" at the init phase.

mikechu-optimizely:
> Looks good to me. We need to double check of its side-effects.
> 
> For example -
> 
> https://github.com/optimizely/javascript-sdk/blob/961f6a124f132bc49bee0b318acdca2f7f39ad2b/lib/optimizely/index.ts#L1761
> 
> They do not want to see many "ERROR"-level logging. Better stay silent except for "ODP disabled" at the init phase.

This is an EXCELLENT catch @jaeopt. Thanks.  Let me comb through and find these scenarios.

mikechu-optimizely:
Temp close

PR #900, opened 2024-02-08
[FSSDK-9966] use coveralls-next instead of coveralls

raju-opti:
## Summary
- use coveralls-next instead of coveralls as it depends on deprecated packages

## Test plan

## Issues
- FSSDK-9966


pulak-opti (approved):
LGTM!

PR #902, opened 2024-02-08
[FSSDK-9987] fix: Conditional ODP instantiation

mikechu-optimizely:
## Summary
- When explicit `odpOptions.disabled = true` is passed, avoid ODP instantiation
- Fix Jest test setting in VSCode
- Lint updates

## Test plan

- Existing unit and e2e test suite are expected to pass

## Issues
- FSSDK-9987

### DIFF
--- a/lib/index.browser.tests.js
+++ b/lib/index.browser.tests.js
@@ -1,11 +1,11 @@
 /**
- * Copyright 2016-2020, 2022-2023 Optimizely
+ * Copyright 2016-2020, 2022-2024 Optimizely
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


jaeopt (approved):
LGTM - this looks good. but I do not see cleaning up "odpManager.enabled" flag discussed. They'll be in a separate PR?

mikechu-optimizely:
> LGTM - this looks good. but I do not see cleaning up "odpManager.enabled" flag discussed. They'll be in a separate PR?

Correct. Need small PRd

mikechu-optimizely:
Actually @jaeopt  The more I think about what we paired up on:

We have 3 states:
1. Default: ODP is `enabled = true` but we find that the datafile does not have ODP integration so we go into a enabled = false state with minimal class instantiated and heap allocated in case ... 
1. Enabled: ODP stack is activated by the datafile and queuing and processing is active and running with needed memory space 
1. Explicitly disabled: Developer sets `odpOptions.disabled: true` and client never instantiates ODP stack (addressed in this PR)

DM me we can talk more but I think this PR should close the issue.

PR #905, opened 2024-02-23
[FSSDK-9570] add explicit entry point for node, browser and react_native

raju-opti:
## Summary
- add explicit entry point for node, browser and react_native

## Test plan
- tested manually with logging that the correct bundle is loaded

## Issues
- FSSDK-9570


mikechu-optimizely (approved):
LGTM

PR #906, opened 2024-02-28
[FSSDK-9999] add release workflow

raju-opti:
## Summary
- add release workflow

## Test plan
- tested all steps except the actual release and jellyfish reporting manually in another project

## Issues
- FSSDK-9999


.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+          elif [[ "$VERSION" == *"-alpha"* ]]; then
+            echo "npm-tag=alpha" >> "$GITHUB_OUTPUT"
+          elif [[ "$VERSION" == *"-rc"* ]]; then
+            echo "npm-tag=rc" >> "$GITHUB_OUTPUT"
+          else
+            echo "npm-tag=v$(echo $VERSION | awk -F. '{print $1}')" >> "$GITHUB_OUTPUT"
+          fi
+
+      - id: release
+        name: test, build and publish to npm
```
mikechu-optimizely, line 56: NIT: 😁 

```suggestion
        name: Test, build and publish to npm
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
+name: Publish SDK to NPM
+
+on:
+  release:
+    types: [published, edited]
```
mikechu-optimizely, line 5: I ran into some problems with `edited`. (based on my memory) If I modified anything about the release, including the text, it would re-run this workflow.  I think.

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
+name: Publish SDK to NPM
+
+on:
```
mikechu-optimizely, line 3: Did you want to add a `workflow_dispatch` that would always `--dry-run` for testing? Just a thought.

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ ! github.event.release.draft }}
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
```
mikechu-optimizely, line 19:
```suggestion
          node-version: 18
```

16 is already EOL and 18 is in maintenance...but 20 is another Jira ticket.

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+      - id: latest-release
+        name: Export latest release git tag
+        run: |
+          echo "latest-release-tag=$(curl -qsSL \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+            "${{ github.api_url }}/repos/${{ github.repository }}/releases/latest" \
+          | jq -r .tag_name)" >> $GITHUB_OUTPUT
+
+      - id: npm-tag
```
mikechu-optimizely, line 37: Love this. Very clever and more reliable.

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+          BROWSERSTACK_USERNAME: ${{ secrets.BROWSERSTACK_USERNAME }}
+          BROWSERSTACK_ACCESS_KEY: ${{ secrets.BROWSERSTACK_ACCESS_KEY }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
+        run: npm publish --tag=${{ steps.npm-tag.outputs['npm-tag'] }}
+
+      - name: Report results to Jellyfish
+        uses: optimizely/jellyfish-deployment-reporter-action@main
+        if: ${{ always() && (steps.release.outcome == 'success' || steps.release.outcome == 'failure') }}
+        with:
+          jellyfish_api_token: ${{ secrets.JELLYFISH_API_TOKEN }}
```
mikechu-optimizely, line 67: @zashraf1985 Was this put in the "Organization secrets"?  It could be that I'm just not seeing in the Setting for this repo.

mikechu-optimizely (approved):
LGTM. Nothing show-stopping but some thoughts 👇

.github/workflows/release.yml
zashraf1985: Yes! This is in the org secrets. Please try to run the workflow and see if it can access it. If not, then it can be the issue with the configuration of org secrets since its a public repo

.github/workflows/release.yml
raju-opti: Yeah, this re runs when anything is edited. But I think it is critical to republish if the tag is edited?

.github/workflows/release.yml
raju-opti: Some steps rely on the release event object, which won't be available for workflow_dispatch event I guess

.github/workflows/release.yml
raju-opti: Yeah, but according to product, we support one previous version to the current active version, which means we support 16. So I try to do everything with that version

.github/workflows/release.yml
raju-opti: updated

.github/workflows/release.yml
mikechu-optimizely: Oh I took that to mean that we support the current SDK version and the 1 SDK version prior.

.github/workflows/release.yml
```diff
@@ -0,0 +1,80 @@
...
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
+        run: |
+          if [[ ${{ github.event_name }} == "workflow_dispatch" ]]; then
+            DRY_RUN="--dry-run"
+          fi
+          npm publish --tag=${{ steps.npm-tag.outputs['npm-tag'] }} $DRY_RUN
+
+      - name: Report results to Jellyfish
+        uses: optimizely/jellyfish-deployment-reporter-action@main
+        if: ${{ always() && (steps.release.outcome == 'success' || steps.release.outcome == 'failure') }}
```
mikechu-optimizely, line 77: For dry-run via `workflow_dispatch`, should we update the `if` here?

.github/workflows/release.yml
mikechu-optimizely: Resolving. We can stick with 16 until we get moving towards 20.

mikechu-optimizely (approved):
LGTM.

.github/workflows/release.yml
raju-opti: added condition

PR #907, opened 2024-02-29
[FSSDK-9999] prepare release 5.1.0

raju-opti:
also remove separate version enums for node and browser as both are released as same version

## Issues
- FSSDK-9999


mikechu-optimizely (approved):
LGTM

PR #912, opened 2024-03-07
[FSSDK-9586] restore node engine to >=14

raju-opti:
also add back unit test ci for node14

v5.0.0 was released with engine.node >=14.0.0. So updating this technically constitutes a breaking change.

## Issues
- FSSDK-9586


mikechu-optimizely (approved):
LGTM. Thanks!

PR #920, opened 2024-04-02
[FSSDK-10090] Refactor ODP integration

raju-opti:
## Summary

Things fixed:
- ODPManager should not be running and scheduling timer if ODP is not integrated to the project (which causes memory leak if one sdk instance is created per request)
- CreateUserContext should work even when called before the datafile is downloaded and should send the `identify` ODP events after datafile download completes
- Other automatic odp events (vuid registration, client initialized) should also be sent after datafile is available and should not be dropped if batching is disabled.
- ODP manager should distinguish between when odp config is still not loaded and when odp is not integrated at all
- made ODP config  immutable
- ODP event manager should not schedule timers to flush events if event batching is disabled (ie batch size == 1)
- other code quality improvements

## Test plan
- added new tests where applicable, and updated old tests

## Issues
- FSSDK-10090

### DIFF
--- a/lib/core/odp/odp_config.ts
+++ b/lib/core/odp/odp_config.ts
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-import { config } from 'chai';
 import { checkArrayEquality } from '../../utils/fns';
 
 export class OdpConfig {
--- a/lib/core/odp/odp_config.ts
+++ b/lib/core/odp/odp_config.ts
@@ -14,166 +14,71 @@
  * limitations under the License.
  */
 
+import { config } from 'chai';
 import { checkArrayEquality } from '../../utils/fns';
 
-export type NoOdpIntegrationConfig = {
+export class OdpConfig {
+  /**
+   * Host of ODP audience segments API.
+   */
+  readonly apiHost: string;
+
+  /**
+   * Public API key for the ODP account from which the audience segments will be fetched (optional).
+   */
+  readonly apiKey: string;
+
+  /**
+   * Url for sending events via pixel.
+   */
+  readonly pixelUrl: string;
+
+  /**
+   * All ODP segments used in the current datafile (associated with apiHost/apiKey).
+   */
+  readonly segmentsToCheck: string[];
+
+  constructor(apiKey: string, apiHost: string, pixelUrl: string, segmentsToCheck: string[]) {
+    this.apiKey = apiKey;
+    this.apiHost = apiHost;
+    this.pixelUrl = pixelUrl;
+    this.segmentsToCheck = segmentsToCheck;
+  }
+
+  /**
+   * Detects if there are any changes between the current and incoming ODP Configs
+   * @param configToCompare ODP Configuration to check self against for equality
+   * @returns Boolean based on if the current ODP Config is equivalent to the incoming ODP Config
+   */
+  equals(configToCompare: OdpConfig): boolean {
+    return (
+      this.apiHost === configToCompare.apiHost &&
+      this.apiKey === configToCompare.apiKey &&
+      this.pixelUrl === configToCompare.pixelUrl &&
+      checkArrayEquality(this.segmentsToCheck, configToCompare.segmentsToCheck)
+    );
+  }
+}
+
+type OdpNotIntegratedConfig = {
   readonly integrated: false;
 }
 
-export type OdpIntegrationConfig = {
+type OdpIntegratedConfig = {
   readonly integrated: true;
-  readonly apiHost: string;
-  readonly apiKey: string;
-  readonly pixelUrl?: string;
-  readonly segmentsToCheck?: string[];
+  readonly odpConfig: OdpConfig;
 }
 
-export type OdpConfig = (NoOdpIntegrationConfig | OdpIntegrationConfig) & {
-  equals(odpConfig: OdpConfig): boolean;
-}
-
-function areOdpConfigsEqual(config1: OdpConfig, config2: OdpConfig): boolean {
+export const odpIntegrationEquals = (config1: OdpIntegrationConfig, config2: OdpIntegrationConfig): boolean => {
   if (config1.integrated !== config2.integrated) {
     return false;
   }
+
   if (config1.integrated && config2.integrated) {
-    return (
-      config1.apiHost === config2.apiHost &&
-      config1.apiKey === config2.apiKey &&
-      config1.pixelUrl === config2.pixelUrl &&
-      checkArrayEquality(config1.segmentsToCheck || [], config2.segmentsToCheck || [])
-    );
+    return config1.odpConfig.equals(config2.odpConfig);
   }
+
   return true;
 }
 
-export function createOdpIntegrationConfig(
-  apiHost: string,
-  apiKey: string,
-  pixelUrl?: string,
-  segmentsToCheck?: string[]
-): OdpConfig {
-  return {
-    integrated: true,
-    apiHost,
-    apiKey,
-    pixelUrl,
-    segmentsToCheck,
-    equals: function(odpConfig: OdpConfig) {
-      return areOdpConfigsEqual(this, odpConfig)
-    }
-  };
-}
-
-export function createNoOdpIntegrationConfig(): OdpConfig {
-  return {
-    integrated: false,
-    equals: function (odpConfig: OdpConfig) {
-      return areOdpConfigsEqual(this, odpConfig)
-    }
-  };
-}
-
-// export class OdpConfig {
-//   /**
-//    * Host of ODP audience segments API.
-//    * @private
-//    */
-//   private _apiHost: string;
-
-//   /**
-//    * Getter to retrieve the ODP server host
-//    * @public
-//    */
-//   get apiHost(): string {
-//     return this._apiHost;
-//   }
-
-//   /**
-//    * Public API key for the ODP account from which the audience segments will be fetched (optional).
-//    * @private
-//    */
-//   private _apiKey: string;
-
-//   /**
-//    * Getter to retrieve the ODP API key
-//    * @public
-//    */
-//   get apiKey(): string {
-//     return this._apiKey;
-//   }
-
-//   /**
-//    * Url for sending events via pixel.
-//    * @private
-//    */
-//   private _pixelUrl: string;
-
-//   /**
-//    * Getter to retrieve the ODP pixel URL
-//    * @public
-//    */
-//   get pixelUrl(): string {
-//     return this._pixelUrl;
-//   }
-
-//   /**
-//    * All ODP segments used in the current datafile (associated with apiHost/apiKey).
-//    * @private
-//    */
-//   private _segmentsToCheck: string[];
-
-//   /**
-//    * Getter for ODP segments to check
-//    * @public
-//    */
-//   get segmentsToCheck(): string[] {
-//     return this._segmentsToCheck;
-//   }
-
-//   constructor(apiKey?: string, apiHost?: string, pixelUrl?: string, segmentsToCheck?: string[]) {
-//     this._apiKey = apiKey ?? '';
-//     this._apiHost = apiHost ?? '';
-//     this._pixelUrl = pixelUrl ?? '';
-//     this._segmentsToCheck = segmentsToCheck ?? [];
-//   }
-
-//   /**
-//    * Update the ODP configuration details
-//    * @param {OdpConfig} config New ODP Config to potentially update self with
-//    * @returns true if configuration was updated successfully
-//    */
-//   update(config: OdpConfig): boolean {
-//     if (this.equals(config)) {
-//       return false;
-//     } else {
-//       if (config.apiKey) this._apiKey = config.apiKey;
-//       if (config.apiHost) this._apiHost = config.apiHost;
-//       if (config.pixelUrl) this._pixelUrl = config.pixelUrl;
-//       if (config.segmentsToCheck) this._segmentsToCheck = config.segmentsToCheck;
-
-//       return true;
-//     }
-//   }
-
-//   /**
-//    * Determines if ODP configuration has the minimum amount of information
-//    */
-//   isReady(): boolean {
-//     return !!this._apiKey && !!this._apiHost;
-//   }
-
-//   /**
-//    * Detects if there are any changes between the current and incoming ODP Configs
-//    * @param configToCompare ODP Configuration to check self against for equality
-//    * @returns Boolean based on if the current ODP Config is equivalent to the incoming ODP Config
-//    */
-//   equals(configToCompare: OdpConfig): boolean {
-//     return (
-//       this._apiHost === configToCompare._apiHost &&
-//       this._apiKey === configToCompare._apiKey &&
-//       this._pixelUrl === configToCompare._pixelUrl &&
-//       checkArrayEquality(this.segmentsToCheck, configToCompare._segmentsToCheck)
-//     );
-//   }
-// }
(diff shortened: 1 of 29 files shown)


lib/core/odp/odp_manager.ts
```diff
@@ -145,21 +211,26 @@ export abstract class OdpManager implements IOdpManager {
...
       return null;
     }
 
     if (VuidManager.isVuid(userId)) {
       return this.segmentManager.fetchQualifiedSegments(ODP_USER_KEY.VUID, userId, options);
     }
 
-    return this.segmentManager.fetchQualifiedSegments(ODP_USER_KEY.FS_USER_ID, userId, options);
+    const foo = await this.segmentManager.fetchQualifiedSegments(ODP_USER_KEY.FS_USER_ID, userId, options);
+    console.log('foo is ', foo);
```
muzahidul-opti, line 232: I think renaming `foo` to `fetchedSegtments` would improve the code readability.
 ```
 const fetchedSegments = await this.segmentManager.fetchQualifiedSegments(ODP_USER_KEY.FS_USER_ID, userId, options);
console.log('fetchedSegment is ', fetchedSegments);
return fetchedSegments;
```

File: lib/core/odp/odp_segment_manager.ts [hunk 2]
```diff
@@ -93,11 +93,15 @@ export class OdpSegmentManager implements IOdpSegmentManager {
...
     options: Array<OptimizelySegmentOption>
   ): Promise<string[] | null> {
-    const { apiHost: odpApiHost, apiKey: odpApiKey } = this.odpConfig;
-
-    if (!odpApiKey || !odpApiHost) {
-      this.logger.log(LogLevel.WARNING, ERROR_MESSAGES.FETCH_SEGMENTS_FAILED_INVALID_IDENTIFIER);
-      return null;
+    // const { apiHost: odpApiHost, apiKey: odpApiKey } = this.odpConfig;
+
+    // if (!odpApiKey || !odpApiHost) {
```
muzahidul-opti, line 98: Maybe we can remove the boiler code.

lib/core/odp/odp_manager.ts
raju-opti: My bad, this was intended for debug purposes. removed

lib/core/odp/odp_event_manager.ts
```diff
@@ -315,38 +333,28 @@ export abstract class OdpEventManager implements IOdpEventManager {
    * @private
    */
   private processQueue(shouldFlush = false): void {
```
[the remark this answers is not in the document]
raju-opti: processQueue is synchronous, but it is called asynchronously when timeout occurs during flush. If a batch becomes full when sendEvent() is called, then it is called synchronously. But e. And the apiManger eventApiManager.sendEvents is called asynchronously (https://github.com/optimizely/javascript-sdk/blob/8cbb767abf12e38f971d7c177eef53797de35fad/lib/core/odp/odp_event_manager.ts#L395). Also, the eventApiManager.sendEvents itself is asynchronous. So the event loop will not be blocked.

lib/core/odp/odp_config.ts
```diff
@@ -19,90 +19,29 @@ import { checkArrayEquality } from '../../utils/fns';
...
-  private _apiHost: string;
-
-  /**
-   * Getter to retrieve the ODP server host
-   * @public
-   */
-  get apiHost(): string {
-    return this._apiHost;
-  }
+  readonly apiHost: string;
```
mikechu-optimizely, line 23: NIT: Is it good to add the `public` access modifier for clarity?

lib/core/odp/odp_config.ts
```diff
@@ -112,10 +51,33 @@ export class OdpConfig {
...
+export type OdpNotIntegratedConfig = {
+  readonly integrated: false;
+}
+
+export type OdpIntegratedConfig = {
+  readonly integrated: true;
+  readonly odpConfig: OdpConfig;
+}
+
+export const odpIntegrationEquals = (config1: OdpIntegrationConfig, config2: OdpIntegrationConfig): boolean => {
```
mikechu-optimizely, line 71: NIT
```suggestion
export const odpIntegrationAreEqual = (config1: OdpIntegrationConfig, config2: OdpIntegrationConfig): boolean => {
```

lib/core/odp/odp_event_api_manager.ts
```diff
@@ -78,14 +64,9 @@ export abstract class OdpEventApiManager implements IOdpEventApiManager {
    * @param events ODP events to send
    * @returns Retry is true - if network or server error (5xx), otherwise false
    */
-  async sendEvents(events: OdpEvent[]): Promise<boolean> {
+  async sendEvents(odpConfig: OdpConfig, events: OdpEvent[]): Promise<boolean> {
```
mikechu-optimizely, line 67: Maybe adding `public` becomes too verbose. Bah. Nevermind.

lib/core/odp/odp_event_manager.ts
```diff
@@ -30,10 +30,9 @@ const MAX_RETRIES = 3;
 /**
  * Event dispatcher's execution states
  */
-export enum STATE {
-  STOPPED,
-  RUNNING,
-  PROCESSING,
```
mikechu-optimizely, line 36: Was `PROCESSING` used as a sort of lock while operating on an event queue?

mikechu-optimizely: Oh maybe not ... the way we are taking on / processing batches to send.

lib/core/odp/odp_event_manager.ts
```diff
@@ -374,27 +382,23 @@ export abstract class OdpEventManager implements IOdpEventManager {
...
-      setTimeout(async () => {
+      queueMicrotask(async () => {
         let shouldRetry: boolean;
         let attemptNumber = 0;
         do {
-          shouldRetry = await this.apiManager.sendEvents(batch);
+          shouldRetry = await this.apiManager.sendEvents(odpConfig, batch);
           attemptNumber += 1;
-        } while (shouldRetry && attemptNumber < MAX_RETRIES);
+        } while (shouldRetry && attemptNumber < this.retires);
```
mikechu-optimizely, line 401: NIT: `retries` 🛞🛞 😆

lib/core/odp/odp_event_manager.ts
```diff
@@ -454,4 +444,8 @@ export abstract class OdpEventManager implements IOdpEventManager {
   protected getLogger(): LogHandler {
     return this.logger;
   }
+
+  getQueue(): OdpEvent[] {
+    return this.queue;
+  }
```
mikechu-optimizely, line 450: super-NIT: Is it worthwhile to use a getter?

```suggestion
  public get queue(): OdpEvent[] {
     return this.queue;
    }
```

lib/core/odp/odp_manager.ts
```diff
@@ -55,86 +52,155 @@ export interface IOdpManager {
...
+
+    if (odpIntegrationConfig) {
+      this.updateSettings(odpIntegrationConfig);
     }
+  }
 
-    if (!this.eventManager) {
-      this.logger.log(LogLevel.ERROR, ERROR_MESSAGES.ODP_MANAGER_UPDATE_SETTINGS_FAILED_EVENT_MANAGER_MISSING);
-      return false;
+  public getStatus(): Status {
```
mikechu-optimizely, line 139: super-NIT: I love getters for semantic code, but I admit I'm being ridiculous. 😁

lib/core/project_config/project_config_manager.tests.js
```diff
@@ -256,21 +256,20 @@ describe('lib/core/project_config/project_config_manager', function() {
...
-          var newDatafile = testData.getTestProjectConfigWithFeatures();
-          newDatafile.revision = '36';
-          fakeDatafileManager.get.returns(newDatafile);
-
-          updateListener({ datafile: newDatafile });
-          sinon.assert.calledTwice(onUpdateSpy);
-        });
+        await manager.onReady();
+        sinon.assert.calledOnce(onUpdateSpy);
+        var fakeDatafileManager = datafileManager.HttpPollingDatafileManager.getCall(0).returnValue;
```
mikechu-optimizely, line 261: Love these test updates. Thanks.  

Is it worth moving to `let`?

lib/optimizely/index.ts
```diff
@@ -178,29 +175,11 @@ export default class Optimizely implements Client {
...
-          NotificationRegistry.getNotificationCenter(
-            sdkKey,
-            this.logger
-          )?.addNotificationListener(NOTIFICATION_TYPES.OPTIMIZELY_CONFIG_UPDATE, () => this.updateOdpSettings());
-        } else {
-          this.logger.log(LOG_LEVEL.ERROR, ERROR_MESSAGES.ODP_SDK_KEY_MISSING_NOTIFICATION_CENTER_FAILURE);
-        }
-      }
-
+    this.readyPromise = Promise.all([
```
mikechu-optimizely, line 178: I was considering, in the React SDK, if we should use `Promise.allSettled`.

mikechu-optimizely: No wait. That's only if we were using `reject` correctly in React.

tests/odpManager.spec.ts
```diff
@@ -89,142 +125,611 @@ describe('OdpManager', () => {
...
 
-  const odpManagerInstance = (config?: OdpConfig) =>
-    new OdpManager({
-      odpOptions: {
-        eventManager,
-        segmentManager,
-        segmentsRequestHandler: defaultRequestHandler,
-        eventRequestHandler: defaultRequestHandler,
-      },
+  // const odpManagerInstance = (config?: OdpConfig) =>
```
mikechu-optimizely, line 128: I'd say delete this unless you think we'll need to come back to it.

mikechu-optimizely: Oh or did you move these?

mikechu-optimizely (approved):
That was a long one phew. Thanks for putting all your thoughts in. 

👇 are just some non-show stopping, thoughts/suggestions.

lib/core/odp/odp_event_manager.ts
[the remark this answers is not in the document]
muzahidul-opti: Thanks for the explanation. Totally agree with this.

lib/core/odp/odp_config.ts
raju-opti: we don't have any consistent convention in our codebase yet. I prefer omitting the public though

lib/core/odp/odp_event_manager.ts
raju-opti: I think having the same name for the property and getter will cause problem, and didn't want to rename the property, hence ..

lib/core/project_config/project_config_manager.tests.js
raju-opti: yeah, but this is an old JS test file and var is used everywhere, should change everything at once

lib/optimizely/index.ts
raju-opti: We want the readyPromise to reject if any of them fail. But admittedly, this is not done consistently, for example, currently the promise can resolve with { success: false }

lib/core/odp/odp_config.ts
raju-opti: updated

lib/core/odp/odp_event_manager.ts
raju-opti: updated

mikechu-optimizely (approved):
LGTM

lib/optimizely/index.ts
```diff
@@ -178,29 +175,11 @@ export default class Optimizely implements Client {
...
-      if (config.odpManager != null) {
-        this.odpManager = config.odpManager;
-        this.odpManager.eventManager?.start();
-        this.updateOdpSettings();
-        const sdkKey = this.projectConfigManager.getConfig()?.sdkKey;
-        if (sdkKey != null) {
-          NotificationRegistry.getNotificationCenter(
-            sdkKey,
-            this.logger
-          )?.addNotificationListener(NOTIFICATION_TYPES.OPTIMIZELY_CONFIG_UPDATE, () => this.updateOdpSettings());
```
jaeopt, line 198: If you remove this, how do we get new odpconfig to ODPManager on changes?

jaeopt (changes requested):
One clarification - I see the updateOdpConfig on config change is broken. Can you take a look?

lib/core/odp/odp_event_manager.ts
```diff
@@ -283,10 +301,11 @@ export abstract class OdpEventManager implements IOdpEventManager {
    * @private
    */
   private enqueue(event: OdpEvent): void {
-    if (this.state === STATE.STOPPED) {
+    if (this.status === Status.Stopped) {
       this.logger.log(LogLevel.WARNING, 'Failed to Process ODP Event. ODPEventManager is not running.');
       return;
     }
+    const hasNecessaryIdentifiers = this.hasNecessaryIdentifiers;
```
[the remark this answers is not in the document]
raju-opti: this was added for debugging, (ts-mockito was not behaving correctly with this method mocking cause this is method of the subclass), removing

lib/core/odp/odp_event_manager.ts
```diff
@@ -315,38 +333,28 @@ export abstract class OdpEventManager implements IOdpEventManager {
...
       while (this.queueHasBatches()) {
         this.makeAndSend1Batch();
       }
     }
 
-    this.state = STATE.RUNNING;
-    this.setNewTimeout();
+    // no need for periodic flush if batchSize is 1
+    if (this.batchSize > 1) {
+      this.setNewTimeout();
```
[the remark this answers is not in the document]
raju-opti: this was in the previous implementation. I just wrapped it in the if, should I change it in any other way?

lib/optimizely/index.ts
raju-opti: this code segment iteself is inside a listener that is fired on project config update. you can see `      `this.updateOdpSettings();` immediately below this code snippet, which will update the odp config each time project config manager fires onUpdate

lib/core/odp/odp_event_manager.ts
```diff
@@ -374,27 +382,23 @@ export abstract class OdpEventManager implements IOdpEventManager {
...
-    // remove a batch from the queue
-    for (let count = 0; count < this.batchSize; count += 1) {
-      const event = this.queue.shift();
-      if (event) {
-        batch.push(event);
-      } else {
-        break;
-      }
+    if (!this.odpConfig) {
+      return;
```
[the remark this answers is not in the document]
raju-opti: This code just ensure to typescirpt that this.odpCofig is not undefined in the code that follows. This function is guaranteed to be not called when the event manager is not running

lib/core/odp/odp_event_manager.ts
```diff
@@ -315,38 +333,28 @@ export abstract class OdpEventManager implements IOdpEventManager {
...
+    if (this.status !== Status.Running) {
       return;
     }
 
-    if (!this.isOdpConfigurationReady()) {
-      return;
-    }
-
-    // Flush interval occurred & queue has items
+    this.clearCurrentTimeout();
```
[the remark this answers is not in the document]
raju-opti: updated

lib/optimizely/index.ts
jaeopt: I see. You just removed the redundant event handler. It looks good!

jaeopt (approved):
LGTM

PR #923, opened 2024-04-24
[ISSUE-922] - un-restrict user attribute type definition, add tests...

### LINKED ISSUE #922: [OptimizelyProvider UserAttributes Type Definition mismatch] The type definition for the user attributes does not match the documentation and is limiting types allowed
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SDK Version

3.1.0

### Current Behavior

The new package has defined the `UserAttributeValue` to be a limited set of types which contradicts the [documentation stating it supports any type](https://docs.developers.optimizely.com/feature-experimentation/docs/optimizelyprovider#props). What's happening now is that if you have a custom type being set as an attribute you will receive a typescript error because the type is not supported.

### Expected Behavior

The `user` prop in the `OptimizelyProvider` should allow any type definition as stated in the documentation instead of the currently defined type: 
```
export type UserAttributeValue = string | number | boolean | null;
```
⬇️ 
```
export type UserAttributeValue = any;
```
or simply:
```
export type UserAttributes = {
    [name: string]: UserAttributeValue;
};
```
⬇️ 
```
export type UserAttributes = {
    [name: string]: any;
};
```

### Steps To Reproduce

- setup a project with the lastest react sdk
- In the OptimizelyProvider's `user` prop, add a custom attribute that has a json value:
```
            <OptimizelyProvider
                optimizely={optimizelyInstance}
                user={{
                    id: profile?.id || null,
                    attributes: {
                        my_custom_attribute: { name: "something", type: "something" },
                    }
                }}
            >
                {children}
            </OptimizelyProvider>
```
Note the TS error:
image

### React Framework

create-react-app

### Browsers impacted

_No response_

### Link

_No response_

### Logs

_No response_

### Severity

Blocking development

### Workaround/Solution

Solution listed above. Workaround might be to ts-ignore and ensure the attribute is being used.

### Recent Change

_No response_

### Conflicts

_No response_

victor-optimizely:
## Summary
- Reverts functionality for the type definition on user attributes (allow _any_ type)
- adds some tests

## Test plan
See [#922]

## Issues
- Fixes #922

### DIFF
--- a/lib/shared_types.tests.ts
+++ b/lib/shared_types.tests.ts
@@ -1,0 +1,43 @@
+import { UserAttributes } from "./shared_types";
+import { assert } from "chai";
+
+describe("shared_types", () => {
+    describe("UserAttributes", () => {
+        it("allows any attribute to be defined", () => {
+            const booleanValue: UserAttributes = {
+                fixed: true,
+                broken: false
+            };
+            const numberValue: UserAttributes = {
+                first: 1,
+                last: 999
+            };
+            const stringValue: UserAttributes = {
+                name: "Rick Sanchez",
+                dimension: "c137"
+            };
+            const nullValue: UserAttributes = {
+                no: null,
+                empty: null
+            };
+            const custom = {
+                sidekick1: "Morty Smith",
+                sidekick2: "Summer Smith"
+            }
+            const objectValue: UserAttributes = {
+                custom
+            };
+            assert.isBoolean(booleanValue.fixed);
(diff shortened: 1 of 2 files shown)


victor-optimizely:
Our team deals with the admin center. We were setting the user's current instance object as a value here. After further investigation we are no longer using any of the fields in the object and moving forward we can just set the specific field we'll use. Since we're updating from 2.X -> 3.X a breaking change is expected I will close this. Thanks for looking into it.

PR #930, opened 2024-05-17
[FSSDK-10198] making isOdpIntegrated public

junaed-optimizely:
## Summary
Client should contain public facing API to provide information about ODP integration

## Test plan
Existing test should pass
## Issues
- [FSSDK-10198](https://jira.sso.episerver.net/browse/FSSDK-10198)

### DIFF
--- a/lib/core/odp/odp_manager.ts
+++ b/lib/core/odp/odp_manager.ts
@@ -216,7 +216,7 @@ export abstract class OdpManager implements IOdpManager {
     }
 
     if (!this.odpIntegrationConfig.integrated) {
-      this.logger.log(LogLevel.INFO, ERROR_MESSAGES.ODP_NOT_INTEGRATED);
+      this.logger.log(LogLevel.ERROR, ERROR_MESSAGES.ODP_NOT_INTEGRATED);
       return null;
     }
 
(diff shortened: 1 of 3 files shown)


lib/optimizely/index.ts
```diff
@@ -1723,7 +1719,7 @@ export default class Optimizely implements Client {
     }
   }
 
-  private isOdpIntegrated(): boolean {
+  public isOdpIntegrated(): boolean {
```
mikechu-optimizely, line 1722: Now that we're making it public facing, can you add jsdoc?

mikechu-optimizely (changes requested):
Small necessary change, but looks good.

lib/core/odp/odp_manager.ts
```diff
@@ -241,7 +240,7 @@ export abstract class OdpManager implements IOdpManager {
     }
 
     if (!this.odpIntegrationConfig.integrated) {
-      this.logger.log(LogLevel.ERROR, ERROR_MESSAGES.ODP_NOT_INTEGRATED);
+      this.logger.log(LogLevel.INFO, ERROR_MESSAGES.ODP_NOT_INTEGRATED);
```
jaeopt, line 243: This change looks good. We also need to change ERROR -> to INFO for "registerVuid()" api call as well.

jaeopt (changes requested):
One required change

lib/optimizely/index.ts
```diff
@@ -1718,7 +1718,10 @@ export default class Optimizely implements Client {
       this.logger.error(ERROR_MESSAGES.ODP_EVENT_FAILED, e);
     }
   }
-
+  /**
+   * Checks if ODP (Optimizely Data Platform) is integrated into the project.
+   * @returns { boolean }
```
mikechu-optimizely, line 1723:
```suggestion
   * @returns { boolean } `true` if ODP settings were found in the datafile otherwise `false`
```

mikechu-optimizely (approved):
LGTM.

jaeopt (approved):
LGTM

PR #936, opened 2024-06-27
[FSSDK-10316] crypto and text encoder polyfill addition for React native

junaed-optimizely:
## Summary
- `crypto` module is not supported by React native. Our SDK currently throws "crypto.getRandomValues() not available" error. A polyfill has been added to avoid that issue.
- Hermes (React native: JS engine) did not have `TextEncoder` support until recently. A polyfill has been added for react-native <= 0.73

## Test plan
Modified existing React Native test to adjust with the current changes. Existing test should pass
## Issues
[FSSDK-10316](https://jira.sso.episerver.net/browse/FSSDK-10316)


__mocks__/fast-text-encoding.ts
mikechu-optimizely, line 1: NIT: please add an additional line at the end of the files

__mocks__/react-native-get-random-values.ts
```diff
@@ -0,0 +1 @@
+export {}
```
mikechu-optimizely, line 1: NIT: We should create a pre-commit hook to handle formatting 😁 Maybe biome

mikechu-optimizely (approved):
I added some nit but LGTM. Thanks for const+let

__mocks__/react-native-get-random-values.ts
junaed-optimizely: Biome sounds great. I have heard great things about it. But it will require some efforts to cleanup existing eslint + prettier and integrate Biome. Maybe another ticket?

PR #946, opened 2024-09-24
[FSSDK-10665] fix: Github Actions YAML files vulnerable to script injections corrected

FarhanAnjum-opti:
## Summary
Instead of directly using variables that can contain user input, new intermediate variables were introduced so that script injection attacks could be mitigated. Also **default environment variables** were used wherever possible.

## Issues
- [FSSDK-10665](https://jira.sso.episerver.net/browse/FSSDK-10665)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,7 +44,6 @@ jobs:
           LATEST_RELEASE_TAG="${{ steps.latest-release.outputs['latest-release-tag']}}"
           
           if [[ ${{ github.event_name }} == "workflow_dispatch" ]]; then
-            GITHUB_REF=$GITHUB_REF
             RELEASE_TAG=${GITHUB_REF#refs/tags/}
           else
             RELEASE_TAG=$ALTERNATE_RELEASE_TAG
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,20 +32,22 @@ jobs:
           echo "latest-release-tag=$(curl -qsSL \
             -H "Accept: application/vnd.github+json" \
             -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -32,20 +32,21 @@ jobs:
...
             -H "Accept: application/vnd.github+json" \
             -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
-            "${{ github.api_url }}/repos/${{ github.repository }}/releases/latest" \
+            "$GITHUB_API_URL/repos/$GITHUB_REPOSITORY/releases/latest" \
           | jq -r .tag_name)" >> $GITHUB_OUTPUT
 
       - id: npm-tag
         name: Determine NPM tag
+        env:
+          ALTERNATE_RELEASE_TAG: ${{ github.event.release.tag_name }}
```
raju-opti, line 41: Can we rename it to GITHUB_RELEASE_TAG instead?

raju-opti (changes requested):

.github/workflows/release.yml
FarhanAnjum-opti: Sure. Renaming it.

.github/workflows/release.yml
```diff
@@ -32,20 +32,21 @@ jobs:
           echo "latest-release-tag=$(curl -qsSL \
             -H "Accept: application/vnd.github+json" \
             -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
-            "${{ github.api_url }}/repos/${{ github.repository }}/releases/latest" \
+            "$GITHUB_API_URL/repos/$GITHUB_REPOSITORY/releases/latest" \
```
mikechu-optimizely, line 35: Are GITHUB_API_URL and GITHUB_REPOSITORY some defaults in all workflows or do we have to set them in an `env:` block?

raju-opti (approved):
LGTM

PR #950, opened 2024-10-11
[FSSDK-10711] Make use of VUID as an opt-in

mikechu-optimizely:
## Summary
- Client-side use of the visitor user ID (VUID) should be opt-in
- VuidManager has been hoisted to the top optimizely client level
- Additional devX and doc improvements

⚠️ NOTE: this is not being merged to the `master` branch yet, but will be released.  We'll need to cherry-pick later.

## Test plan

- Existing tests should pass
- New tests should also pass
- FSC end-to-end tests should pass

## Issues
[FSSDK-10711](https://jira.sso.episerver.net/browse/FSSDK-10711)


raju-opti (changes requested):
We also need to handle the following requirement from the TDD:
* createUserContext() will be rejected when userId is not provided

lib/plugins/odp_manager/index.browser.ts
```diff
@@ -53,14 +51,14 @@ interface BrowserOdpManagerConfig {
 // Client-side Browser Plugin for ODP Manager
 export class BrowserOdpManager extends OdpManager {
   static cache = new BrowserAsyncStorageCache();
-  vuidManager?: VuidManager; 
-  vuid?: string;
+  vuid?: string = "";
```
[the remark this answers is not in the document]
mikechu-optimizely: Ah. I was taking a cue from other tests that [look for empty string for user ID](https://github.com/optimizely/javascript-sdk/blob/5.x.x/lib/optimizely/index.tests.js#L6436) (flag key, etc) and added my unit tests to match.  

I'll update `vuid` to be `undefined` or a value....

lib/index.browser.ts
```diff
@@ -146,6 +154,7 @@ const createInstance = function(config: Config): Client | null {
       isValidInstance,
       odpManager: odpExplicitlyOff ? undefined
         : BrowserOdpManager.createInstance({ logger, odpOptions: config.odpOptions, clientEngine, clientVersion }),
+      vuidManager: await VuidManager.instance(cache, vuidManagerOptions),
```
raju-opti, line 157: let's get rid of this VuidManager.instance. We don't need to have it as a singleton. In the VuidManager, we will keep the vuidEnabled option in a class field. Here can just use new VuidManager(cache, option).

Let's move the initialization in a init() method in VuidManager. This is async. In the optimizely class, we should add vuidManager.init() in the readyPromise dependency.

lib/plugins/vuid_manager/index.ts
```diff
@@ -64,12 +68,17 @@ export class VuidManager implements IVuidManager {
 
   /**
    * Gets the current instance of the VUID Manager, initializing if needed
-   * @param cache Caching mechanism to use for persisting the VUID outside working memory   *
+   * @param cache Caching mechanism to use for persisting the VUID outside working memory
+   * @param options Options for the VUID Manager
    * @returns An instance of VuidManager
    */
-  static async instance(cache: PersistentKeyValueCache): Promise<VuidManager> {
+  static async instance(cache: PersistentKeyValueCache, options: VuidManagerOptions): Promise<VuidManager> {
```
raju-opti, line 75: let's rename this to init() make it an instance method, not a static one. Also add this to the VuidManager interface.

lib/core/odp/odp_manager.ts
```diff
@@ -112,22 +108,14 @@ export abstract class OdpManager implements IOdpManager {
...
-    if (this.isVuidEnabled()) {
-      readinessDependencies.push(this.initializeVuid());
-    }
-
     this.initPromise = Promise.all(readinessDependencies);
 
     this.onReady().then(() => {
       this.ready = true;
-      if (this.isVuidEnabled() && this.status === Status.Running) {
-        this.registerVuid();
```
raju-opti, line 129: this needs to move to the optimizely class. Let's make the registerVuid method below part of the OdpManager interface, and make it take the vuid as a parameter. Then in the optimizely constructor, we should do this

`this.onReady().then(() => {
  if (this.isVuidEnabled() && this.odpManager) {
    this.odpManager.registerVuid(this.vuidManager.vuid)
  }
})`

lib/core/odp/odp_manager.ts
```diff
@@ -283,41 +271,4 @@ export abstract class OdpManager implements IOdpManager {
...
-  /**
-   * Returns VUID value if it exists
-   */
-  abstract getVuid(): string | undefined;
-
-  protected initializeVuid(): Promise<void> {
-    return Promise.resolve();
-  }
-
-  private registerVuid() {
```
raju-opti, line 301: let make it `public registerVuid(vuid: string)`

lib/index.browser.ts
```diff
@@ -51,7 +53,8 @@ let hasRetriedEvents = false;
  * @return {Client|null} the Optimizely client object
  *                           null on error
  */
-const createInstance = function(config: Config): Client | null {
+// TODO: @raju-opti I'm not sure how to handle async VuidManager.instance() making createInstance async
```
raju-opti, line 56: suggested changes in other comments. This should stay sync.

raju-opti (changes requested):

lib/plugins/vuid_manager/index.ts
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import { VuidManagerOptions } from '../../shared_types';
```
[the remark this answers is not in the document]
raju-opti: We might want to have few other parameters in future, like whether already cached vuid should be removed from localstorage etc.

lib/core/odp/odp_manager.ts
```diff
@@ -47,11 +46,12 @@ export interface IOdpManager {
 
   sendEvent({ type, action, identifiers, data }: OdpEvent): void;
 
-  isVuidEnabled(): boolean;
-
-  getVuid(): string | undefined;
+  registerVuid(vuid: string | undefined): void;
```
raju-opti, line 49: can we make the parameter just string?

lib/index.react_native.ts
```diff
@@ -108,6 +109,11 @@ const createInstance = function(config: Config): Client | null {
 
     const { clientEngine, clientVersion } = config;
 
+    const cache = new BrowserAsyncStorageCache();
```
raju-opti, line 112: There is ReactNativeAsyncStorageCache, we should probably use that instead

lib/plugins/odp_manager/index.browser.ts
```diff
@@ -169,35 +147,29 @@ export class BrowserOdpManager extends OdpManager {
...
-  }
+    if (!this.odpIntegrationConfig.integrated) {
+      this.logger.log(LogLevel.INFO, ERROR_MESSAGES.ODP_NOT_INTEGRATED);
+      return;
+    }
 
-  isVuidEnabled(): boolean {
-    return true;
-  }
+    if (!vuid || !VuidManager.isVuid(vuid)) {
```
raju-opti, line 164: can we remove this check by removing undefined from the vuid parameter type and assuming the caller mast pass a valid vuid?

raju-opti (approved):
A few non critical comments

lib/plugins/vuid_manager/index.ts
```diff
@@ -44,48 +65,58 @@ export class VuidManager implements IVuidManager {
...
-  private _vuid: string;
+  private _vuid: string | undefined;
 
   /**
    * Get the current VUID value being used
    */
-  get vuid(): string {
+  get vuid(): string | undefined {
+    if (!this._vuid) {
+      this.logger.log(LogLevel.ERROR, 'VUID is not initialized. Please call initialize() before accessing the VUID.');
```
jaeopt, line 75: Wondering if we need error log here. Also, it looks like redudnant log for getVuid() function above.

lib/plugins/vuid_manager/index.ts
```diff
@@ -44,48 +65,58 @@ export class VuidManager implements IVuidManager {
...
+  async initialize(): Promise<void> {
+    if (!this.vuidEnabled) {
+      await this.cache.remove(this._keyForVuid);
+      return;
     }
 
-    if (!this._instance._vuid) {
-      await this._instance.load(cache);
+    if (!this._vuid) {
+      await this.load(this.cache);
```
jaeopt, line 118: It looks like this load creates/saves _vuid. We should skip it if vuid disabled.

jaeopt (changes requested):
I see most issues fixed. A few more comments.

lib/plugins/vuid_manager/index.ts
mikechu-optimizely: Is the early return above on L110 a good way to skip the load?

jaeopt: Looks like. should skip any vuid action other than remove you did  in 110.

jaeopt (approved):
LGTM

PR #951, opened 2024-10-21
[FSSDK-10797] send vuid automatically with odp events

raju-opti:
## Summary
- vuid is added automatically to all odp events

## Test plan
- new test added

## Issues
- FSSDK-10797

### DIFF
--- a/lib/core/odp/odp_event_manager.ts
+++ b/lib/core/odp/odp_event_manager.ts
@@ -46,7 +46,7 @@ export interface IOdpEventManager {
 
   stop(): Promise<void>;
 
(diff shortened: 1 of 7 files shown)


lib/core/odp/odp_event_manager.ts
mikechu-optimizely, line 1: Thanks for rename.  More align with actual action.

mikechu-optimizely (approved):
LGTM

jaeopt (approved):
LGTM - we need to make sure this vuid-in-events should be covered in FSC.
Wondering if we missed it before or have bugs in the test set up.

PR #955, opened 2024-11-05
[FSSDK-10843] prepare for release 6.0.0

raju-opti:
## Issues
- FSSDK-10843

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@optimizely/optimizely-sdk",
-  "version": "5.3.4",
+  "version": "6.0.0",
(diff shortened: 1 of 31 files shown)


package.json
```diff
@@ -94,7 +94,7 @@
   },
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=14.0.0"
+    "node": ">=16.0.0"
```
junaed-optimizely, line 97: Should we set this to 18 as Node.JS officially does not support 16 anymore ?

junaed-optimizely (changes requested):

package.json
raju-opti: we might do that in v7. Usually we try to support one previous version

junaed-optimizely (approved):
LGTM!

PR #956, opened 2024-11-06
[FSSDK-10867] ensure odpManager.setVuid is done before resolving onReady

raju-opti:
## Summary
- ensure odpManager.setVuid is done before resolving onReady so that identify event contains the vuid.

## Test plan
- existing tests should pass

## Issues
- FSSDK-10867

### DIFF
--- a/lib/optimizely/index.ts
+++ b/lib/optimizely/index.ts
@@ -181,28 +181,26 @@ export default class Optimizely implements Client {
     this.eventProcessor = config.eventProcessor;
 
     const eventProcessorStartedPromise = this.eventProcessor.start();
-
+    
     this.readyPromise = Promise.all([
       projectConfigManagerReadyPromise,
       eventProcessorStartedPromise,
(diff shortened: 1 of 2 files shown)


lib/optimizely/index.ts
```diff
@@ -181,28 +181,26 @@ export default class Optimizely implements Client {
...
       config.vuidManager ? config.vuidManager.configure(this.vuidOptions ?? { enableVuid: false }) : Promise.resolve(),
     ]).then(promiseResults => {
-      // Only return status from project config promise because event processor promise does not return any status.
-      return promiseResults[0];
-    });
-
-    this.readyTimeouts = {};
-    this.nextReadyTimeoutId = 0;
-
-    this.onReady().then(({ success }) => {
```
junaed-optimizely, line 198: Out of curiosity, why is this before to after transition ?

junaed-optimizely: So, is it a bug in the production now? Or this is just an adjustment with the upcoming vuid changes ?

PR #961, opened 2024-11-15
[FSSDK-10464] update import for uuid

raju-opti:
## Summary
- update the way uuid is imported cause it was causing issues for some clients

## Test plan

## Issues
- FSSDK-10464


jaeopt (approved):
LGTM

PR #965, opened 2024-11-18
[FSSDK-10882] ProjectConfigManager SSR support

junaed-optimizely:
## Summary
This PR makes the `ProjectConfigManager` SSR compatible
## Test plan
- Existing mock has been updated to adjust with the config manager
- New testcases have been added for coverage
## Issues
- [FSSDK-10882](https://jira.sso.episerver.net/browse/FSSDK-10882)


lib/project_config/project_config_manager.ts
```diff
@@ -79,6 +81,11 @@ export class ProjectConfigManagerImpl extends BaseService implements ProjectConf
       return;
     }
 
+    if(this.isSsr) {
```
raju-opti, line 84: if isSsr is true no datafile is provided, but only a datafileManager is provided, then onRunning() will wait indefinitely cause we are dropping the datafile. We can move this condition before the previous if on line 78 and update the message inside for isSsr case

raju-opti: Also can we add a test that if isSsr is provided without a datafile, onRunning() is rejected?

lib/optimizely/index.ts
```diff
@@ -146,6 +146,7 @@ export default class Optimizely implements Client {
       this.updateOdpSettings();
     });
 
+    this.projectConfigManager.setSsr(config.isSsr)
```
raju-opti, line 149: can we add a test that isSsr is passed to the projectConfigManager?

Please do this in a new lib/optimizely/index.spec.ts file instead of the old js test file. We will eventually get rid of the whole js test file.

lib/index.node.tests.js
```diff
@@ -88,6 +89,22 @@ describe('optimizelyFactory', function() {
         assert.equal(optlyInstance.clientVersion, '5.3.4');
       });
 
+      it('should create an instance of optimizely with ssr flag, project config must be available', () => {
```
raju-opti, line 92: could you please add the test in a new index.node.spec.ts file instead? We want to get rid of all JS tests. We will only update existing tests in these old tests file and will add all new tests in new ts test files.

lib/project_config/project_config_manager.ts
junaed-optimizely: Good point. I have missed that edge case. Thanks for pointing it out..

lib/optimizely/index.ts
junaed-optimizely: Ok. I did not know about that change.

lib/index.node.tests.js
junaed-optimizely: This won't be required anymore due to the new `index.spec.ts` coverage

PR #967, opened 2024-11-25
[FSSDK-10941] event processor files and directories cleanup - part 2

raju-opti:
## Summary
- event processor files and directories cleanup - part 2

## Issues
- FSSDK-10941


lib/event_processor/event_builder/log_event.ts
```diff
@@ -258,10 +257,10 @@ export function buildConversionEventV1(data: ConversionEvent): EventV1 {
   }
 }
 
-export function formatEvents(events: ProcessableEvent[]): Event {
+export function buildLogEvent(events: UserEvent[]): LogEvent {
```
junaed-optimizely, line 260: NIT: Wouldn't it be better to make it plural as `buildLogEvents` ?

junaed-optimizely (approved):
LGTM! Thanks again for the cleanup and naming improvements. I just have one change suggestion.

lib/event_processor/event_builder/log_event.ts
raju-opti: One LogEvent is a container for multiple UserEvents. This is also called LogEvent in other SDKs

PR #974, opened 2024-12-09
[FSSDK-10985] Implement log message file generator

raju-opti:
## Summary
- implement generator to convert log message files

## Test plan
- tested the script manually

## Issues
- FSSDK-10985

### DIFF
--- a/messageGenerator.ts
+++ b/messageGenerator.ts
@@ -1,0 +1,36 @@
+import path from 'path';
+import { writeFile } from 'fs/promises';
+
+const generate = async () => {
+  const inp = process.argv.slice(2);
+  for(const filePath of inp) {
+    console.log('generating messages for: ', filePath);
+    const parsedPath = path.parse(filePath);
+    const fileName = parsedPath.name;
+    const dirName = parsedPath.dir;
+    const ext = parsedPath.ext;
+
+    const genFilePath = path.join(dirName, `${fileName}.gen${ext}`);
+    console.log('generated file path: ', genFilePath);
+    const exports = await import(filePath);
+    const messages : Array<any> = [];
+
+    let genOut = '';
+
+    Object.keys(exports).forEach((key, i) => {
+      const msg = exports[key];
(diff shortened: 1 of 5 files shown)


messageGenerator.ts
```diff
@@ -0,0 +1,36 @@
...
+    await writeFile(genFilePath, genOut, 'utf-8');
+  };
+}
+
+try {
+  generate();
+} catch(e) {
+  console.error(e);
+  process.exit(1);
+}
```
junaed-optimizely, line 36: This will only catch synchronous exceptions. It might be better to wrap this with an Immediately invoked async function and add `await` before `generate` to handle async exceptions as well

package.json
```diff
@@ -88,7 +88,8 @@
     "coveralls": "nyc --reporter=lcov npm test",
     "prepare": "npm run build",
     "prepublishOnly": "npm test && npm run test-ci",
-    "postbuild:win": "@powershell copy \"dist/index.lite.d.ts\" \"dist/optimizely.lite.es.d.ts\" && @powershell copy \"dist/index.lite.d.ts\" \"dist/optimizely.lite.es.min.d.ts\" && @powershell copy \"dist/index.lite.d.ts\" \"dist/optimizely.lite.min.d.ts\""
+    "postbuild:win": "@powershell copy \"dist/index.lite.d.ts\" \"dist/optimizely.lite.es.d.ts\" && @powershell copy \"dist/index.lite.d.ts\" \"dist/optimizely.lite.es.min.d.ts\" && @powershell copy \"dist/index.lite.d.ts\" \"dist/optimizely.lite.min.d.ts\"",
+    "genmsg": "jiti messageGenerator ./lib/errorMessages.ts"
```
junaed-optimizely, line 92: I would prefer this filename to be consistent with the rest. So insteand of `errorMessages.ts` it would be better to name this `error_messsages.ts`.

junaed-optimizely (changes requested):

PR #977, opened 2024-12-11
[FSSDK-10993] additional cleanup for project config manager

raju-opti:
## Issues
- FSSDK-10993


lib/project_config/config_manager_factory.ts
```diff
@@ -55,15 +58,25 @@ export const getPollingConfigManager = (
   const backoff = new ExponentialBackoff(1000, updateInterval, 500);
   const repeater = new IntervalRepeater(updateInterval, backoff);
 
+  let startupLogs: StartupLog[] = []
```
junaed-optimizely, line 61: NIT: We can use const here.

junaed-optimizely (approved):
Thanks for the improvements. LGTM! Just one NIT

PR #979, opened 2024-12-13
[FSSDK-11006] remove assign function and use spreading

raju-opti:
## Test plan

## Issues
- FSSDK-11006


junaed-optimizely (approved):
LGTM!

PR #987, opened 2025-01-22
[FSSDK-11090] use path alias for log message imports

raju-opti:
## Summary
add aliases for log message modules, which allows using a different generated file during build, facilitating tree shaking of log messages.

## Test plan

## Issues
- FSSDK-11090


junaed-optimizely (approved):
LGTM!

PR #989, opened 2025-01-24
[FSSDK-11090] remove unused plugins directory

raju-opti:
## Issues
- FSSDK-11090


lib/optimizely/index.tests.js
```diff
@@ -314,7 +307,7 @@ describe('lib/optimizely', function() {
       optlyInstance = new Optimizely({
         clientEngine: 'node-sdk',
         projectConfigManager: mockConfigManager,
-        errorHandler: errorHandler,
+
```
junaed-optimizely, line 310: NIT: Seems like `errorHandler` is replaced with empty lines.

junaed-optimizely (approved):
LGTM! Just one NIT

lib/optimizely/index.tests.js
raju-opti: updated

Issue #924, opened 2024-04-26
[BUG] Using client.createUserContext leads to setTimeout is not defined error on Akamai EdgeWorker


dsafonov-grid:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SDK Version

5.3.0

### Current Behavior

```javascript
const optimizelyUserContext = optimizelyClient.createUserContext(userId, udo);
```
leads to 

ReferenceError setTimeout is not defined ReferenceError: setTimeout is not defined%0A at e.onReady (main.js:1:84427)%0A at new e (main.js:1:28336)%0A at e.createUserContext (main.js:1:84961)%0A at main.js:1:93456%0A at Generator.next (<anonymous>)%0A at main.js:1:461%0A at new Promise (<anonymous>)%0A at a (main.js:1:206)%0A at jr (main.js:1:92339)

### Expected Behavior

Works same as before

### Steps To Reproduce

1. In Akamai Edge Env
2. Run your starter kit https://github.com/optimizely/akamai-edgeworker-starter-kit with 5.3.0

### SDK Type

Edge/Lite

### Workaround/Solution

The cause of this behavior I think is: calling **onReady** https://github.com/optimizely/javascript-sdk/blob/75418da42a93dd94e63885685d3c55605fdaf6d4/lib/optimizely_user_context/index.ts#L67
because stack trace pointed to this https://github.com/optimizely/javascript-sdk/blob/75418da42a93dd94e63885685d3c55605fdaf6d4/lib/optimizely/index.ts#L1407

I can see that you pass **shouldIdentifyUser** as _true_ to createUserContext all the time. 
And another function **shouldIdentifyUser** as _false_ createInternalUserContext, is private, so for the time being is it ok to use createInternalUserContext and to suppress typescript? 

This approach works fine
```typescript
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-expect-error
    const optimizelyUserContext = optimizelyClient.createInternalUserContext(userId, udo);
```

junaed-optimizely:
Hey @dsafonov-grid, [v5.3.5](https://github.com/optimizely/javascript-sdk/releases/tag/v5.3.5) is out with the fix!

Thanks!

PR #995, opened 2025-01-29
[FSSDK-11098] using readyPromise instead of onReady

junaed-optimizely:
## Summary
- #924 has been addressed

## Test plan
- Existing test should pass
## Issues
- [FSSDK-11098](https://jira.sso.episerver.net/browse/FSSDK-11098)


raju-opti (approved):
LGTM

PR #1000, opened 2025-02-06
Junaed/fssdk 11034 tests js to ts

junaed-optimizely:
## Summary
Test conversions from JS (Mocha, Chai, Sinon) to TS (Vitest).

## Test plan
- TS test addition

## Issues
- [11034](https://jira.sso.episerver.net/browse/FSSDK-11034)


lib/project_config/optimizely_config.tests.js
```diff
@@ -1,916 +0,0 @@
-/**
```
raju-opti, line 1: please keep the old files for now. We will remove them later

lib/project_config/project_config.spec.ts
```diff
@@ -0,0 +1,1132 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { describe, it, expect, beforeEach, afterEach, vi, assert, Mock } from 'vitest';
+import { forEach, cloneDeep } from 'lodash';
+import { sprintf } from '../utils/fns';
+import fns from '../utils/fns';
```
raju-opti, line 19: we should get rid of the default export and only used named exports (for better tree-shaking. Please use named exports, and later we will remove the default export when every default import is re written

lib/project_config/project_config.tests.js
```diff
@@ -1,974 +0,0 @@
-/**
```
raju-opti, line 1: please keep the old tests for now

package.json
```diff
@@ -119,6 +119,7 @@
     "@rollup/plugin-commonjs": "^11.0.2",
     "@rollup/plugin-node-resolve": "^7.1.1",
     "@types/chai": "^4.2.11",
+    "@types/lodash": "^4.17.15",
```
raju-opti, line 122: I think we should just get rid of lodash at some point. forEach import from lodash is not needed, and for cloneDeep we can use JSON stringify and parse. We can remove loadsh when every usage of it is removed

raju-opti (changes requested):

raju-opti (approved):
LGTM

PR #1004, opened 2025-02-17
[FSSDK-11114] add type test for entrypoints

raju-opti:
## Summary
- add vitest [type tests](https://vitest.dev/guide/testing-types) for the entrypoints to ensure that all three entrypoints export the same api

## Test plan
- added type test for entrypoints

## Issues
- FSSDK-11114


junaed-optimizely (approved):
LGTM!

PR #1007, opened 2025-02-20
[FSSDK-10992] refactor build and export map

raju-opti:
## Test plan
- manually tested with demo projects for node/browser

## Issues
- FSSDK-10992


junaed-optimizely (approved):
LGTM!

PR #1008, opened 2025-02-24
[FSSDK-10992] add universal entrypoint

raju-opti:
## Summary
The universal entrypoint does not contain any platform specific code, so that it can be used with any platform including the edge platforms

## Issues
- FSSDK-10992


junaed-optimizely (approved):
LGTM!

PR #1010, opened 2025-02-26
[FSSDK-11125] implement CMAB client

raju-opti:
## Test plan
- added new tests 

## Issues
- FSSDK-11125

### DIFF
--- a/lib/core/decision_service/cmab/cmab_client.ts
+++ b/lib/core/decision_service/cmab/cmab_client.ts
@@ -29,7 +29,7 @@ export interface CmabClient {
     experimentId: string,
     userId: string,
     attributes: UserAttributes,
-    cmabUuid: string
+    cmabUuid: string,
   ): Promise<string>
 }
 
@@ -58,7 +58,7 @@ export class DefaultCmabClient implements CmabClient {
     experimentId: string,
     userId: string,
     attributes: UserAttributes,
-    cmabUuid: string
+    cmabUuid: string,
   ): Promise<string> {
     const url = sprintf(CMAB_PREDICTION_ENDPOINT, experimentId);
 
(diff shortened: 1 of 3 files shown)


lib/core/decision_service/cmab/cmab_client.ts
```diff
@@ -0,0 +1,116 @@
...
+  private requestHandler: RequestHandler;
+  private retryConfig?: RetryConfig;
+
+  constructor(config: CmabClientConfig) {
+    this.requestHandler = config.requestHandler;
+    this.retryConfig = config.retryConfig;
+  }
+
+  async fetchVariation(
+    experimentId: string,
```
jaeopt, line 58:
```suggestion
  async fetchDecision(
    ruleId: string,
```
I understand cmab server expects "experimentId" as the param. 
We can consider using "rule" and "decision" consistently internally in SDKs and map to cmab server "variation" and "experimentId".

lib/core/decision_service/cmab/cmab_client.ts
```diff
@@ -0,0 +1,116 @@
...
+      value: attributes[key],
+      type: 'custom_attribute',
+    }));
+
+    const body = {
+      instances: [
+        {
+          visitorId: userId,
+          experimentId,
+          attributes: cmabAttributes,
```
jaeopt, line 76:
```suggestion
          attributes: cmabAttributes,
          experimentId,
```
This may be too picky, but we want all SDKs copy the best definition :)

jaeopt (changes requested):
Looks good. Some suggestions about terms. Let me know if not happy with the new terms :)

jaeopt (approved):
LGTM

PR #1022, opened 2025-04-03
[FSSDK-11392] add support for async user profile service

raju-opti:
## Summary
- this PR adds support for async user profile service which will be useful for server side usage of the SDK.

## Test plan
- added tests

## Issues
- FSSDK-11392


jaeopt (approved):
LGTM

PR #1029, opened 2025-04-22
[FSSDK-11399] support traffic allocation for cmab

raju-opti:
## Summary
support traffic allocation for cmab experiment rules. A bucket call is first made to check if the user is in the experiment before calling the cmab service.

## Test plan
- added test for cmab bucketing

## Issues
- FSSDK-11399

### DIFF
--- a/lib/core/decision_service/index.ts
+++ b/lib/core/decision_service/index.ts
@@ -27,7 +27,6 @@ import {
   getExperimentFromId,
   getExperimentFromKey,
   getFlagVariationByKey,
-  getTrafficAllocation,
   getVariationIdFromExperimentAndVariationKey,
   getVariationFromId,
   getVariationKeyFromId,
@@ -44,6 +43,7 @@ import {
   FeatureFlag,
   OptimizelyDecideOption,
   OptimizelyUserContext,
+  TrafficAllocation,
   UserAttributes,
   UserProfile,
   UserProfileService,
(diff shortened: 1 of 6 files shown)


lib/core/decision_service/index.ts
```diff
@@ -573,6 +593,14 @@ export class DecisionService {
     bucketingId: string,
     userId: string
   ): BucketerParams {
+    let trafficAllocationConfig: TrafficAllocation[] = experiment.trafficAllocation;
+    if (experiment.cmab) {
+      trafficAllocationConfig = [{
+        entityId: CMAB_DUMMY_ENTITY_ID,
+        endOfRange: experiment.cmab.trafficAllocation
+      }];
+    }
```
jaeopt, line 602: A good idea :)

lib/shared_types.ts
```diff
@@ -159,6 +159,7 @@ export interface Experiment {
   forcedVariations?: { [key: string]: string };
   isRollout?: boolean;
   cmab?: {
+    trafficAllocation: number;
     attributeIds: string[];
```
jaeopt, line 163: in TDD, it is defined as "attributes" instead of "attributeIds". I use "attributes" in my datafile PRs.

jaeopt: I like this "attributeIds" instead of "attributes". It's consistent with other names in datafile.
Let's change keep it here and change TDD. Let me know so I can change them in my PR too.

jaeopt (approved):
LGTM - lovely. Let me know about TDD name change in my comment.

lib/shared_types.ts
raju-opti: Yeah. The datafile field is attributes, we rename it to attributeIds in the sdk when parsing. Probably better to change the datafile field to attributeIds as well. I'll create a separate PR

PR #1031, opened 2025-04-23
[FSSDK-11403] init doc update

junaed-optimizely:
## Summary
Doc update for JS SDK v6

## Test plan
Existing test should pass

## Issues
- [FSSDK-11403](https://jira.sso.episerver.net/browse/FSSDK-11403)


README.md
```diff
@@ -32,95 +33,146 @@ Optimizely Rollouts is [free feature flags](https://www.optimizely.com/free-feat
...
-<script src="https://unpkg.com/@optimizely/optimizely-sdk/dist/optimizely.browser.umd.min.js"></script>
-
-<!-- You can also use the unminified version if necessary -->
-<script src="https://unpkg.com/@optimizely/optimizely-sdk/dist/optimizely.browser.umd.js"></script>
-```
-
-When evaluated, that bundle assigns the SDK's exports to `window.optimizelySdk`. If you wish to use the asset locally (for example, if unpkg is down), you can find it in your local copy of the package at dist/optimizely.browser.umd.min.js. We do not recommend using this method in production settings as it introduces a third-party performance dependency.
-
-As `window.optimizelySdk` should be a global variable at this point, you can continue to use it like so:
+### Initialization (Using NPM)
```
raju-opti, line 80: can we update it to:

Initialization (Using package managers e.g NPM, Yarn etc)

raju-opti (approved):
LGTM with a small comment

PR #1039, opened 2025-05-02
[FSSDK-11197] EventTags type fix

junaed-optimizely:
## Summary
`EventTags` type fix.
The issue that has been raised in React SDK - https://github.com/optimizely/react-sdk/issues/303, that the $opt_event_properties does not support object even though the doc demonstrated so - https://docs.developers.optimizely.com/feature-experimentation/docs/track-event-javascript#example. 

## Test plan
Existing test should pass

## Issues
[FSSDK-11197](https://jira.sso.episerver.net/browse/FSSDK-11197)

PR #1040, opened 2025-05-02
[FSSDK-11197] EventTags type fix for v5

junaed-optimizely:
## Summary
`EvenTags` type fix.

## Test plan
Existing test should pass

## Issues
FSSDK-11197

### DIFF
--- a/lib/shared_types.ts
+++ b/lib/shared_types.ts
@@ -73,10 +73,7 @@ export interface UserProfile {
   experiment_bucket_map: ExperimentBucketMap;
 }
 
-export type EventTags = {
-  [key: string]: string | number | null;
-};
-
+export type EventTags =  Record<string, unknown>;
 export interface UserProfileService {
   lookup(userId: string): UserProfile;
   save(profile: UserProfile): void;
--- a/lib/shared_types.ts
+++ b/lib/shared_types.ts
@@ -73,7 +73,13 @@ export interface UserProfile {
   experiment_bucket_map: ExperimentBucketMap;
 }
 
-export type EventTags =  Record<string, unknown>;
+export type EventTags = {
+  revenue?: string | number | null;
+  value?: string | number | null;
+  $opt_event_properties?: Record<string, unknown>;
+  [key: string]: unknown;
+};
+
 export interface UserProfileService {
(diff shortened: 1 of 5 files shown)


lib/shared_types.ts
```diff
@@ -73,10 +73,7 @@ export interface UserProfile {
   experiment_bucket_map: ExperimentBucketMap;
 }
 
-export type EventTags = {
-  [key: string]: string | number | null;
-};
-
+export type EventTags =  Record<string, unknown>;
```
RowMur, line 76: The [docs](https://docs.developers.optimizely.com/feature-experimentation/docs/include-event-tags#reserved-tag-keys) mention some reserved keys so I think this type can be stricter and actually just use those. Perhaps it should be...

```ts
export type EventTags = {
  revenue: number
  value: number
  $opt_event_properties: Record<string, unknown>
}
```

junaed-optimizely: This is technically a breaking change and we have plan to add this in the next major version of the SDK which is very close to release. You can check #1039 for more detail about that. 

For version 5 we do not want to apply this and keep it as close as possible to the original definition.

lib/utils/event_tag_utils/index.ts
```diff
@@ -42,7 +41,8 @@ export function getRevenueValue(eventTags: EventTags, logger: LoggerFacade): num
     return null;
   }
 
-  const parsedRevenueValue = typeof rawValue === 'string' ? parseInt(rawValue) : rawValue;
+  const parsedRevenueValue =
```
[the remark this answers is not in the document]
junaed-optimizely: FYI, Here is one [explanation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#using_parseint_on_non-strings) on why we should use [Math.trunc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc) instead of [parseInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) when it comes to `number` primitive. That might not be a bad idea.

raju-opti (approved):

PR #1041, opened 2025-05-02
[FSSDK-11492] make public promise rejection messages non tree-shakable

raju-opti:
## Summary
- The mechanism in place for errorNotifier ensures that error messages used with OptimizelyError get tree-shaken if errorNotifier is not used. But some promise rejection errors are part of the public api and those messages must not be tree shake. This PR makes those message non tree-shakable.

## Issues
- FSSDK-11492


lib/service.ts
```diff
@@ -130,3 +130,7 @@ export abstract class BaseService implements Service {
   
   abstract stop(): void;
 }
+
+
+export const SERVICE_FAILED_TO_START = '%s failed to start, reason: %s';
+export const SERVICE_STOPPED_BEFORE_RUNNING = '%s stopped before running';
```
junaed-optimizely, line 136: NIT: Its just cosmetic! I'd prefer importing these constants from top of the file than bottom.

junaed-optimizely (approved):
LGTM!

PR #1042, opened 2025-05-06
[FSSDK-11494] fix UserAttributes type

raju-opti:
## Summary
- Current type does not work when there are control attributes. This PR fixes that

## Issues
- FSSDK-11494

### DIFF
--- a/lib/event_processor/event_builder/user_event.ts
+++ b/lib/event_processor/event_builder/user_event.ts
@@ -254,23 +254,30 @@ const buildVisitorAttributes = (
   attributes?: UserAttributes,
   logger?: LoggerFacade
 ): VisitorAttribute[]  => {
-  const builtAttributes: VisitorAttribute[] = [];
-  // Omit attribute values that are not supported by the log endpoint.
-  if (attributes) {
-    Object.keys(attributes || {}).forEach(function(attributeKey) {
-      const attributeValue = attributes[attributeKey];
-      if (isAttributeValid(attributeKey, attributeValue)) {
-        const attributeId = getAttributeId(configObj, attributeKey, logger);
-        if (attributeId) {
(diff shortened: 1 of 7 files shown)


lib/event_processor/event_builder/user_event.ts
```diff
@@ -254,23 +254,30 @@ const buildVisitorAttributes = (
...
-            value: attributeValue!,
-          });
-        }
+  const builtAttributes: VisitorAttribute[] = [];
+  Object.keys(attributes).forEach(function(attributeKey) {
+    const attributeValue = attributes[attributeKey];
+
+    if (typeof attributeValue === 'object' || typeof attributeValue === 'undefined') {
+      return;
+    }
```
junaed-optimizely, line 268: What's the reason for adding this guard if `isAttributeValid` already validates the attribute?

junaed-optimizely (changes requested):

lib/event_processor/event_builder/user_event.ts
raju-opti: this is needed for type narrowing. this value is used in line 276 which require string | number | boolean

junaed-optimizely: We could narrow the type  in `isAttributeValid` function using type predicates. 
```ts
export function isAttributeValid(
  attributeKey: unknown,
  attributeValue: unknown
): attributeValue is string | number | boolean {
   // other code
}
```

raju-opti: isAttributeValid does more validations than just checking that the value is string | number | boolean. It also takes attributeKey as parameter. I think this would not be appropriate to turn this into a type predicate

junaed-optimizely (approved):

PR #1043, opened 2025-05-06
[FSSDK-11497] fire config update events synchronously

raju-opti:
## Summary
- fire the project config updates synchronously. this will allow the optimizely class to propagate the initial project config synchronously to all sub components. This can be useful when the client is initialized with a datafile. One downside is, the update event handler needs to be registered before calling start() on the config manager, but since we only use the config manager internally, we can maintain that.

## Issues
- FSSDK-11497

### DIFF
--- a/lib/project_config/project_config_manager.ts
+++ b/lib/project_config/project_config_manager.ts
@@ -18,7 +18,6 @@ import { createOptimizelyConfig } from './optimizely_config';
 import {  OptimizelyConfig } from '../shared_types';
 import { DatafileManager } from './datafile_manager';
 import { ProjectConfig, toDatafile, tryCreatingProjectConfig } from './project_config';
-import { scheduleMicrotask } from '../utils/microtask';
 import { Service, ServiceState, BaseService } from '../service';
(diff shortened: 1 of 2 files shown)


lib/project_config/project_config_manager.ts
```diff
@@ -166,9 +165,7 @@ export class ProjectConfigManagerImpl extends BaseService implements ProjectConf
       if (this.projectConfig?.revision !== config.revision) {
         this.projectConfig = config;
         this.optimizelyConfig = undefined;
-        scheduleMicrotask(() => {
-          this.eventEmitter.emit('update', config);
-        })
```
junaed-optimizely, line 171: Looks like it goes way back to the ODP implementation. Why it was needed in the first place ? and why it is not required now?

raju-opti: this is not strictly "required", but might be useful if someone wants to send odp events without awaiting onReady().

It was done async previously cause I think 'update' event listeners were being added after the initial datafile parse took place, so the intitial config update was being missed.

junaed-optimizely (approved):

PR #1053, opened 2025-05-15
[FSSDK-11513] limit number of events in the eventStore

raju-opti:
## Summary
- saving failed events in the event store indefinitely might cause an app to consume too much storage space. This PR limits the number of events stored in the eventStore.

## Test plan
- added tests

## Issues
- FSSDK-11513

### DIFF
--- a/lib/event_processor/batch_event_processor.ts
+++ b/lib/event_processor/batch_event_processor.ts
@@ -62,7 +62,6 @@ export type BatchEventProcessorConfig = {
 
 type EventBatch = {
   request: LogEvent,
-  // ids: string[],
   events: EventWithId[],
 }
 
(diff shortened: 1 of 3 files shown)


lib/event_processor/batch_event_processor.ts
```diff
@@ -31,13 +31,16 @@ import { FAILED_TO_DISPATCH_EVENTS, SERVICE_NOT_RUNNING } from "error_message";
...
+import { EVENT_STORE_FULL } from "../message/log_message";
 
 export const DEFAULT_MIN_BACKOFF = 1000;
 export const DEFAULT_MAX_BACKOFF = 32000;
+export const MAX_EVENTS_IN_STORE = 500;
 
 export type EventWithId = {
   id: string;
   event: ProcessableEvent;
+  notStored?: boolean;
```
junaed-optimizely, line 43: NIT: For better readability, consider renaming `notStored` to `stored` or` isStored` —positive flags are generally easier to understand.

junaed-optimizely (approved):
LGTM! Its a nice optimization!

lib/event_processor/batch_event_processor.ts
raju-opti: actually, as we are storing these in the cache, not having the field for the stored events will be better. We could create a separate type for the store, but wanted to avoid that too. Hope it's not too confusing.

PR #1056, opened 2025-05-16
[FSSDK-11510] refactor unsupported factories to not throw

raju-opti:
## Summary
- previously we were throwing from unsupported factories. This PR changes them to return undefined instead. It will make writing isomorphic apps easier.

## Issues
- FSSDK-11510


junaed-optimizely (approved):
LGTM! Nice improvement!

PR #1072, opened 2025-06-25
[FSSDK-11207] add multi region support for logx events

raju-opti:
## Summary
logx event will now be dispatched to different urls based on region field in the datafile

## Test plan
- added and upldated relevant tests

## Issues
- FSSDK-11207


jaeopt (approved):
LGTM

PR #1086, opened 2025-10-06
[FSSDK-11898] serialize concurrent cmab service calls

raju-opti:
## Summary

The cmab service caches the results of a cmab prediction retrieved from the server and returns it for subsequent calls. This ensures a consistent value is returned for getDecision() within the cache ttl. However, when there is no cached value, if there are concurrent calls to gertDecision() for same userId and ruleId combination, all of these will cause a call to the server and may potentially return different values.

The solution is to run concurrent calls for same userId and ruleId combinations one after another. To achieve this, we put each (userId, ruleId) combination in one of the predefined bucktes by hashing the (userId, ruleId) combination and serialize all calls for that particular hash % (num_buckets).

## Test plan
- added tests

## Issues
- FSSDK-11898

### DIFF
--- a/lib/utils/executor/serial_runner.spec.ts
+++ b/lib/utils/executor/serial_runner.spec.ts
@@ -13,12 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, beforeEach } from 'vitest';
 
 import { SerialRunner } from './serial_runner';
 import { resolvablePromise } from '../promise/resolvablePromise';
 import { exhaustMicrotasks } from '../../tests/testUtils';
-import { Maybe } from '../type';
 
 describe('SerialRunner', () => {
   let serialRunner: SerialRunner;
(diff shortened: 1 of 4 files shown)


lib/utils/executor/serial_runner.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+import { SerialRunner } from './serial_runner';
+import { resolvablePromise } from '../promise/resolvablePromise';
+import { exhaustMicrotasks } from '../../tests/testUtils';
+import { Maybe } from '../type';
```
junaed-optimizely, line 21: We can cleanup this import

lib/utils/executor/serial_runner.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+  });
+
+  it('should reject with same error when the passed function rejects', async () => {
+    const error = new Error('test error');
+    const fn = () => Promise.reject(error);
+    
+    await expect(serialRunner.run(fn)).rejects.toThrow(error);
+  });
+
+  it('should execute multiple async functions in order', async () => {
```
junaed-optimizely, line 45: NIT: This test is very verbose and seems like we don't have to make it this complicated. Can we make it simpler ?

junaed-optimizely (approved):
Great solution! LGTM!

lib/utils/executor/serial_runner.spec.ts
raju-opti: There might be, though I cannot think one immediately. Do you have any suggestions?

Issue #991, opened 2025-01-27
[BUG] Typo in Type Definition: experimentsIds should be experimentIds


alycecristines:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### SDK Version

"@optimizely/optimizely-sdk": "5.3.4"

### Current Behavior

There is a typo in the type definition of the library. The object for getConfig events returned by the SDK uses the property experimentIds, but the type definition incorrectly specifies it as experimentsIds.

### Expected Behavior

The type definition should correctly use experimentIds to match the actual property returned by the SDK.

### Steps To Reproduce

Use the SDK method that returns an object containing config events array with experimentIds array.
Observe the type definition mismatch, which causes TypeScript to throw an error.

### SDK Type

Browser

### Node Version

v20.15.1

### Browsers impacted

all

### Severity

Minor issue

PR #1088, opened 2025-10-07
[FSSDK-11925] fix type in event.experimentIds field in projectConfig

raju-opti:
## Issues
- FSSDK-11925
- #991

### DIFF
--- a/lib/core/project_config/index.ts
+++ b/lib/core/project_config/index.ts
@@ -50,6 +50,10 @@ interface Event {
   key: string;
   id: string;
   experimentsIds: string[];
+
+  // the field is named experimentIds in the datafile, but this type previously defined it as experimentsIds.
+  // keeping both to avoid breaking changes, removed experimentsIds in v6.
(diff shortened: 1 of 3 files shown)


lib/core/project_config/index.ts
```diff
@@ -142,6 +146,16 @@ export const createProjectConfig = function(datafileObj?: JSON, datafileStr: str
...
   projectConfig.__datafileStr = datafileStr === null ? JSON.stringify(datafileObj) : datafileStr;
 
+  // Copy experimentIds to experimentsIds in each event to fix typo in property name
+  // https://github.com/optimizely/javascript-sdk/issues/991
+  if (!projectConfig.events) {
+    projectConfig.events = [];
+  }
+  
+  projectConfig.events.forEach(event => {
+    event.experimentsIds = event.experimentIds;
```
junaed-optimizely, line 156: Since you are modifying events here, you need to copy the events array first in the `createMutationSafeDatafileCopy` (similar to rollouts, groups etc...) to avoid mutating the original datafile.

junaed-optimizely (changes requested):
I have once concern and that might be reason for failing tests.

PR #1097, opened 2025-10-20
[FSSDK-11981] bug bash addressed

junaed-optimizely:
## Summary

**Logging improvements for CMAB cache operations:**

* Added new log message constants for CMAB cache events (`IGNORE_CMAB_CACHE`, `RESET_CMAB_CACHE`, `INVALIDATE_CMAB_CACHE`, `CMAB_CACHE_ATTRIBUTES_MISMATCH`) in `lib/message/log_message.ts`.

**Added missing logger in `DefaultCmabService`.**

## Test plan
Existing test should pass

## Issues
- FSSDK-11981

### DIFF
--- a/lib/index.browser.tests.js
+++ b/lib/index.browser.tests.js
@@ -67,6 +67,7 @@ var getLogger = () => ({
   warn: () => {},
   error: () => {},
   child: () => getLogger(),
+  setName: () => {},
 })
 
 describe('javascript-sdk (Browser)', function() {


lib/client_factory.ts
```diff
@@ -66,6 +66,7 @@ export const getOptimizelyInstance = (config: OptimizelyFactoryConfig): Optimize
   const cmabService = new DefaultCmabService({
     cmabClient,
     cmabCache,
+    logger
```
raju-opti: instead of the root logger, we can pass in a logger.child(). Also, we should set the name of the logger in cmabService similar to for example `ProjectConfigManagerImpl`

lib/core/decision_service/cmab/cmab_service.ts
```diff
@@ -121,13 +134,15 @@ export class DefaultCmabService implements CmabService {
 
     if (cachedValue) {
       if (cachedValue.attributesHash === attributesHash) {
+        this.logger?.debug(CMAB_CACHE_HIT, userId, ruleId);
         return { variationId: cachedValue.variationId, cmabUuid: cachedValue.cmabUuid };
       } else {
+        this.logger?.debug(CMAB_CACHE_MISS, userId, ruleId);
```
raju2592: I think the previous log was better here. We can add the cache miss log in a new else block for the outer if(cachedValue) {}

PR #1112, opened 2025-11-24
[FSSDK-12092] add platform verification for imports 

raju-opti:
## Summary
This PR implements a comprehensive platform isolation system for a multi-platform JavaScript SDK that supports Browser, Node.js, and React Native. The system enforces that platform-specific code cannot be imported by incompatible platforms, preventing runtime errors when bundling for different targets.

**Key changes:**
- Adds validation scripts that check `__platforms` exports and import compatibility
- Adds script to automatically add `__platforms` exports based on file name
- Integrates ESLint rule to enforce platform declarations at development time
- Adds extensive documentation and troubleshooting guides
- Updates all source files to include `__platforms` exports

## Test plan
- extensive manual testing

## Issues
- FSSDK-12092

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "clean:win": "(if exist dist rd /s/q dist)",
     "lint": "tsc --noEmit && eslint 'lib/**/*.js' 'lib/**/*.ts'",
     "validate-platform-isolation": "./scripts/platform-validator.js --validate",
-    "fix-platform-isolation": "./scripts/platform-validator.js --fix",
+    "fix-platform-export": "./scripts/platform-validator.js --fix-export",
     "test-isolation-rules": "./scripts/test-validator.js",
     "test-vitest": "vitest run",
     "test-mocha": "TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\" }' mocha -r ts-node/register -r tsconfig-paths/register -r lib/tests/exit_on_unhandled_rejection.js 'lib/**/*.tests.ts' 'lib/**/*.tests.js'",
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -2,9 +2,28 @@
 
 This directory contains build and validation scripts for the JavaScript SDK.
 
+## platform-validator.js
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -97,6 +100,7 @@
     "uuid": "^10.0.0"
   },
   "devDependencies": {
+    "@biomejs/biome": "^2.3.7",
```
junaed-optimizely: Are we using this ?

lib/error/error_handler.ts
```diff
@@ -1,3 +1,5 @@
+import { Platform } from './../platform_support';
+
```
junaed-optimizely: This should not be before copyright text!

lib/message/error_message.ts
```diff
@@ -1,3 +1,5 @@
+import { Platform } from './../platform_support';
+
```
junaed-optimizely: This should not be before copyright text!

lib/odp/constant.ts
```diff
@@ -1,3 +1,5 @@
+import { Platform } from './../platform_support';
+
```
junaed-optimizely: This should not be before copyright text!

lib/utils/microtask/index.ts
```diff
@@ -1,3 +1,5 @@
+import { Platform } from './../../platform_support';
+
```
junaed-optimizely: This should not be before copyright text!

lib/platform_support.ts
junaed-optimizely: Please add copyright text here!

lib/message/log_message.ts
```diff
@@ -1,3 +1,5 @@
+import { Platform } from './../platform_support';
+
```
junaed-optimizely: I can see multiple files where this import has been inserted at the top which should not be the case. I think  it might be something to do with the `ensurePlaftormImport` function!

junaed-optimizely (changes requested):

PR #1141, opened 2026-03-20
fix: remove polyfills from the library

matinzd:
Removing two import statements that are no not needed. Polyfills should be imported/configured by the user in their project. Some people might want to use https://github.com/margelo/react-native-quick-crypto 

/cc @mrousavy

### DIFF
--- a/lib/index.react_native.ts
+++ b/lib/index.react_native.ts
@@ -13,9 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import 'fast-text-encoding';
-import 'react-native-get-random-values';
-
 import { Client, Config } from './shared_types';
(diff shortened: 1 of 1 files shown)


lib/index.react_native.ts
```diff
@@ -13,9 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import 'fast-text-encoding';
-import 'react-native-get-random-values';
-
```
junaed-optimizely, line 18: I agree users should be able to bring their own polyfills, but removing these imports outright would be a breaking change.

A better approach for the current version would be conditional imports (similar to how we handle `@react-native-async-storage/async-storage`) - check if the globals already exist before requiring the polyfill packages.

junaed-optimizely (changes requested):

PR #1152, opened 2026-05-26
[AI-FSSDK] [FSSDK-12670] Block ODP identify event for single identifier

jaeopt:
## Summary

Fixes ODP identify event to only send when multiple valid identifiers exist. Previously, the SDK sent an identify event even with a single identifier (just fs_user_id or just vuid), which provides no identity-linking value and generates unnecessary network traffic.

## Changes
- Added identifier count guard in `DefaultOdpManager.identifyUser` to skip event dispatch when fewer than 2 valid identifiers are present
- Added debug-level logging when the identify event is skipped due to single identifier
- Updated tests for single-identifier cases to expect no event sent

## Jira Ticket
[FSSDK-12670](https://optimizely-ext.atlassian.net/browse/FSSDK-12670)

## Notes
- Reference implementation: Python SDK PR #513

[FSSDK-12670]: https://optimizely-ext.atlassian.net/browse/FSSDK-12670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/odp/odp_manager.ts
+++ b/lib/odp/odp_manager.ts
@@ -32,6 +32,7 @@ import { Maybe } from '../utils/type';
 import { sprintf } from '../utils/fns';
 import { SERVICE_STOPPED_BEFORE_RUNNING } from '../service';
 import { Platform } from '../platform_support';
+import { ODP_IDENTIFY_NOT_DISPATCHED } from '../message/log_message';
 
 export interface OdpManager extends Service {
   updateConfig(odpIntegrationConfig: OdpIntegrationConfig): boolean;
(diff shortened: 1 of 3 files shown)


lib/odp/odp_manager.ts
```diff
@@ -229,6 +229,13 @@ export class DefaultOdpManager extends BaseService implements OdpManager {
       identifiers.set(ODP_USER_KEY.FS_USER_ID, finalUserId);
     }
 
+    // Identify requires 2+ identifiers to link (e.g., vuid + fs_user_id).
+    // A single identifier has no cross-reference value and generates unnecessary traffic.
+    if (identifiers.size < 2) {
+      this.logger.log(LogLevel.Debug, 'ODP identify event is not dispatched (fewer than 2 valid identifiers).');
```
raju-opti: this is not a supported interface. Should use `this.logger.debug(msg)` instead

raju-opti: also, logger can be undefined. should use this.logger?.debug

raju-opti (changes requested):

raju-opti (approved):

Issue #1155, opened 2026-06-11
[BUG] Content-Length set from String.length (UTF-16), not UTF-8 byte length — truncates multi-byte request bodies


terraelise:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### SDK Version

6.4.0 (@optimizely/optimizely-sdk); also on master as of June 2026. Node.js 18+ (tested on 24).

## Summary

`RequestHandler.makeRequest` (Node) sets the `content-length` header from the request body's JavaScript string `.length` (UTF-16 code-unit count), but writes the body to the socket as UTF-8. For any request body containing multi-byte UTF-8 characters (non-Latin text, full-width punctuation, emoji, accented characters), the header under-reports the actual byte size, so the receiving server gets a **truncated** body.

For the event dispatcher this shows up as `400` responses from the events endpoint (`https://logx.optimizely.com/v1/events`) with an "Invalid json" / parse error, because the JSON is cut off mid-string.

## Affected code

`lib/utils/http_request_handler/request_handler.node.ts`, in `makeRequest`:

```ts
headers: {
  ...headers,
  'content-length': String(data?.length || 0), // ← UTF-16 code units, not bytes
},
...
if (data) {
  request.write(data); // ← written as UTF-8
}
```

Present in `@optimizely/optimizely-sdk@6.4.0` and still on `master` as of June 2026.

## Why it's wrong

`String.prototype.length` counts UTF-16 code units, not bytes; `request.write(string)` encodes as UTF-8 by default. The two differ for any character outside the single-byte range, so `Content-Length` is too small and the body is truncated.

## Impact

- Any event whose payload contains a multi-byte character is rejected (HTTP 400, truncated/invalid JSON).
- Events are batched, so a single multi-byte event poisons the entire batch's `Content-Length` — co-batched ASCII-only events fail too.
- Affects any app sending non-ASCII tag values/attributes (Japanese, emoji, accented names, smart quotes, etc.). It looks intermittent/random because it depends on the byte composition of each batch.

### Expected Behavior

Content-Length should equal the UTF-8 byte length of the request body, so multi-byte payloads are sent intact and accepted by the events endpoint.

### Steps To Reproduce

- @optimizely/optimizely-sdk: 6.4.0 (also reproduces on `master`)
- Node.js: 18+ (tested on 24)
- Platform: Node server

Standalone Node script (no SDK required) that mimics the handler — sets `Content-Length` from `String.length` and writes the body as UTF-8:

[...]

PR #1156, opened 2026-06-12
[FSSDK-12759] fix content-length header in NodeRequestHandler

raju-opti:
## Summary
- prevously, body.length was being used for content-length header, which is incorrect (The length data property of a String value contains the length of the string in UTF-16 code units, we need the byte length of utf-8 endcoded string). This PR fixes it. 

## Test plan
- added tests
- also tested with real node/python servers

## Issues
- FSSDK-12759
- #1155


junaed-optimizely (approved):
LGTM!

Issue #1147, opened 2026-05-04
[BUG] UUID <v14 has npm audit warning


Develliot:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### SDK Version

6.3.1

### Current Behavior

uuid  <14.0.0
Severity: moderate
uuid: Missing buffer bounds check in v3/v5/v6 when buf is provided - https://github.com/advisories/GHSA-w5hq-g745-h8pq

### Expected Behavior

no warnings when running npm audit

### Steps To Reproduce

1) install latest version of @optimizely/optimizely-sdk
2) run 'npm audit'

### SDK Type

Browser

### Node Version

Node 22 and Node 24

### Browsers impacted

n/a

### Severity

Minor issue

### Workaround/Solution

none

raju-opti:
Hey @Develliot , thank you for reporting this. We only use `uuid.v4`, so it should not have any security implications for the sdk. We will upgrade to [13.0.1](https://github.com/uuidjs/uuid/releases/tag/v13.0.1) in the next release (coming soon!) which has fix for this.

raju-opti:
This has been fixed in `v6.5.0`.

PR #1159, opened 2026-06-25
[FSSDK-12820] Upgrade to uuid v13

raju-opti:
## Summary

Upgrade uuid to v13 (ESM-only). Update rollup config to bundle uuid into CJS outputs instead of keeping it as an external dependency, since CJS consumers cannot require() an ESM-only package. Add a CJS require hook shim so mocha/ts-node tests can still load uuid in CommonJS mode.

## Test plan
- added test app to validate nodejs require(). Also added to CI

## Issues
- FSSDK-12820
- #1147


junaed-optimizely (approved):
LGTM!

PR #1170, opened 2026-07-21
[FSSDK-12933] Refactor message generator to use TS compiler API

raju-opti:
## Summary
Replace the dynamic-import-based message_generator.ts with a static TS-parser approach (scripts/generate-messages.js) that preserves declaration order and eliminates the jiti runtime dependency. Add a test script and CI job for the generator.

## Test plan
- added test for the new script

## Issues
- FSSDK-12933


scripts/generate-messages.js
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env node
```
junaed-optimizely, line 1: `#!` is UNIX only! Not that it matters a lot but `build:win` could potentially be broken for this.

junaed-optimizely (approved):
Approved with one minor concern!

scripts/generate-messages.js
raju-opti: Yeah. But other scripts are also using this currently. Windows build is relevant only for devs' local machine, but we can use wsl nowadays. We can come back and fix the windows build later if needed

PR #1175, opened 2026-09-07
[FSSDK-13110] cache update

junaed-optimizely:
## Summary
- Bump the datafile cache key to `opt-datafile-v6-<sdkKey>` so v6 never reads a datafile persisted by an older SDK under the old key.
- Make `AsyncStorageCache/LocalStorageCache` `Store<string>` (no internal JSON.stringify/parse) and move event serialization into `EventStore`.

## Test plan
- npx vitest run — all 1091 tests pass, no type errors.
- Updated cache and EventStore specs for string-only behavior.
## Issues
- FSSDK-13110
- https://github.com/optimizely/react-sdk/issues/347

### DIFF
--- a/lib/project_config/polling_datafile_manager.ts
+++ b/lib/project_config/polling_datafile_manager.ts
@@ -73,7 +73,7 @@ export class PollingDatafileManager extends BaseService implements DatafileManag
       logger,
     } = config;
     this.cache = cache;
-    this.cacheKey = 'opt-datafile-' + sdkKey;
+    this.cacheKey = 'opt-datafile-v6-' + sdkKey;
     this.sdkKey = sdkKey;
     this.datafileAccessToken = datafileAccessToken;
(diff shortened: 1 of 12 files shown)


lib/project_config/polling_datafile_manager.ts
```diff
@@ -73,7 +73,7 @@ export class PollingDatafileManager extends BaseService implements DatafileManag
       logger,
     } = config;
     this.cache = cache;
-    this.cacheKey = 'opt-datafile-' + sdkKey;
+    this.cacheKey = 'opt-datafile-v6-' + sdkKey;
```
raju-opti: the reference to v6 might be a bit confusing, for example when we move to v7. we can rename to something else like `optimizely-datafile`

raju-opti (changes requested):
one small comment

raju-opti (approved):

